PR #155, opened 2022-09-14
Feature/context debugger

JesmoDev:
Adds a debugger for stores registered at the outer level of the app.
It has a lot of limitations for now and will be rewritten at a later point.
But for now it can be used to speed up the process of checking the data in the stores.


iOvergaard:
Just closing this so it doesn't take up space. Please re-open if it should be merged later on.

PR #158, opened 2022-09-22
Feature/dashboard telemetry

loivsen:


iOvergaard (changes requested):
Please add a .test.ts file with the default "instanceof" and "isAccessible" tests

iOvergaard (approved):

PR #175, opened 2022-09-27
Feature/Data Type Config

madsrasmussen:


temp-schema-generator/manifests.ts
```diff
@@ -112,9 +99,12 @@ export interface MetaTreeItemAction {
 	icon: string;
 	weight: number;
 }
-export interface MetaPropertyEditorUI extends IPrevalues {
+export interface MetaPropertyEditorUI {
+	label: string;
+	propertyEditor: string;
 	icon: string;
 	group: string;
+	config?: PropertyEditorConfig;
```
iOvergaard, line 107: Would it be realistic to have multiple configs to a property editor UI, e.g. having the config registered as its own extension instead of hooked on to a specific propertyEditorUI?

(Sorry if this is starting a bigger discussion)

temp-schema-generator/manifests.ts
```diff
@@ -128,6 +118,7 @@ export interface MetaEditorView {
 	editors: string[];
 	pathname: string;
 	weight: number;
+	label: string;
```
iOvergaard, line 121: There's a TODO somewhere to discuss if `label` should be required on all manifests so we know what to call them if we ever need to display anything regarding a specific manifest in the backoffice, e.g. a table of all manifests. What do you think about that?

vite.config.ts
```diff
@@ -1,8 +1,19 @@
...
 // https://vitejs.dev/config/
 export default defineConfig({
-  build: {
-    sourcemap: true,
-  },
+	build: {
+		sourcemap: true,
+	},
+	plugins: [
+		viteStaticCopy({
```
iOvergaard, line 10: Is there a particular reason this doesn't work out of the box in the normal "public" folder?

src/mocks/browser-handlers.ts
```diff
@@ -18,6 +19,7 @@ const handlers = [
 	...dataTypeHandlers,
 	...documentTypeHandlers,
 	...treeHandlers,
+	...propertyEditorHandlers,
```
iOvergaard, line 22: This should probably be added to `e2e-handlers.ts` as well so we have some mocked property editors in Playwright.

iOvergaard (changes requested):

package.json
```diff
@@ -16,20 +16,22 @@
...
-		"build:production": "tsc && vite build",
+		"pre-run": "npm run generate:icons",
+		"dev": "npm run pre-run && vite",
+		"build": "npm run pre-run && tsc && vite build --mode staging",
+		"build:production": "npm run pre-run && tsc && vite build",
 		"preview": "vite preview --open",
-		"test": "web-test-runner --coverage",
-		"test:watch": "web-test-runner --watch",
-		"test:e2e": "playwright test",
+		"test": "npm run pre-run && web-test-runner --coverage",
```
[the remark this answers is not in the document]
madsrasmussen: I did it this way so we didn't have to commit the output files. If we want to change it I need to commit those?

package.json
```diff
@@ -44,6 +46,7 @@
 		"@umbraco-ui/uui-modal-sidebar": "file:umbraco-ui-uui-modal-sidebar-0.0.0.tgz",
 		"element-internals-polyfill": "^1.1.14",
 		"lit": "^2.3.1",
+		"lodash": "^4.17.21",
```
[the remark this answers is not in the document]
madsrasmussen: Ah, good catch. I will change.

package.json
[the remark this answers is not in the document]
iOvergaard: Yeah sure, let's just commit the output files. They only change when a new icon is added, right?

temp-schema-generator/manifests.ts
madsrasmussen: I thought about it, but I couldn't come up with a case where you need to extend another Property Editors config. You can't really access the config from within their editor.

In my opinion we should leave it like this for now. Then we can come back to it if a good use case shows up. (Let me know if you have one in mind :D )

temp-schema-generator/manifests.ts
madsrasmussen: We currently have a `name` property on all manifests that is used for this. The label is used in cases where we need a shorter "display" version of the name. Ex. Trees, Editor Views etc.

iOvergaard: Yeah I know. Was just thinking of the TODO if we should remove it.

package.json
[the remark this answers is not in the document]
madsrasmussen: Yeah. we will have to run the script manually if icons are updated, added or deleted. I have committed the files now. Then we can change if we forget it :)

src/mocks/browser-handlers.ts
madsrasmussen: This is updated now 👍

package.json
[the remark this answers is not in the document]
madsrasmussen: I am running into some issues with Playwright after the change to lodash-es. It seems like Node is complaining about the import of es modules. I can't figure out why it is only a problem here. I have seen some suggestions to add `"type": "module"` to the `package.json `but then I can't get the `playwright.config` file to work with Typescript.

https://github.com/microsoft/playwright/issues/12536

Is this something you have experienced when you set up Playwright?

temp-schema-generator/manifests.ts
madsrasmussen: I have added label to the section manifest which is the last one (for now) that needs it. I have also removed the TODO.

iOvergaard (approved):

PR #206, opened 2022-10-12
Feature/users section (Work in progress)

JesmoDev:
WIP User and User Group section


madsrasmussen (approved):
We are merging this PR because it consist changes we need in other places at the moment. A full review will happen when the feature is done.

PR #227, opened 2022-10-24
Feature/dashboard examine management

loivsen:


src/backoffice/dashboards/examine-management/views/section-view-examine-indexers.ts
```diff
@@ -0,0 +1,224 @@
...
+	@state()
+	private _indexData!: Index;
+
+	private _notificationService?: UmbNotificationService;
+	private _modalService?: UmbModalService;
+
+	private async _getIndexData() {
+		try {
+			const index = await SearchResource.getSearchIndexByIndexName({ indexName: this.indexName });
+			this._indexData = index as Index;
```
iOvergaard, line 98: Is this not automatically inferred as `Index` without specifying `as Index` here and elsewhere?

src/backoffice/dashboards/examine-management/views/section-view-examine-overview.ts
```diff
@@ -0,0 +1,177 @@
...
+		super();
+		this._getIndexers();
+		this._getSearchers();
+
+		this.consumeContext('umbNotificationService', (notificationService: UmbNotificationService) => {
+			this._notificationService = notificationService;
+		});
+		this.consumeContext('umbModalService', (modalService: UmbModalService) => {
+			this._modalService = modalService;
+		});
```
iOvergaard, line 102: Consider the new `consumeAllContexts` method to combine these two here and elsewhere

src/backoffice/dashboards/examine-management/views/section-view-examine-searchers.ts
```diff
@@ -0,0 +1,306 @@
...
+import { UmbModalService } from '../../../../core/services/modal';
+import { UmbNotificationService } from '../../../../core/services/notification';
+import { UmbNotificationDefaultData } from '../../../../core/services/notification/layouts/default';
+
+import { UmbContextConsumerMixin } from '@umbraco-cms/context-api';
+
+import { ApiError, ProblemDetails, SearchResult, SearchResource, Field } from '@umbraco-cms/backend-api';
+
+import '../../../../core/services/modal/layouts/fields-viewer/fields-viewer.element';
+import '../../../../core/services/modal/layouts/fields-viewer/fields-settings.element';
```
iOvergaard, line 14: Let's make a shortcut to import from the services folder, just like we did for `@umbraco-cms/backend-api` and others. I believe they are made in the `tsconfig.json` file.

src/backoffice/dashboards/examine-management/views/section-view-examine-searchers.ts
```diff
@@ -0,0 +1,306 @@
...
+				const data: UmbNotificationDefaultData = { message: error.message ?? 'Could not fetch search results' };
+				this._notificationService?.peek('danger', { data });
+			}
+		}
+	}
+
+	private _updateFieldFilter() {
+		this._searchResults?.map((doc) => {
+			const document = doc.fields?.filter((field) => {
+				return field.name?.toUpperCase() != 'NODENAME';
```
iOvergaard, line 163: Consider using `!==` instead of `!=` for strict compare.

```suggestion
				return field.name?.toUpperCase() !== 'NODENAME';
```

src/core/mocks/domains/examine-management.handlers.ts
```diff
@@ -0,0 +1,76 @@
...
+	rest.get(umbracoPath('/search/index'), (_req, res, ctx) => {
+		return res(
+			// Respond with a 200 status code
+			ctx.status(200),
+			ctx.json<PagedIndex>(PagedIndexers)
+		);
+	}),
+
+	// TODO: when using the umbracoPath helper you have to write parameters like {indexName}. MSW wants parameters as :indexName
+	rest.get('/umbraco/management/api/v1/search/index/:indexName', (_req, res, ctx) => {
```
iOvergaard, line 17: The `umbracoPath` function just prepends the base URL to the path. So it should work even with `:indexName`. Are you sure this does not work?

```suggestion
	rest.get(umbracoPath('/search/index/:indexName'), (_req, res, ctx) => {
```

src/backoffice/dashboards/examine-management/views/section-view-examine-indexers.ts
```diff
@@ -0,0 +1,224 @@
...
+				</uui-table-row>
+				<uui-table-row>
+					<uui-table-cell style="width:0px; font-weight: bold;"> fieldCount </uui-table-cell>
+					<uui-table-cell>${this._indexData.fieldCount} </uui-table-cell>
+				</uui-table-row>
+				${this._indexData.providerProperties
+					? Object.entries(this._indexData.providerProperties).map((entry) => {
+							return html`<uui-table-row>
+								<uui-table-cell style="width:0px; font-weight: bold;"> ${entry[0]} </uui-table-cell>
+								<uui-table-cell clip-text> ${JSON.stringify(entry[1]).replace(/,/g, ', ')} </uui-table-cell>
```
iOvergaard, line 194: I think it works a lot better by just printing out the value, which will inherently invoke its own `toString` function.

```suggestion
								<uui-table-cell clip-text> ${entry[1]} </uui-table-cell>
```

Example where boolean values are still there and strings don't have quotes:

image

src/backoffice/dashboards/examine-management/dashboard-examine-management.element.ts
```diff
@@ -1,17 +1,66 @@
...
+	}
 
 	render() {
-		return html`
-			<uui-box>
-				<h1>Examine Management</h1>
-			</uui-box>
-		`;
+		return html` ${this.backbutton
+				? html` <a href="/section/settings/dashboard/examine-management/"> &larr; Back to overview </a> `
```
iOvergaard, line 61: I haven't actually toyed with it yet, but do we have a way to route relative to the current path, so we don't have to hardcode the entire URL in places such as this?

iOvergaard (changes requested):
Overall a very nice dashboard!

src/backoffice/dashboards/examine-management/dashboard-examine-management.element.ts
loivsen: I guess something like ${window.location.href}/index/${indexName}` would work when going from the overview and into the index view (and similar for searcher view), but then going back to the overview we would have to grab the URL and cut off the last part. I'm not sure if there are any other ways.

src/core/mocks/domains/examine-management.handlers.ts
loivsen: Seems to work, not sure why I didnt test it after we got the new API

src/backoffice/dashboards/examine-management/views/section-view-examine-overview.ts
loivsen: Sure

src/backoffice/dashboards/examine-management/dashboard-examine-management.element.ts
loivsen: This is going to be a TODO for the future...

iOvergaard (approved):

PR #234, opened 2022-10-31
Move Property Editor manifests from server to client

madsrasmussen:


src/core/extensions-registry/models.ts
```diff
@@ -0,0 +1,94 @@
...
+}
+
+export interface ManifestElement extends ManifestBase {
+	type: ManifestStandardTypes;
+	js?: string;
+	elementName?: string;
+	meta?: any;
+}
+
+export interface ManifestCustom extends ManifestBase {
```
iOvergaard, line 83: Not sure we can use this anymore, since the type forwarded from the server could be any string. A Vendr extension might have the type of "Vendr.Extension" so we should just have an interface with a generic string for fallback cases.

iOvergaard (approved):

PR #273, opened 2022-11-22
Feature/dashboard models builder

loivsen:


src/core/utils/code-block.ts
```diff
@@ -0,0 +1,49 @@
...
+				overflow-y: auto;
+				overflow-wrap: anywhere;
+			}
+		`,
+	];
+
+	render() {
+		return html`<div id="container">
+			<uui-scroll-container>
+				<slot></slot>
```
iOvergaard, line 39: Where did we land on the `<pre>` and `<code>` fields here?

src/core/utils/code-block.ts
```diff
@@ -0,0 +1,49 @@
...
+import { css, html, LitElement } from 'lit';
+import { customElement } from 'lit/decorators.js';
+
+/**
+ *  A simple styled box for showing code-based error messages.
+ *  @slot the full message
+ *
+ */
+@customElement('uui-code-block')
+export class UUICodeBlock extends LitElement {
```
iOvergaard, line 11: Add this as an export in the `index.ts` file in the same folder

iOvergaard (changes requested):

src/core/utils/code-block.ts
loivsen: Forgot to sync my changes before I requested a review - but i wrapped it in a <pre>. Normally you add in <code> too for each line of code, but i figured maybe the element may be used for other types of error messages too, not only a code block.  Maybe the element should be renamed?

iOvergaard: I think the name is fine. I would also just wrap it in `<code></code>` since any dumps of huge texts could be whatever and contain all sorts of stuff.

iOvergaard (approved):

PR #274, opened 2022-11-23
dashboard performance profiling

loivsen:


src/backoffice/dashboards/performance-profiling/dashboard-performance-profiling.element.ts
```diff
@@ -1,15 +1,85 @@
...
+		this.consumeAllContexts(['umbNotificationService'], (instances) => {
+			this._notificationService = instances['umbNotificationService'];
+		});
+	}
 
 	render() {
 		return html`
 			<uui-box>
 				<h1>Performance Profiling</h1>
+				<p>
```
iOvergaard, line 58: This text should only be shown if the profiler is actually enabled. If not, there is another text to show if you check "profiler.html":

```html
                    <p>
                        Umbraco currently does not run in debug mode, so you can't use the built-in profiler. This is how it should be for a production site.
                    </p>
                    <p>
                        Debug mode is turned on by setting <b>debug="true"</b> on the <b>&lt;compilation /&gt;</b> element in web.config.
                    </p>
```

src/backoffice/dashboards/performance-profiling/dashboard-performance-profiling.element.ts
```diff
@@ -1,15 +1,85 @@
...
+					If you want the profiler to be activated by default for all page renderings, you can use the toggle below. It
+					will set a cookie in your browser, which then activates the profiler automatically. In other words, the
+					profiler will only be active by default in your browser - not everyone else's.
+				</p>
+
+				<uui-toggle
+					label="Activate the profiler by default"
+					label-position="left"
+					.checked="${this._profilingStatus}"></uui-toggle>
```
iOvergaard, line 76: This should not show the value of profiling status but rather show if a specific cookie is enabled. We need to figure out if we are doing this with cookies or not, but I would save a state in localStorage for now if this has been toggled or not.

iOvergaard (changes requested):

iOvergaard (changes requested):
I like the feeling of the dashboard, but we mainly need to figure out how actually to enable debug mode. It's got to work different when the backoffice is detached, but for now let's just do as I suggested further down below and leave it up to the backenders what

iOvergaard (approved):

PR #310, opened 2022-12-13
Header apps as extension-slot

nielslyngsoe:
* Using the extension slot for Header Apps.
* Enabling the extensions to use the meta-object from the manifest.
* Sorting extensions in extension slots based on weight, if present.

Notice the improved version, compared to the one located in the tree-refactor PR.


src/core/extensions-api/use-extension-data.mixin.ts
```diff
@@ -0,0 +1,13 @@
+import type { HTMLElementConstructor } from '../models';
```
madsrasmussen, line 1: This doesn't seem to be used anymore. Can we delete this mixin?

src/backoffice/components/backoffice-frame/backoffice-header-apps.element.ts
```diff
@@ -0,0 +1,84 @@
...
+				display: flex;
+				align-items: center;
+				gap: var(--uui-size-space-2);
+			}
+		`,
+	];
+
+	constructor() {
+		super();
+		this._registerEditorApps();
```
madsrasmussen, line 23: Rename to _registerHeaderApps

src/backoffice/header-apps/header-app-current-user.element.ts
```diff
@@ -0,0 +1,33 @@
+import { UUITextStyles } from '@umbraco-ui/uui-css/lib';
+import { css, CSSResultGroup, html, LitElement } from 'lit';
+import { customElement } from 'lit/decorators.js';
+
+@customElement('umb-backoffice-header-app-user')
```
madsrasmussen, line 5: Can we change this element name to` umb-header-app-current-user` ?

madsrasmussen (approved):

PR #318, opened 2022-12-15
Feature/resource api mixin

iOvergaard:
This adds a controller that is very helpful in using the API resources. It has generic methods to return a tuple with the response of an API resource like:

```ts
const [result, error] = await this.resourceController.tryExecute(...);

if (error) {
   console.error(error);
} else if (result) {
   // do something with result
}
```

More importantly, it also has a helper method that will handle generic error notifications in a way that ensures, that any developer will use the "best approach" set forward by Umbraco HQ (because we can change the controller if something better comes up):

```ts
const [profilingStatus] = await this._resourceController.tryExecuteAndNotify(ProfilingResource.getProfilingStatus());
if (profilingStatus) {
  this._profilingStatus = profilingStatus.enabled;
}
```

Here we are only interested in the result if it's there, and if not we know that something went wrong and can avoid removing the loading state, and we can be sure that the controller has notified the user appropriately.

The controller also supports you in doing your own error handling by using `ResourceController.execute<T>(resource)`, which will return the API resource directly.

Lastly, the controller understands the lifetime of its connected component and, if given a CancelablePromise, will send out an AbortSignal to cancel any running requests if the component is dismounted (i.e. if the user clicks on another tab). For the time being, all resources will be automatically added to this queue, but we might look into having long-lived promises running that will not be canceled if need be. But you are always free to call the API resources manually.

The controller currently uses an experimental way of getting the NotificationService by way of instrumenting a ContextConsumer manually. This can and will be improved soon-ish, when we build the structure surrounding UmbracoElementMixin and our own controller support!

## How to test

1. Go to `performance-profiling.handlers.ts` and make it return an error like this:
```ts
return res(ctx.status(400), ctx.json({ title: 'API ERROR', detail: 'Something went wrong' }));
```
2. Go to Settings -> Profiling dashboard
3. Verify that an error notification is shown


nielslyngsoe (commented):
Based on reading the code it looks good. 👍

nielslyngsoe (approved):
Good for now, though this is not how we want to solve controllers. But this shows the need of resources.

PR #330, opened 2022-12-19
Feature: Language tree and Workspace scaffold

madsrasmussen:
The PR sets up the scaffold for the language tree and the language workspace.

The PR also makes it possible to register a tree without a custom element.  It will then render a default tree. This will make it faster to set up basic trees that doesn't need any custom markup.


nielslyngsoe (approved):
looks good and works :-)

PR #384, opened 2023-01-09
Convert usage of RxJS BehaviorSubjects into the UniqueBehaviorSubject

nielslyngsoe:
please review https://github.com/umbraco/Umbraco.CMS.Backoffice/pull/380 first.


package.json
```diff
@@ -42,6 +42,7 @@
 		"generate:api": "openapi --input https://raw.githubusercontent.com/umbraco/Umbraco-CMS/v13/dev/src/Umbraco.Cms.Api.Management/OpenApi.json --output src/core/backend-api --postfix Resource --useOptions",
 		"generate:api-dev": "openapi --input http://localhost:9000/umbraco/swagger/v1/swagger.json --output src/core/backend-api --postfix Resource --useOptions",
 		"storybook": "npm run wc-analyze && start-storybook -p 6006",
+		"storybook:build": "npm run wc-analyze && build-storybook",
```
iOvergaard, line 45: This script seems to exist already?

iOvergaard (approved):

package.json
nielslyngsoe: Yes, but the 'existing' one is written in a different format than how we have written the name of the other scripts. As we write it like topic:action

So I just found it naturally that you could write storybook:build for that script. But we might remove the old one?

PR #401, opened 2023-01-11
render method on extension slot

nielslyngsoe:
Enable to parse a renderMethod to the umb-extension-slot.

This example shows how we can wrap or control the template rendered for each extension:

```
	public myExtensionWrapperMethod = (manifest: any) => {
		return html`<bla>${manifest.component}</bla>`;
	};

	render() {
		return html`
			<umb-extension-slot id="apps" type="headerApp" .renderMethod=${this.myExtensionWrapperMethod}>
			</umb-extension-slot>
		`;
	}
```


madsrasmussen:
I think we also need to make it possible to somehow wrap all extensions in some custom markup? Let's say I have a group of extensions that I want to render as a grid. Then I need to control the container around the items.

madsrasmussen (approved):

PR #406, opened 2023-01-12
Adds padding/breathing room to dashboard content like we have today

warrenbuckley:
## Before
image

## After
image


nielslyngsoe:
Hi @warrenbuckley we did this initially, but we don't know if all dashboards want to be presented this way. So we are looking for other ways to get this in a good way.

Feel free to add margin to all dashboards, as we already have done on some workspaces / collection views etc.

warrenbuckley:
@nielslyngsoe I thought it would be OK to do this, as the current backoffice all dashboards are given padding/margin.
I can update each one manually, but its more code replicated to maintain in several places rather than one.

Have we had a use case or seen a place so far where the dashboard required zero margin?

Also, when developers from the community build their own dashboards, they will have to add the same padding in as us, to keep it looking consistent.

iOvergaard:
@nielslyngsoe and @warrenbuckley - did you come to a conclusion on padding here?

warrenbuckley:
@iOvergaard nope not heard anything back - this is not critical PR but happy to do the work either way

nielslyngsoe:
@warrenbuckley I guess we can add the padding pr default for dashboards, we can always change it later.. cool.. i will approve.

nielslyngsoe (approved):

PR #407, opened 2023-01-12
Updated Welcome Dashboard in Settings section

warrenbuckley:
Updated HTML for settings dashboard to use a grid of uui-box's to be more visually appealing.

Before this update, it was a very plain list of short descriptions and links, hopefully with a card/uui-box each in the grid with CTA buttons, makes it more visual appealing and likely people to use the links available to them.

## Before
image

## After
image


iOvergaard (approved):
LGTM

PR #410, opened 2023-01-13
Change order of sidebar-menu-items aka Trees in Settings section

warrenbuckley:
Change sidebar-menu-item manifest for Setting trees to match same order as current backoffice

## Note
I moved the Member Types menu-item into the Settings section as before.


madsrasmussen (approved):

madsrasmussen:
Thanks! This one looks good too 😃

PR #416, opened 2023-01-13
Do not use Context Alias for Unique for Consumers

nielslyngsoe:
Do not use Context Alias for Unique for Consumers
As we may have multiple different consumers consuming the same thing. Example Multiple Resource controllers might consume a NotiicationContext each.
Using the alias as unique would make each of these replace the other as we do not want that.


iOvergaard (approved):
LGTM

PR #422, opened 2023-01-16
Feature/dashboard health check

loivsen:


src/backoffice/settings/dashboards/health-check/views/health-check-group.element.ts
```diff
@@ -0,0 +1,204 @@
...
+	private _checks?: HealthCheckWithResult[] | null;
+
+	@state()
+	private _keyResults?: any;
+
+	private _api?: UmbHealthCheckContext;
+
+	constructor() {
+		super();
+		this.consumeContext<UmbHealthCheckDashboardContext>('umbHealthCheckDashboard', (instance) => {
```
iOvergaard, line 86: We need an `UmbContextToken` for this context, which it can export itself in the same file where the dashboard context is being created.

```ts
export const UMB_HEALTHCHECK_DASHBOARD_CONTEXT_TOKEN = new UmbContextToken<UmbHealthCheckDashboardContext>('umbHealthCheckDashboard');
```

src/backoffice/settings/dashboards/health-check/views/health-check-group.element.ts
```diff
@@ -0,0 +1,204 @@
...
+			this._healthCheckContext = instance;
+
+			this._api = this._healthCheckContext?.apis.get(this.groupName);
+
+			this._api?.getGroupChecks(this.groupName);
```
iOvergaard, line 91: It looks like we not only _get_ the checks but actually _perform_ the checks when you click on a group:

image

Is there a way to rectify that where we still show the checks but do not perform the results before the button "Check" is clicked?

iOvergaard: P.S. We only show the checks on the second run, when you have clicked the button anyway.

src/backoffice/settings/dashboards/health-check/views/health-check-group.element.ts
```diff
@@ -0,0 +1,204 @@
...
+			return html`
+				<div class="header">
+					<h2>${this._group.name}</h2>
+					<uui-button
+						label="Get checks"
+						color="positive"
+						look="primary"
+						.state="${this._buttonState}"
+						@click="${this._buttonHandler}">
+						Get checks
```
iOvergaard, line 125: Should this be called "Perform checks" or something? I don't remember the name in the old backoffice.

src/backoffice/settings/dashboards/health-check/views/health-check-group.element.ts
```diff
@@ -0,0 +1,204 @@
...
+
+	renderCheckResults(key: string) {
+		const checkResults = this._keyResults?.find((result: any) => result.key === key);
+		return html`<uui-icon-registry-essential>
+			<div class="check-results-wrapper">
+				${checkResults?.results.map((result: any) => {
+					return html`<div class="check-result">
+						<div class="check-result-description">
+							<span>${this.renderIcon(result.resultType)}</span>
+							<p>${result.message}</p>
```
iOvergaard, line 148: Running this on the actual backend reveals, that the messages may contain raw HTML, which we do not render by default. The question is if we should render it (trust the server) or remove it from the server?

image

iOvergaard (changes requested):
Generally looks good. I think we can improve the context class a bit, but not important for now (maybe we need to update the backend a bit later on). I was only able to find a few weirdnesses in the group element 👍

src/backoffice/settings/dashboards/health-check/views/health-check-group.element.ts
loivsen: Yeah me and @madsrasmussen discussed that when opening a specific group directly via anURL, forexample /health-check/configuration, we wanted to load only the data from configuration.
But there is currently only two APIs available: either we get all checks / data from all the groups, but with no check.results performed, or get a single group with check.results performed.
So the dashboard is currently kind of mimicking the behavior we wanted by deleting the check.results (see health-check.context.ts)
That means that we expecting to have additional api in there future  that can get the group check, without actually performing the check.results.

Another option would be to get all groups without the check.results and only use the part we need for configuration, etc.

src/backoffice/settings/dashboards/health-check/views/health-check-group.element.ts
loivsen: I think its depends on if the server contains anything important HTML wise. Personally I think its best if we can avoid it from the server.

src/backoffice/settings/dashboards/health-check/views/health-check-group.element.ts
madsrasmussen: Correct, this was a workaround to test an alternativ data structure. The current endpoints did too much at the same time. We need to talk to the backend team about the changes. I just totally forgot about it.

src/backoffice/settings/dashboards/health-check/views/health-check-group.element.ts
iOvergaard: I just had a look at the other messages from the server. They contain some tokens that might be valuable to render appropriately. In truth, we have a lot of HTML in the translation files which we can't get rid of. I fear we have no choice but to trust the server and render it.

src/backoffice/settings/dashboards/health-check/views/health-check-group.element.ts
iOvergaard: I think we need to run it by @elit0451 then. It would be good to have an endpoint that just fetches the checks and another that performs the checks and returns the results.

src/backoffice/settings/dashboards/health-check/views/health-check-group.element.ts
loivsen: It was called "Get group" but I like "Perform checks" a lot better.

iOvergaard:
LGTM

iOvergaard:
are you still working on it @madsrasmussen or ready to merge?

iOvergaard (approved):

PR #426, opened 2023-01-17
UmbContextAlias

iOvergaard:
Before:

```ts
this.consumeContext('some-context-alias', (_instance) => {
  // _instance type unknown
});
```

After:
```ts
// My Store
export class MyStore {}

export const MY_STORE_CONTEXT_ALIAS = new UmbContextToken<MyStore>(MyStore.name);
```

```ts
// Element
import { MY_STORE_CONTEXT_ALIAS } from './my-store';

this.consumeContext(MY_STORE_CONTEXT_ALIAS, (_instance) => {
  // _instance type MyStore
});
```

We can still use strings (magic strings) to provide and consume. The `UmbContextToken` works with strings in the background (using the alias provided as the first attribute when new'ing it up), so it has a nice fallback if you didn't update your code yet. 

`UmbContextToken` is solely a DX offer to ourselves and package developers to hit up the globals we are providing in `UmbBackofficeElement`.


madsrasmussen:
It looks really good 🥳 

The only thing I noticed is in some places we base our context alias on a hardcoded string that is camelcased: 'umbSomethingStore', and in some other places we use this class name UmbModalService. It will result in contexts being provided under both camelcased and pascal cased aliases. I think it could confuse a bit if someone is using the string alias. Would it be possible to align these so we do it the same way on all cases?

These are could easily look up on the backoffice element as an example:
Screenshot 2023-01-19 at 10 38 42



I also really liked the token name we talked about. Would it make sense to call it a ContextToken? Then we could keep alias's as strings?

iOvergaard:
@madsrasmussen Thank you so much! I'll align the aliases. And I agree, ContextToken is a better description :)

madsrasmussen (approved):

PR #429, opened 2023-01-20
Language Workspace

JesmoDev:
Added the language overview and workspace


src/backoffice/settings/languages/workspace/language-root/language-root-table-delete-column-layout.element.ts
```diff
@@ -0,0 +1,70 @@
...
+		this.consumeContext(UMB_MODAL_SERVICE_CONTEXT_TOKEN, (instance) => {
+			this._modalService = instance;
+		});
+	}
+
+	private _handleDelete(event: MouseEvent) {
+		event.stopImmediatePropagation();
+		if (!this._languageStore) return;
+
+		const modalHandler = this._modalService?.confirm({
```
iOvergaard, line 34: Should we consider making a more generic 'confirm' dialog with an optional parameter to customise the text?

src/backoffice/settings/languages/workspace/language-root/language-root-table-name-column-layout.element.ts
```diff
@@ -0,0 +1,23 @@
+import { UUITextStyles } from '@umbraco-ui/uui-css';
+import { css, html, LitElement, nothing } from 'lit';
+import { customElement, property } from 'lit/decorators.js';
+
+@customElement('umb-language-root-table-name-column-layout')
+export class UmbLanguageRootTableNameColumnLayoutElement extends LitElement {
+	static styles = [UUITextStyles, css``];
+
+	@property({ attribute: false })
+	value!: any;
```
iOvergaard, line 10: If you don't have the type available at the moment, you could always do 
```
value: { key: string, name: string };
```

```suggestion
  value!: { key: string, name: string };
```

src/backoffice/settings/languages/workspace/language/language-workspace.context.ts
```diff
@@ -0,0 +1,72 @@
+import { UmbLanguageStore, UmbLanguageStoreItemType } from '../../language.store';
+import type { LanguageDetails } from '@umbraco-cms/models';
+import { UmbControllerHostInterface } from 'src/core/controller/controller-host.mixin';
```
iOvergaard, line 3: Please use alias paths instead and go through the files with similar paths.

```suggestion
import { UmbControllerHostInterface } from '@umbraco-cms/controller';
```

src/backoffice/settings/languages/workspace/language/language-workspace.context.ts
```diff
@@ -0,0 +1,72 @@
...
+	constructor(host: UmbControllerHostInterface, entityKey: string) {
+		this.host = host;
+		this._entityKey = entityKey;
+
+		this._data = new UniqueBehaviorSubject<LanguageDetails>(DefaultLanguageData);
+		this.data = this._data.asObservable();
+
+		new UmbContextConsumerController(host, 'umbLanguageStore', (_instance: UmbLanguageStore) => {
+			this._store = _instance;
+			if (!this._store) {
```
iOvergaard, line 37: Not sure I understand what this is trying to achieve. Are you unsure if you get the store here or something undefined instead? In theory, this callback should only be run if there is an actual store by that name.

src/backoffice/settings/languages/workspace/language/language-workspace.context.ts
```diff
@@ -0,0 +1,72 @@
...
+	protected _storeObserver?: UmbObserverController<LanguageDetails>;
+
+	constructor(host: UmbControllerHostInterface, entityKey: string) {
+		this.host = host;
+		this._entityKey = entityKey;
+
+		this._data = new UniqueBehaviorSubject<LanguageDetails>(DefaultLanguageData);
+		this.data = this._data.asObservable();
+
+		new UmbContextConsumerController(host, 'umbLanguageStore', (_instance: UmbLanguageStore) => {
```
iOvergaard, line 35: Please use the STORE_ALIAS here, or the token.

```suggestion
		new UmbContextConsumerController(host, STORE_ALIAS, (_instance) => {
```

src/backoffice/settings/languages/workspace/language/language-workspace.element.ts
```diff
@@ -1,19 +1,79 @@
...
-import { UMB_NOTIFICATION_SERVICE_CONTEXT_TOKEN } from 'src/core/notification';
-
+import { UUITextStyles } from '@umbraco-ui/uui-css';
+import { css, html, nothing } from 'lit';
+import { customElement, property } from 'lit/decorators.js';
+import { UUIInputElement, UUIInputEvent } from '@umbraco-ui/uui';
+import { UmbWorkspaceLanguageContext } from './language-workspace.context';
 import { UmbLitElement } from '@umbraco-cms/element';
+import type { LanguageDetails } from '@umbraco-cms/models';
+import 'src/backoffice/shared/components/workspace/actions/save/workspace-action-node-save.element.ts';
```
iOvergaard, line 8: Use a relative path instead.

iOvergaard (changes requested):
Overall looks good. I think some of my comments are related to existing stuff since it seems like you copied your store from another store, but please see if we can get rid of some tech debt already, which could then be applied back to the pre-existing stores :)

iOvergaard (approved):

PR #430, opened 2023-01-20
Type specific Subjects + Tests of these

nielslyngsoe:
A Subject for Objects and one for Arrays and methods to update/append the data of these.

Notice the Array has the ability to set a uniqueCompare method, which will ensure that appended data replaces existing data if matched.


iOvergaard (approved):
LGTM

PR #433, opened 2023-01-23
Feature: Multiple Text String Property Editor

madsrasmussen:
The PR adds the functionality for the Multiple Text String property editor. 

* The Editor supports adding and deleting items
* Updating values.
* It supports a disabled state
* It supports a readonly state
* It supports the min and max configuration from the data type.

https://user-images.githubusercontent.com/6078361/214395875-7b43745d-5236-4be1-ae93-66edf4095df0.mp4

I ran into a couple of challenges with validation during development:

* I had to stop valid and invalid events from bubbling to prevent triggering errors further up the chain.
* I also had a challenge with the required validation because the multiple-text-string-input doesn't use the value property but a custom items property. We will need a way to swap out the default missingValue validator.
* If a field has been marked as invalid and the validation message has bubbled to an outer message container. The message doesn't get cleared if the invalid element is removed from the dom. (see the end of the video).


iOvergaard (approved):
LGTM but should we try and put the events into a library and use an alias?

```
import { UmbChangeEvent, UmbInputEvent, UmbDeleteEvent } from '../../../../../../core/events';
```

madsrasmussen:
Ah, yeah. Good catch. I couldn't figure out how many of these "global" events we would get, so I din't know what the best home for them was.

I will merge this now, then we can add the lib in a new PR.

PR #438, opened 2023-01-24
Feature/auth app

iOvergaard:
Migrate the login element to a standalone vite app, so that the CMS can use it to show the login page on the server, but we still retain access to the entire frontend API such as the context and all the shared stuff.

In the future, we might envision a more complete mono-repo where each app and library have a more isolated build, test and lint environment. For now, the main tsconfig.json file includes all source files, but the auth app also has its own tsconfig to be able to spin up a local vite server to test that individual component. It will also only be built as a library (it uses vite's library mode) and it works well.

As for linting and testing, the auth app is also included in the global eslint and web-test-runner, but it has its own Playwright config for e2e testing since Playwright needs to know what server to spin up beforehand.


nielslyngsoe:
Code looks good, of to test it.

nielslyngsoe (approved):

PR #442, opened 2023-01-25
Create method to enable Contexts to be initialised with a host reference

nielslyngsoe:
enables to parse a method to the context-provider-element instead of the value.

` .create=${host => new UmbDocumentTypeDetailStore(host)`


iOvergaard (approved):
LGTM!

PR #491, opened 2023-01-31
Feature: Repository Pattern + Templates Scaffold

madsrasmussen:
This PR adds a proof of concept for the repository pattern. It uses the Template Tree and Workspace as the example.

Currently most data communication for a silo has gone through a store. There has been a couple of challenges with this. These are some of the headlines I have tried to solve:

### Data from different sources:
We want to lay the foundation that data for a silo can come from different sources depending on the state of the app. It can be from a server, an offline database, a store, a Signal-R connection, etc.

That means we need one entry point, a repository, for elements (and contexts) to work with data, and then it is up to this repository to figure out how it will retrieve the data.

**The communication will look like this:**
A repository can be called directly from an element, but will often be instantiated in a context controller, like the Workspace Context Controller.

Frame 1

### Requesting data from the wrong context
Requesting data through a Store gives det wrong context. A store lives in the global context. Any notifications (or other context aware dependencies) will always show up in the global context. We might have scenarios (infinite editing, tree action menu) etc., that need to show a notification elsewhere. A repository is instantiated in the context it is used. We then get the correct local context and can show notifications in the correct place.

### Subscribing to updates in the right places
We need to be aware of subscribing to data updates (ex: trough a Store) at the right time. We don't want to accidentally overwrite data that is currently being edited by a user. 

Ex. when we start a new workspace, we will retrieve data for the entity and create a new observable of that data. All elements inside that workspace can be notified about updates to the local observable. When we save the data it will be sent to the sever (maybe put in a store). The workspace and local data will be destroyed when navigating away.

Optionally we can have a subscription to the Store, and get notified if anything changes on the same piece of data elsewhere. We can then notify the user about the changes so they can decide how to handle it. The data could have been deleted at the same time they were editing it.

### Async element controllers
We start to see a pattern where many of our classes depend on the host element, to request contexts. If they expose any methods that rely on any of these dependencies, the code will break if it hasn't been received when the method is called.

I have currently solved this by making all methods async. Then the class can ensure that a dependency is ready before any method internals are run. It gave me another challenge with observables. Sometimes we want to return an observable when requesting a piece of data. I have currently solved this by letting the promise return an observable as part of the response instead. That means requesting data would look something like this:

```javascript
const myRepo = new MyRepo(this);
const  { data, updates } = await myRepo.getItems();

// If I don't want updates, I can use the items like this:
console.log(data.items);

// I can get notified about new/updated items like this:
updates.subscribe((items) => console.log(items));
```


nielslyngsoe (approved):

mattbrailsford:
Just a bit late to this, but I'm wondering if `Repository` is not the right terminology here? In OOP a repository is usually responsible for the access of a single entity type, where as a `Store` can usually access other related entities.

I wonder if better terminology is a `Facade` https://en.wikipedia.org/wiki/Facade_pattern or something similar?

PR #492, opened 2023-02-01
Language workspace v2

JesmoDev:
## Description
Further work on language workspace.

- Splits up current languages and available languages
- You can no longer have more than one language of the same culture
- no more bulk actions on languages
- correctly implements default language (only one, and always one)
- adds warning when changing culture
- cleans up components and use native private accessor (`#name `instead of `private _name`) wherever possible


src/core/mocks/domains/language.handlers.ts
```diff
@@ -22,6 +22,11 @@ export const handlers = [
 		return res(ctx.status(200), ctx.json(response));
 	}),
 
+	rest.get(umbracoPath('/languages'), (req, res, ctx) => {
```
iOvergaard, line 25: This endpoint doesn't seem to exist on the language controller. What is it trying to do?

madsrasmussen: I have updated the code to use the correct culture resource. I will get this PR merged.

madsrasmussen (approved):

PR #503, opened 2023-02-06
Feature/improve extensions workspace

warrenbuckley:
* Sorts extensions by type alphabetically
* Sorts those extensions with same type by weight
* Updates unregister UUI button styling to danger & use trash icon
* Adds weight column
* For any 'custom extension' it appends the name of the extension
* Adds a confirmation dialog to the unregister

image

### DIFF
--- a/libs/extensions-api/registry/extension.registry.ts
+++ b/libs/extensions-api/registry/extension.registry.ts
@@ -80,4 +80,19 @@ export class UmbExtensionRegistry {
 			)
 		) as Observable<Array<ExtensionType>>;
 	}
+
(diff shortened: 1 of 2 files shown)


madsrasmussen:
Hi @warrenbuckley. Thats cool. Would it be possible to add i filter on the type too?

warrenbuckley:
@madsrasmussen when you say filter. Do you mean like a search/filter box?
If so, do we have a current implementation anywhere else I could use as point of reference?

madsrasmussen (approved):

PR #515, opened 2023-02-12
Property editor: Radio button list

JesmoDev:
## Description

Adds a radio button list property editor
## Types of changes

## Checklist


src/core/mocks/data/data-type.data.ts
```diff
@@ -277,7 +277,15 @@ export const data: Array<DataTypeDetails> = [
 		isFolder: false,
 		propertyEditorModelAlias: 'Umbraco.RadioButtonList',
 		propertyEditorUIAlias: 'Umb.PropertyEditorUI.RadioButtonList',
-		data: [],
+		data: [
+			{
+				alias: 'itemList',
+				value: [
+					{ label: 'Label 1', key: '123' },
```
iOvergaard, line 284: Looks like the old backoffice uses "value" instead of "key", but "label" is fine

iOvergaard (changes requested):
Code looks good, but we need to reflect the old backoffice in terms of options supported and the names of the old data structure, i.e. "items" instead of "itemList". 

We probably don't have a complete overview of all supported options, but try and spin up the old backoffice and see. I know there is a "useLabel" option on some of them for instance.

iOvergaard:
It looks like this property editor was modeled after the checkbox list, which also seems to have the same inherent problems. Would you fix that property editor as well, please?

Prevalues:
```js
items: {
  0: { sortOrder: 0, value: 'Whatever' },
  1: { sortOrder: 10, value: 'Whatever again' },
  2: { sortOrder: 5, value: "Look at me, I'm special"},
}
```

Values (what is saved on change and accessible through $scope.model.value):
```js
// Checkbox list (supports multiple)
$scope.model.value = [ 'Whatever', 'Whatever again' ]

// Radiobutton list (only 1 value at a time)
$scope.model.value = 'Whatever'
```

iOvergaard (approved):

PR #528, opened 2023-02-16
Feature/property editor multi url picker

loivsen:
Multi url picker together with a link picker as a modal window
Also fixed the config values of color picker


libs/utils/udi-service.ts
```diff
@@ -0,0 +1,3 @@
+export function buildUdi(entityType: string, guid: string) {
+	return 'umb://' + entityType + '/' + guid.replace(/-/g, '');
```
iOvergaard, line 2: Consider using template literals for readability

```suggestion
	return `umb://${entityType}/${guid.replace(/-/g, '')}`;
```

src/backoffice/shared/components/input-color-picker/input-color-picker.element.ts
```diff
@@ -1,19 +1,19 @@
...
+@customElement('umb-input-color-picker')
+export class UmbInputColorPickerElement extends FormControlMixin(UmbLitElement) {
 	static styles = [UUITextStyles];
 
 	@property({ type: Boolean })
 	showLabels = false;
 
 	@property()
-	colors?: string[];
+	swatches?: any[];
```
iOvergaard, line 16: We need a model for these somewhere, maybe @umbraco-cms/models ?

src/backoffice/shared/components/input-multi-url-picker/input-multi-url-picker.element.ts
```diff
@@ -0,0 +1,164 @@
...
+	min?: number;
+
+	/**
+	 * Min validation message.
+	 * @type {boolean}
+	 * @attr
+	 * @default
+	 */
+	@property({ type: String, attribute: 'min-message' })
+	minMessage = 'This field need more items';
```
iOvergaard, line 49:
```suggestion
	minMessage = 'This field needs more items';
```

src/core/modal/layouts/link-picker/modal-layout-link-picker.element.ts
```diff
@@ -0,0 +1,223 @@
...
+	connectedCallback() {
+		super.connectedCallback();
+		this._link.icon = this.data?.icon;
+		this._link.name = this.data?.name;
+		this._link.published = this.data?.published ?? true;
+		this._link.queryString = this.data?.queryString;
+		this._link.target = this.data?.target;
+		this._link.trashed = this.data?.trashed ?? false;
+		this._link.udi = this.data?.udi;
+		this._link.url = this.data?.url;
```
iOvergaard, line 102: I think we can make one call to `this._link` to set it initially instead of a new call per line.

iOvergaard (changes requested):
I like the direction this is going in. You have left a few TODOs in your code and I left a few comments ☺️

src/backoffice/shared/components/input-multi-url-picker/input-multi-url-picker.element.ts
```diff
@@ -0,0 +1,164 @@
...
+	@property()
+	ignoreUserStartNodes?: boolean;
+
+	/**
+	 * @type {"small" | "medium" | "large"}
+	 * @attr
+	 * @default "small"
+	 */
+	@property()
+	overlaySize?: 'small' | 'medium' | 'large' | 'full';
```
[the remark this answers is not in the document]
loivsen: You are right that we do. I really thought we didnt, I corrected it now.

src/core/modal/layouts/link-picker/modal-layout-link-picker.element.ts
loivsen: The issue was i wanted to keep config and data separate in this file, but the data are sent together with config into `this.data`.

I think I found a better better way though

loivsen:
@iOvergaard thanks for the feedback so far.
Are you able look into `input-multi-url-picker.element.ts` again? Its regarding value property for this file. I still am unsure how we should save all this data in a good way, so that `property-editor-ui-multi-url-picker.element.ts` can use it in the change event, similar to how the other editors do it (see how color picker, media picker, document picker do it)

iOvergaard (approved):
LGTM

PR #532, opened 2023-02-20
Storybook: Sets Getting Started as default story

warrenbuckley:
Configures StoryBook to have a custom order of stories, just for the 'Guides' section so that the 'Getting Started' guide loads first when visiting/viewing the storybook.

image

## Docs from StoryBook
https://storybook.js.org/docs/react/writing-stories/naming-components-and-hierarchy#sorting-stories


iOvergaard (approved):
LGTM!

PR #537, opened 2023-02-21
Adds Basic Modal to ModalService

warrenbuckley:
## What
Adds a new method to the ModalService for `openBasic()` that takes the following three properties to allow displaying simple modals in a sidebar that are read-only information and need some simple HTML.
* header 
* content
* overlaySize

## Why
Package developers and implementors sometimes want to display simple data or information in a sidebar. This would save them on creating a specific new HTML WebComponent for the layout view and allows them to pass the Lit HTML content for the header and the main content of the modal.

## Testing
* The `<umb-debug>` component PR I wrote previously has had the specific dialog layout component deleted and now passes its HTML into modalservice call of `openBasic()`

image
image


warrenbuckley:
Any thoughts @iOvergaard ?

iOvergaard (approved):
LGTM and a nice feature!

PR #546, opened 2023-02-22
workspace info

nielslyngsoe:
TBD


src/backoffice/shared/components/history-ui-maker/history-ui-maker.element.ts
```diff
@@ -0,0 +1,160 @@
...
+export interface logItem {
+	userName: string;
+	timestamp: string;
+	logType: logType;
+	parameters: string;
+}
+
+export type logType = 'Save' | 'Publish' | 'ContentVersionPreventCleanup' | 'ContentVersionEnableCleanup';
+
+@customElement('umb-history-ui-maker')
```
nielslyngsoe, line 15: I would name this component `umb-history-list`

where each entry is a component on its own. called something like `umb-history-item`.


Looks good :-)

madsrasmussen: I have updated the names. Then we can get this merged.

It looks really good 🥇

PR #556, opened 2023-02-27
Feature: Entity Action + Workspace libs

madsrasmussen:
This PR makes an Entity Action and Workspace lib. That makes it possible for packages to access base classes for Entity Actions, Entity Bulk Actions and Workspace Actions.

As part of the PR, I have set up two new aliases:

**@umbraco-cms/entity-action**:  Will be used to import utils for Enity Actions, Entity Bulk Actions and defaul action API's like `UmbDeleteEntityAction`

**@umbraco-cms/workspace**:  Will be used to import utils for everything related to workspaces. I have moved the files related to a Workspace Action, but more utils will follow.

As part of the PR, I have removed the need to pass a repository to a Workspace Action. Almost all workspaces already uses a repository, so I have aligned the interface. All workspaces now exposes their repository. We can now use this in a Workspace Action.


iOvergaard (approved):
LGTM

PR #559, opened 2023-02-28
Feature: Section Sidebar App extension point

madsrasmussen:
This PR adds a new extension points for Section Sidebar Apps. It makes it possible to compose the Section Sidebar of many different UI elements. 

section-sidebar-item

This means that a sidebar should be able to end up something like this:
section-sidebar-items

In our case we want to add an "Application Language select element" only in the Content section sidebar.

**Known issues:**
The sidebar is not 100% as flexible as I want it to be. It is currently only possible to add sidebar items on top of the navigation. Our menu routing is too tightly coupled to the section, but it needs to be fixed in another PR. All current code should still works as expected.

What to test:
* Test that the app language select is only rendered in the Content Section


src/backoffice/media/section.manifests.ts
```diff
@@ -29,15 +29,16 @@ const dashboards: Array<ManifestDashboardCollection> = [
...
-	name: 'Media Sidebar Menu',
+const sectionSidebarMenu: ManifestSectionSidebarMenu = {
+	type: 'sectionSidebarMenu',
+	alias: 'Umb.SectionSidebarMenu.Media',
+	name: 'Media Section Sidebar Menu',
 	weight: 100,
 	meta: {
 		label: 'Media',
 		sections: [sectionAlias],
+		menu: 'Umb.Menu.Media',
```
nielslyngsoe, line 40: I think this one might be a bit hard to understand? If I understood correctly its a aias for the menu items to bind to?

Totally unconcidered though, could it be the alias, and there by be eliminated? aka. 'Umb.SectionSidebarMenu.Media'

Or would that be fragile?

nielslyngsoe: Ah, I realized this is something that goes into the menu. I'll get in touch.

nielslyngsoe (approved):
Code is good, And I have clicked around a bit, and it seems alright :-)

PR #560, opened 2023-02-28
Move the WebComponent/LitElement that was inside the story file to ow…

warrenbuckley:
Move the WebComponent/LitElement that was inside the story file to own file
ESBuild errors was not happy it was inside a story file

@iOvergaard can you test & verify this is all bellissimo now! 

Also a thought does it make sense to update the Github CI actions to ensure no error is given when doing `npm run storybook:build`


iOvergaard (approved):
Well, it builds. That's important. Let's see how it looks like on the static website when we merge 🗡️

PR #563, opened 2023-03-01
Bug/Rename notification service

madsrasmussen:
Rename `UmbNotificationService` to `UmbNotificationContext`.

To make it clear that a notification can show up in different places depending on the context. We have renamed the service.


iOvergaard (approved):
LGTM

PR #571, opened 2023-03-01
Bugfix: Rename Modal Service

madsrasmussen:
Rename `UmbModalService` to `UmbModalContext`.

To make it clear that a modal can show up in different places depending on the context. We have renamed the service.


iOvergaard (approved):
LGTM

PR #574, opened 2023-03-02
Feature/server packages v2

iOvergaard:
A lot of updates to make server packages work, including changes to the way we load entrypoints, where we now support an onInit function, and some changes to the build script for the CMS target.

Also updated all tokens to use strings instead of runtime classname, which otherwise gets mangled by the minifier so context-api didn't really work for extensions.


libs/extensions-api/registry/extension.registry.ts
```diff
@@ -13,7 +15,7 @@ export class UmbExtensionRegistry {
 	private _extensions = new BehaviorSubject<Array<ManifestBase>>([]);
 	public readonly extensions = this._extensions.asObservable();
 
-	register(manifest: ManifestTypes): void {
+	register(manifest: ManifestTypes, rootHost?: UmbControllerHostInterface): void {
```
madsrasmussen, line 18: We should probably make the host a part of the constructor, but we can change that when we provide it trough the context API instead.

madsrasmussen (approved):

madsrasmussen:
Looks good to me

PR #575, opened 2023-03-02
Feature: Store Extension Point

madsrasmussen:
The PR adds a new extension point for stores and tree stores.

The PR goes through all areas and registers the stores. This makes it possible to remove the hardcoded stores from the Backoffice Element. The rest of the hardcoded contexts should be registered trough each areas entry points.


src/backoffice/backoffice.element.ts
```diff
@@ -83,45 +57,20 @@ export class UmbBackofficeElement extends UmbLitElement {
 	constructor() {
 		super();
 
+		// TODO: find a way this is possible outside this element.
 		this.provideContext(UMB_MODAL_CONTEXT_TOKEN, new UmbModalContext());
 		this.provideContext(UMB_NOTIFICATION_CONTEXT_TOKEN, new UmbNotificationContext());
-
-		// TODO: find a way this is possible outside this element. It needs to be possible to register stores in extensions
 		this.provideContext(UMB_CURRENT_USER_STORE_CONTEXT_TOKEN, new UmbCurrentUserStore());
```
iOvergaard, line 63: What about this store? Do we need to new it up here because it doesn't extend StoreBase?

src/backoffice/backoffice.element.ts
```diff
@@ -83,45 +57,20 @@ export class UmbBackofficeElement extends UmbLitElement {
...
-		new UmbTemplateDetailStore(this);
-		new UmbLanguageStore(this);
-
 		this.provideContext(UMB_APP_LANGUAGE_CONTEXT_TOKEN, new UmbAppLanguageContext(this));
 		this.provideContext(UMB_BACKOFFICE_CONTEXT_TOKEN, new UmbBackofficeContext());
-		this.provideContext(UMB_CURRENT_USER_HISTORY_STORE_CONTEXT_TOKEN, new UmbCurrentUserHistoryStore());
 		new UmbThemeContext(this);
-
-		new UmbPackageStore(this);
 		new UmbServerExtensionController(this, umbExtensionsRegistry);
```
iOvergaard, line 67: Same as CurrentUserStore?

src/backoffice/backoffice.element.ts
```diff
@@ -83,45 +57,20 @@ export class UmbBackofficeElement extends UmbLitElement {
...
+		this.provideContext(UMB_CURRENT_USER_HISTORY_STORE_CONTEXT_TOKEN, new UmbCurrentUserHistoryStore());
+
+		// Register All Stores
+		this.observe(umbExtensionsRegistry.extensionsOfTypes(['store', 'treeStore']), (stores) => {
```
iOvergaard, line 71: Should we add a mechanism for "scope" of store, so you decide if it's registered globally, or do you just not add a manifest for a store, if you want to spin it up elsewhere?

iOvergaard (dismissed):
I like the concept. Just a few questions and changes!

src/backoffice/members/members/repository/member.store.ts
```diff
@@ -0,0 +1,29 @@
...
+	append(member: MemberDetails) {
+		this.#data.append([member]);
+	}
+
+	remove(uniques: string[]) {
+		this.#data.remove(uniques);
+	}
+}
+
+export const UMB_MEMBER_STORE_CONTEXT_TOKEN = new UmbContextToken<UmbMemberStore>(UmbMemberStore.name);
```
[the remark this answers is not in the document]
madsrasmussen: This is updated now 👍

src/backoffice/members/members/repository/member.store.ts
```diff
@@ -0,0 +1,29 @@
...
+ * @export
+ * @class UmbMemberStore
+ * @extends {UmbStoreBase}
+ * @description - Data Store for Members
+ */
+export class UmbMemberStore extends UmbStoreBase {
+	#data = new ArrayState<MemberDetails>([], (x) => x.key);
+
+	constructor(host: UmbControllerHostInterface) {
+		super(host, UmbMemberStore.name);
```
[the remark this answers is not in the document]
madsrasmussen: This is updated now 👍

src/backoffice/backoffice.element.ts
madsrasmussen: I think it is ok to assume a store is always globally available. At least I don't think it is something we need to worry about right now.

src/backoffice/backoffice.element.ts
madsrasmussen: I left a couple of "special stores" and other globally registered contexts. Some if them we might come back to and register en the registry other context might "just" be registered as part of each areas entry point. 

I think we should do a second round of Backoffice Element cleanup. I just needed your packages/entry point work to be included to get the full overview.

nielslyngsoe: Sounds like awaiting an entry point extension point.

src/backoffice/backoffice.element.ts
nielslyngsoe: Sounds good to me.

nielslyngsoe (approved):

PR #580, opened 2023-03-08
Feature: Modal extension point

madsrasmussen:
### Register in the extension registry

The manifest

```json
{
	"type": "modal",
	"alias": "My.Modal",
	"name": "My Modal",
	"js": "../path/to/my-modal.element.js"
}
```

### Create a `Modal Token`

A modal token is a string that identifies a modal. It should be the modal extension alias. It is used to open a modal and also to set default options for the modal. The options can be overwritten when opening a modal.

```ts
interface MyModalData = {
	headline: string;
	content: string;
}

interface MyModalResult = {
	myReturnData: string;
}

const MY_MODAL_TOKEN = new ModalToken<MyModalData, MyModalResult>('My.Modal', {
	type: 'sidebar',
	size: 'small'
});
```

### Open a modal

A modal is opened by calling the `open` method on the `UmbModalContext`. The methods will accept a `Modal Token` (or extension alias), an optional dataset, and optional modal options. It returns an instance of `UmbModalHandler`.

```ts
import { html, LitElement } from 'lit';
import { UmbElementMixin } from '@umbraco-cms/element';
import { UmbModalContext, UMB_MODAL_CONTEXT_ALIAS } from '@umbraco-cms/modal';
import { MY_MODAL_TOKEN } from '../path/to/my-modal-token.js';

class MyElement extends UmbElementMixin(LitElement) {
	#modalContext?: UmbModalContext;

	constructor() {
		super();
		this.consumeContext(UMB_MODAL_CONTEXT_ALIAS, (instance) => {
			this.#modalContext = instance;
			// modalContext is now ready to be used
		});
	}

	#onClick() {
		const data = {'data goes here'};
		const options {'options go here'};
		const modalHandler = this.#modalContext?.open(MY_MODAL_TOKEN), data, options);

		modalHandler?.onSubmit().then((data) => {
			// run logic when modal is submitted
		}, () => {
			// run logic when modal is closed
		});
	}

	render() {
		return html`<button @click=${this.#onClick}>Open modal</button>`;
	}
}
```


nielslyngsoe (approved):
fantastic

PR #582, opened 2023-03-09
Adds modal layout for embedded media picker in RTE

nathanwoulfe:
~~Will need to be relocated once #580 is merged, will leave as a draft until then.~~

Adds modal for embedded media in RTE, per updated modal implementation


madsrasmussen (approved):
Looks good to me

PR #587, opened 2023-03-13
Server packages: Notification lib

iOvergaard:
## Changes

1. Move the default layout from "libs/notification" to "src/core/notification" and include it directly in app.ts to prevent it from being bundled with the notification lib (the default layout is always accessible in the browser)
2. Now only the UmbNotificationContext, UmbNotificationHandler, and an interface for the default data are available through the lib with no side-effects


nielslyngsoe (approved):
looks good

PR #588, opened 2023-03-13
Feature: Flexible Workspace Routing

madsrasmussen:


nielslyngsoe (approved):
Fixes the routes, still some issuees but nearly as god as before. so getting this in.

PR #589, opened 2023-03-13
Feature/model libs

iOvergaard:
## Changes
1. Add rollup to model libs
2. Add rollup to store lib
3. Fix one import path


nielslyngsoe (approved):
looks good

PR #590, opened 2023-03-13
Feature: Manifest conditions

madsrasmussen:
Split the manifest meta into "Meta" and "Conditions" to make it more clear what each part does.


nielslyngsoe (approved):
looks good and works

PR #595, opened 2023-03-15
bugfix: entity mock save now accepts an item instead of an array

JesmoDev:
Changes the save function in the mock data entity class to accept a single item instead of an array


madsrasmussen (approved):
👍

PR #597, opened 2023-03-15
make observer re-subscribe on hostConnected if unsubscribed by hostDi…

nielslyngsoe:
Ensures subscriptions only are 'active' when elements are in the DOM.

Specifically, this PR fixes so an element that has been re-appended in the DOM still has/gets a subscription.


iOvergaard (approved):
LGTM

PR #600, opened 2023-03-16
Remove invalid dot from email and URL validation strings

JasonElkin:
## Description

This removes the dot from the default email and URL regexes that can result in bad email addresses/URLs

## Types of changes

## Motivation and context

The default validation allows _any_ character (UTF-8, emoji etc.) that are not valid in TLDs - I've seen a few instances of this being missed in the wild (it's been a problem in the Backoffice for a long time).

The old Backoffice actually defines this regex in code like this: `[a-zA-Z0-9_\.\+-]+@[a-zA-Z0-9-]+\.[a-zA-Z0-9-\.]+` but it's not properly escaped and gets transformed by the gulp build into `[a-zA-Z0-9_.+-]+@[a-zA-Z0-9-]+.[a-zA-Z0-9-.]+`, which is wrong.

## How to test?
Create a new property editor, choose email or URL custom validation.

image

## Checklist

## N.B - this is an opinionated regex!

I appreciate that we want a best-case regex suitable for most users, and no regex is 100% for these kinds of things, but my edge-case spider-sense is tingling 🕷️

### Email addresses
Technically, an email address such as `name@tld` is perfectly valid, and there are TLDs with MX records in the wild ([like guatemala's .gt](https://mxtoolbox.com/SuperTool.aspx?action=mx%3agt&run=toolpage)).

I'd imagine, as when this was first written, that most users would want to make sure that there is a dot in the second part of the email... but maybe we want to take this opportunity to remove that?

### URLs
The previous URL regex does allow for `localhost`, when this is fixed it wont. I'm not sure if that's a problem? I'd imagine that most people validating a URL in the backoffice would want to make sure it's "real", which localhost technically isn't , but I'm not 100% on that.

### DIFF
--- a/src/backoffice/shared/modals/property-settings/property-settings-modal.element.ts
+++ b/src/backoffice/shared/modals/property-settings/property-settings-modal.element.ts
@@ -123,7 +123,7 @@ export class UmbPropertySettingsModalElement extends UmbModalBaseElement<undefin
 		{
 			name: 'Validate as an email address',
 			value: 'email',
-			validation: '[a-zA-Z0-9_.+-]+@[a-zA-Z0-9-]+.[a-zA-Z0-9-.]+',
+			validation: '[a-zA-Z0-9_.+-]+@[a-zA-Z0-9-]+\\.[a-zA-Z0-9-.]+',
 		},
 		{
 			name: 'Validate as a number',
@@ -133,7 +133,7 @@ export class UmbPropertySettingsModalElement extends UmbModalBaseElement<undefin
 		{
 			name: 'Validate as an URL',
 			value: 'url',
-			validation: 'https?://[a-zA-Z0-9-.]+.[a-zA-Z]{2,}',
+			validation: 'https?://[a-zA-Z0-9-.]+\\.[a-zA-Z]{2,}',
 		},
 		{
 			name: '...or enter a custom validation',


nul800sebastiaan:
In my humble opinion: this validation has always had edge-csase complaints over the years, I personally would make it not require a dot after the `@` because that is "more correct". It still doesn't apply to all edge cases though as I recall. 
Maybe we should either have multiple email regex examples in the dropdown, or none and link to docs to pick the most applicable one?

JasonElkin:
Edge case complaints @nul800sebastiaan? [I can't think what you mean](https://github.com/umbraco/Umbraco-CMS/issues/10404) 😉

I'd personally prefer `[a-zA-Z0-9_.+-]+@[a-zA-Z0-9.-]+`, happy to update.

Not sure about multiple options, how does one signpost the difference in the backoffice?

mikecp:
What about letters with accents, which are now accepted in emails IIRC 😁😅

iOvergaard:
> Edge case complaints @nul800sebastiaan? [I can't think what you mean](https://github.com/umbraco/Umbraco-CMS/issues/10404) 😉
> 
> I'd personally prefer `[a-zA-Z0-9_.+-]+@[a-zA-Z0-9.-]+`, happy to update.
> 
> Not sure about multiple options, how does one signpost the difference in the backoffice?

I agree with your suggestion, @JasonElkin. Let's simplify the regex to match the RFC for emails more closely. This version of the backoffice will land in a major version and thus be appropriate to introduce a potential breaking change 😇

Would you update your PR?

iOvergaard (changes requested):
Changes requested

iOvergaard (approved):
Will merge this. I think it's fine for now having a more complex email regex, since it's meant to help the user type a "valid" public email.

PR #614, opened 2023-03-21
correct memoization methods and test

nielslyngsoe:
The memoization methods was done in the wrong way, corrected and made a test to prove its working.


madsrasmussen (approved):
Fantastic. Great with a test for this 🥇

Issue #160, opened 2022-09-23
Dashboard does not scroll when content height exceeds the dashboard height


bkclerke:
**Issue:** When dashboards contain content height that exceeds the dashboard height, the dashboard does not scroll.

**Expected:** The dashboard content would scroll and not be cut off.

![image](https://user-images.githubusercontent.com/4968254/191960953-b4cd7d8b-40bd-4fd4-8a29-a8a6a669dfbc.png)

**Example Branch:** feature/published-dashboard

**Repro Steps:** 
1. Add UUI boxes to dashboard that exceed the height of the dashboard.

markadrake:
# Same or Related Issue

I also see this issue in the `main` branch today.

The `<uui-scroll-container>`'s height is restricted for 2 reasons. First, `<umb-section-dashboards>` has no explicit height. The second cause for this behavior seems to be that the CSS selector: `#scroll-container` applies `height:500px` to the element. 

<img width="1728" alt="image" src="https://user-images.githubusercontent.com/713355/192155526-0e38eb56-37a4-42c6-bf56-ce50a6f8182b.png">

The overflow is scrollable:

https://user-images.githubusercontent.com/713355/192160037-c337068b-d6b3-4dd0-bed3-c236bc484cbe.mov

But obviously, there's unclaimed real estate on the screen.

# How to Remedy this in the Browser

- `<umb-section-dashboards>` was changed to `display:flex` and given `flex-direction:column;` and `height:100%`.
-  `<uui-scroll-container>` was given `flex-grow:1`.
   - The `height:500px` can be removed from `#scroll-container`. 

What I don't know:

- The architecture and intent of `<umb-section-dashboards>`. Will this component always contain `<uui-tab-group>` and `<uui-scroll-container>`? 
   - Are there other use cases? 
   - Do my CSS recommendations produce any side effects?

# Hopefully Constructive Feedback...

- I don't find it to be _great practice_ to use the `id` attribute for styling. I typically instruct others to forget the `id` attribute exists.
   - I think the CSS selector should match the component's name, in this case, `.uui-scroll-container`.
   - I believe CSS can be applied to the root with the `:host` selector. So applying any `id` or `class`(es) may be unnecessary.
- Is the use of two prefixes, `umb` and `uui`, intentional? I.E., does `umb` denote a UI element intended only to be used by Umbraco?

mattbrailsford:
@markadrake just in response to your feedback

[...]

PR #616, opened 2023-03-22
Fixes broken workspace scroll

nathanwoulfe:
Revisiting changes made in #160, after introduction of `umb-router-slot` caused some layout issues

Changes here fix issues with vertical scroll in dashboards and workspaces, where previously these were not scrollable. Tried for a consistent approach where the required styling for `umb-router-slot` and its child `router-slot` is set once and works everywhere, but have still had to manage some views individually.

Also tidied up whitespace around dashboard elements, making this consistent, and cleaned up some redundant CSS.

May not be perfect just yet, but fixes the main issues with scroll


madsrasmussen (approved):
It looks good to me 👍 

I found the issue with double save buttons on Member Groups. The save data type workspace action was set to show up on the Member Group workspace. I have fixed it as part of this PR.

PR #619, opened 2023-03-23
Use wca to generate the JSON format for VSCode for HTMLCustomData

warrenbuckley:
Use `web-component-analyzer` to generate the JSON format for VSCode for HTMLCustomData to help with completions.

## Web Component Analyzer Tool
`wca analyze src --format vscode --outFile vscode-html-custom-data.json`

## VSCode Custom HTML Data

https://github.com/microsoft/vscode-custom-data
https://github.com/microsoft/vscode-custom-data/tree/main/samples/webcomponents
https://code.visualstudio.com/api/extension-guides/custom-data-extension

## Usage
The JSON file would need to be shipped as part of a NPM package then people would need to update their VSCode workspace settings file at `.vscode/settings.json`

## Problems
This will only work in .html files and NOT inside other Lit based WebComponents in the html tagged literal.
I have raised an issue on the popular VSCode plugin `lit-plugin` to see if this would work.

https://github.com/runem/lit-analyzer/issues/300

### DIFF
--- a/.github/workflows/npm-publish-github-packages.yml
+++ b/.github/workflows/npm-publish-github-packages.yml
@@ -38,6 +38,8 @@ jobs:
           node-version: 18
           cache: 'npm'
       - run: npm ci
+      - run: npm run wc-analyze
+      - run: npm run wc-analyze:vscode
       - run: npm run build:libs
       - uses: actions/upload-artifact@v3
         with:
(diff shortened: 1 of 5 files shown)


iOvergaard:
Awesome, yeah. We need to run both wca commands as part of the [publish job](https://github.com/umbraco/Umbraco.CMS.Backoffice/blob/main/.github/workflows/npm-publish-github-packages.yml) as well as copy over both files in the [move-libs.js](https://github.com/umbraco/Umbraco.CMS.Backoffice/blob/main/utils/move-libs.js) helper. Would you put that into this PR?

iOvergaard:
And we need to add more paths to the paths filter in the publish workflow, e.g. `src/**/*.element.ts`

warrenbuckley:
@iOvergaard I think I understood your comments correctly, but please check it over 😄

iOvergaard (approved):

PR #620, opened 2023-03-23
Feature: Tree Item extension point

madsrasmussen:
This PR adds a new extension point to overwrite the rendering and logic of a tree item.

To see an example of a custom tree item. I have made a temporary Document Tree Item where we can render a status symbol for different states. It will also be possible to show the correct node name depending on the app global language.

The following docs are also added to storybook:

# Trees
The tree is a hierarchical structure of nodes. The tree is registered in the Backoffice extension registry. A tree can be rendered anywhere in the Backoffice with the help of the umb-tree element.

## Registering a tree

Tree Manifest

```json
{
  "type": "tree",
  "alias": "My.Tree.Alias",
  "name": "My Tree",
  "meta": {
    "repositoryAlias": "My.Repository.Alias"
  }
},

{
  "type": "treeItem",
  "kind": "entity",
  "alias": "My.TreeItem.Alias",
  "name": "My Tree Item",
  "conditions": {
    "entityType": "my-entity-type",
  },
}
```

The backoffice comes with two different tree item kinds out of the box:
entity and fileSystem

## Rendering a tree

```html
<umb-tree alias="My.Tree.Alias"></umb-tree>
```

## Render a Custom Tree Item

The Tree Item Manifest

```json
{
  "type": "treeItem",
  "alias": "Umb.TreeItem.Alias",
  "name": "My Tree Item",
  "js": "./my-tree-item.element.js",
  "conditions": {
    "entityType": "my-entity-type",
  },
}
```

### The Tree Item Element

```typescript
import { css, html, nothing } from 'lit';
import { customElement, property } from 'lit/decorators.js';
import { UmbElementMixin } from '@umbraco-cms/backoffice/element';
import { UmbMyTreeItemContext, MyTreeItemDataModel } from './my-tree-item.context';

@customElement('my-tree-item')
export class MyTreeItemElement extends UmbElementMixin(LitElement) {
	private _item?: MyTreeItemDataModel;
	@property({ type: Object, attribute: false })
	public get item() {
		return this._item;
	}
	public set item(value: MyTreeItemDataModel | undefined) {
		this._item = value;
		this.#context.setTreeItem(value);
	}

	#context = new UmbMyTreeItemContext(this);

	render() {
		if (!this.item) return nothing;
		return html` <umb-tree-item-base> Some custom markup </umb-tree-item-base>`;
	}
}

export default MyTreeItemElement;
```

### The Tree Item Context

```typescript
// TODO: auto generate this from the interface
export interface UmbTreeItemContext<T> {
	host: UmbControllerHostInterface;
	unique?: string;
	type?: string;

	treeItem: Observable<T | undefined>;
	hasChildren: Observable<boolean>;
	isLoading: Observable<boolean>;
	isSelectable: Observable<boolean>;
	isSelected: Observable<boolean>;
	isActive: Observable<boolean>;
	hasActions: Observable<boolean>;
	path: Observable<string>;

	setTreeItem(treeItem: T | undefined): void;
	requestChildren(): Promise<{
		data: PagedResponse<T> | undefined;
		error: ProblemDetailsModel | undefined;
		asObservable?: () => Observable<T[]>;
	}>;
	toggleContextMenu(): void;
	select(): void;
	deselect(): void;
	constructPath(pathname: string, entityType: string, unique: string): string;
}
```

### Extending the Tree Item Context base

We provide a base class for the tree item context. This class provides some default implementations for the context. You can extend this class to overwrite any of the default implementations.

```typescript
export class UmbMyTreeItemContext extends UmbTreeItemContextBase<MyTreeItemDataModel> {
	constructor(host: UmbControllerHostInterface) {
		super(host, (x: MyTreeItemDataModel) => x.unique);
	}

	// overwrite any methods or properties here if needed
}
```


libs/store/file-system-tree.store.ts
```diff
@@ -0,0 +1,67 @@
...
+	}
+
+	/**
+	 * Updates an item in the store
+	 * @param {string} path
+	 * @param {Partial<FileSystemTreeItemPresentationModel>} data
+	 * @memberof UmbFileSystemTreeStore
+	 */
+	updateItem(path: string, data: Partial<FileSystemTreeItemPresentationModel>) {
+		this.#data.next(partialUpdateFrozenArray(this.#data.getValue(), data, (entry) => entry.path === path));
```
nielslyngsoe, line 30: Here you can use this method instead, that would do the same:

```suggestion
		this.#data.appendOne(data)
```

src/backoffice/documents/documents/tree/tree-item/document-tree-item.element.ts
```diff
@@ -0,0 +1,72 @@
...
+			}
+		`,
+	];
+
+	@property({ type: Object, attribute: false })
+	public item?: DocumentTreeItemResponseModel;
+
+	render() {
+		if (!this.item) return nothing;
+		new UmbDocumentTreeItemContext(this, this.item);
```
nielslyngsoe, line 39: As we have talked about on Slack this might be created too often?
I'm also affright of the readability, when a developer reads this it might not so clear when its replaced — which is also the potential problem, that its re-generateed too often.

src/backoffice/shared/components/tree/tree-item-base/tree-item-base.element.ts
```diff
@@ -0,0 +1,174 @@
...
+				: nothing}
+		`;
+	}
+
+	#renderChildItems() {
+		return html`
+			${this._childItems
+				? repeat(
+						this._childItems,
+						// TODO: get unique here instead of name. we might be able to get it from the context
```
nielslyngsoe, line 161: Ouch, Just for rendering/identification of data changes. I think we should request backend to reveal the key/unique for the data? Otherwise re-rendering/RxJS updates cannot be optimized?

src/backoffice/shared/components/tree/tree-item/tree-item.element.ts
```diff
@@ -0,0 +1,29 @@
...
+export class UmbTreeItemElement extends UmbLitElement {
+	static styles = [UUITextStyles, css``];
+
+	@property({ type: Object, attribute: false })
+	item?: TreeItemPresentationModel;
+
+	render() {
+		return html`<umb-extension-slot
+			type="treeItem"
+			.filter=${(manifests: ManifestTreeItem) => manifests.conditions.entityType === this.item?.type}
```
nielslyngsoe, line 18: Could it make sense not to render the extension slot when there is no item? like rendering nothing?

src/backoffice/shared/components/tree/tree.context.ts
```diff
@@ -23,15 +25,44 @@ export class UmbTreeContextBase implements UmbTreeContext {
...
+		new UmbObserverController(
+			this.host,
+			umbExtensionsRegistry.getByTypeAndAlias('repository', this.tree.meta.repositoryAlias),
+			async (repositoryManifest) => {
+				if (!repositoryManifest) return;
+
+				try {
+					const result = await createExtensionClass<UmbTreeRepository>(repositoryManifest, [this.host]);
+					this.repository = result;
+					this.#checkIfInitialized();
```
nielslyngsoe, line 53: I think we could make a AsPromise on the Observer Controller, which could be 'extended' by the given method returning a PromiseAlikeResult which would postpone its completion.

Then this promise could be used instead of the `#checkIfInitialized`.

Don't know if that concept can do us any good? but that was my thoughts on this one.

nielslyngsoe: I'll just create a task for it in Azure Boards. Then we can revisit in near future.

src/backoffice/shared/components/tree/tree-item/tree-item.element.ts
madsrasmussen: Thanks! Fixed 👍

nielslyngsoe (commented):
looks good :-)

nielslyngsoe (approved):

src/backoffice/shared/components/tree/tree-item-base/tree-item-base.element.ts
nielslyngsoe: Seems alright for now.

PR #622, opened 2023-03-27
Umb debug - Small refactor to prep for Browser Extension

warrenbuckley:
This PR does the following:
Changes the property 'enabled` to `visible` as the presence of umb-debug component on the page will still emit the events to collect the contexts.

The browser extension will insert umb-debug into the DOM where the user clicks/points and inspects an element in browser developer tools, but this will not be shown/visible.

Next is a small refactor, the callback function after collecting the contexts, will massage the data into a format that it can emit as an event that the browser extension can listen for and display the data in.

Rather than doing the work twice, the HTML rendering of umb-debug itself is refactored to avoid repeating same or similar code.


warrenbuckley:
OK looks like I need to fix up some errors first 🙈

warrenbuckley:
@iOvergaard this is all green now 😄

iOvergaard (changes requested):
As discussed on a call, you'll refactor a bit and make sure your DevTools extension works with custom events which you'll also start sending out from app.ts.

warrenbuckley:
Please check this over @iOvergaard about what we discussed and ensure the umb-debug element still works in the Published Status dashboard in the settings section.

`http://localhost:5175/section/settings/dashboard/published-status`

iOvergaard (approved):
LGTM

PR #624, opened 2023-03-28
Feature: One step closer to core code aligned with packages

madsrasmussen:
This PR is an attempt to format our own code base close to how packages could be set up.

Each "area/silo" in the Backoffice folder is considered a package. Documents, Media, Templating etc. As part of each area I have added a `umbraco-package.ts` (it is currently not possible to make this a json file). The `umbraco-package.ts` exports members corresponding to the `umbraco-package.json`. For Documents it could look like this:

```ts
export const name = 'Umbraco.Core.DocumentManagement';
export const version = '0.0.1';
export const extensions = [
  {
    type: 'entryPoint',
    loader: () => import('./index'),
  },
];
```

Each of our packages registers an entry point, that now looks like a "real" extension entry point type.

```ts
import { manifests as dashboardManifests } from './dashboards/manifests';
import { manifests as documentManifests } from './documents/manifests';
import type { UmbEntrypointOnInit } from '@umbraco-cms/backoffice/extensions-api';
import type { ManifestTypes } from '@umbraco-cms/backoffice/extensions-registry';

export const manifests: Array<ManifestTypes> = [
  ...dashboardManifests,
  ...documentManifests,
];

export const onInit: UmbEntrypointOnInit = (_host, extensionRegistry) => {
  extensionRegistry.registerMany(manifests);
};
```

The PR also remove some of the hard coded providers from the backoffice element. I have tried my best to position them in a package I think would make sense.

The PR adds a temp method to the backoffice element that handles our core packages. Some time in the future we should get rid of that, and handle them the same way as any other package, so they potentially could show up in the package section list of packages.

**Known issues**
The PR shows an issue with our router. When all packages are lazy loaded we run into troubles with catch-all routes. The first router in the backoffice doesn't wait extensions to be loaded and will redirect too soon. The result is that all deep linking often fails.


nathanwoulfe:
One benefit I see here is for package developers - if the backoffice itself is comprised of packages, it becomes a concrete example of how to approach building third-party packages. Modularity is awesome too!

iOvergaard:
I like the idea. I have a basic concern in that we _know_ that we have to load our packages, so doesn't it create an unnecessary overhead by doing that asynchronously in the BackofficeElement? It seems we do a lot of work to prove that our own modules can load through an entrypoint extension, but wouldn't change the effect much if each index.ts file just loaded its sections directly into the extension registry.

An advantage, though, is that each entrypoint would be able to check if it's allowed to load, e.g. if some AppSetting on the server said that a particular section was not allowed to load, it could verify that itself in its entrypoint onInit() function before registration.

Anyway, yeah, just some random thoughts. Very keen to hear different takes on this :-)

P.S. The async router problem is also a thing if you want to register e.g. a dashboard that has a route - sometimes you'll see it not routing directly to it if you refresh (race conditioned), so probably the same problem only much bigger here. Hopefully this discovery leads to that problem fixed!

libs/extensions-registry/models.ts
```diff
@@ -171,6 +171,6 @@ export interface ManifestWithMeta extends ManifestBase {
 }
 
 export interface ManifestEntrypoint extends ManifestBase {
-	type: 'entrypoint';
+	type: 'entryPoint';
```
iOvergaard, line 174: Please, all our test packages will stop working, do we have to rename this? 😂

madsrasmussen: Test packages? Where do we have those?

iOvergaard: Sorry, I meant packages on various installations. For instance, I have a v13 branch on my machine with some packages and I distributed those as demo files for the Spark presentation.

Do you feel that this change clarifies what this kind of extension is?

madsrasmussen: Ah, I didn't know that. I just noticed that I misspelled this a couple of times so I just wanted to get this updated before we had too many extensions developed.

nielslyngsoe: In general, it seems that "entry point" as two words is most correct. Though it seems one word is acceptable. But I don't think we should respect previous matters at this stage. So I'm in for the most correct at the time being. With the perspective that we are still free to adjust anything, then we should keep that freedom and don't spend time evaluating changes jet. IMO.

madsrasmussen:
> There is an issue with the `src/public-assets/` folder, files from there are not loaded correctly.

This is also a problem on the main branch, so it is not caused by this PR. Let's fix that in a new PR.

madsrasmussen:
The Mocked Extension issue is fixed in: #657

nielslyngsoe (approved):

PR #625, opened 2023-03-28
Chore: Eslint rule to enforce element suffix

madsrasmussen:
This PR ensures that all elements always suffix their class name with "Element"


iOvergaard (approved):
LGTM but there is a merge conflict

PR #628, opened 2023-03-29
Feature/create language

loivsen:
.


src/backoffice/settings/languages/workspace/language/language-workspace-edit.element.ts
```diff
@@ -58,19 +78,28 @@ export class UmbLanguageWorkspaceThingyElement extends UmbLitElement {
 	render() {
 		return html`<umb-workspace-layout alias="Umb.Workspace.Language">
 			<div id="header" slot="header">
-				<uui-button href="/section/settings/workspace/language-root" compact>
+				<uui-button label="Navigate back" href="/section/settings/workspace/language-root" compact>
```
iOvergaard, line 81: I know this isn't new, but links should be relative because we are using a `<base>` tag

```suggestion
				<uui-button label="Navigate back" href="section/settings/workspace/language-root" compact>
```

src/backoffice/settings/languages/workspace/language/language-workspace.element.ts
```diff
@@ -6,24 +6,33 @@ import { UmbLanguageWorkspaceContext } from './language-workspace.context';
...
+	#element = document.createElement('umb-language-workspace-edit');
 
 	#routerPath? = '';
 
-	// TODO: add create route
 	@state()
 	_routes = [
 		{
 			path: 'edit/:isoCode',
-			component: () => import('./language-workspace-edit.element'),
```
iOvergaard, line 22: Didn't this work before or is there a reason to change this? Now we are loading the edit component every time you see the overview of languages.

src/backoffice/settings/languages/workspace/language/language-workspace-edit.element.ts
```diff
@@ -58,19 +78,28 @@ export class UmbLanguageWorkspaceThingyElement extends UmbLitElement {
...
-				<uui-input value=${ifDefined(this._language?.name)} @input="${this.#handleInput}"></uui-input>
+				${this._isNew
+					? html`<strong>Add language</strong>`
+					: html`<uui-input
+							label="Language name"
+							value=${ifDefined(this._language?.name)}
+							@input="${this.#handleInput}"></uui-input>`}
+			</div>
+			<div slot="footer" id="footer">
+				<a href="/section/settings/workspace/language-root">Languages</a> /
```
iOvergaard, line 92:
```suggestion
				<a href="section/settings/workspace/language-root">Languages</a> /
```

iOvergaard (changes requested):
Looks good, just left a few suggestions :)

src/backoffice/settings/languages/workspace/language/language-workspace.element.ts
loivsen: The idea was to not reload this element when going from create to edit, since they use the same layout the element is reused for both paths

iOvergaard: Does it reload, though, or does it reuse the already existing element?

loivsen: I believe it hits the already existing element provided, but @nielslyngsoe can probably shed some light on this :)

nielslyngsoe: This is the way we are doing it now for workspace routing in general — Seems alright for now, but we might look into when the element is loaded. :-)

iOvergaard (approved):
I tested this up against the v13/dev branch and found a few oddities and fixed those.

One thing I noticed is that some languages have an invalid iso code, so can't be created on the server, but we receive the list of cultures from the server, so would expect those to just work. Something for the backend to look at. "English (Albanian)" is one such example.

PR #629, opened 2023-03-29
Preperations for work on Document Type Workspace

nielslyngsoe:
Some initial work for the Document Type Workspace, which I like to get merged. So we can avoid merg conflict in near future.


libs/context-api/token/entity-workspace-context.token.ts
```diff
@@ -0,0 +1,7 @@
+import { UmbEntityWorkspaceContextInterface } from '../../../src/backoffice/shared/components/workspace/workspace-context/workspace-entity-context.interface';
```
madsrasmussen, line 1: We should avoid loading from the src folder in a lib. We should move this file to the workspace lib instead.

nielslyngsoe: Ah, good one! Thanks. will correct right away

PR #630, opened 2023-03-30
Feature: Create data type folders (1. round)

madsrasmussen:
This PR adds the first iteration of creating, editing and deleting folders.

The PR implements folder management for data types, but I have created generic modal to create and update a folder. The flow is currently added for entity actions. That makes it possible to use the actions menu in the tree.

**Known issue:**
* Currently all folder actions show up on all tree items. We will have to implement extension conditions so we only show delete and rename folder on folders, etc.


iOvergaard (approved):
Tested on v13/dev and seems to work there.

PR #638, opened 2023-04-03
Log viewer links

iOvergaard:
- [x] Let date selector update 'dateFrom' and 'dateTo' params in the url
- [x] Let log level selector update through the 'loglevels' param in the url
- [x] Figure out why URL params disappear after reload

Warning: Cannot be merged before #642


julczka (approved):
All good! Thank you!

PR #641, opened 2023-04-04
Proff of concept for navigating to edit path when created

nielslyngsoe:
Lazy load workspace editor.
Navigate to the edit URL when created.

**Test steps:**
Create a new language
See the URL changed to edit/:isoCode when creating was succesful.


libs/router/generate-route-path-builder.function.ts
```diff
@@ -0,0 +1,21 @@
+import type { ISlashOptions } from '@umbraco-cms/internal/router';
```
iOvergaard, line 1: This type will be moved to "@umbraco-cms/backoffice/router" (so can be imported locally) after #642

libs/router/route.context.ts
```diff
@@ -1,5 +1,6 @@
 import type { IRoutingInfo, ISlashOptions } from 'router-slot';
 import { UmbRoute } from './route.interface';
+import { generateRoutePathBuilder } from '.';
```
iOvergaard, line 3: Consider importing directly from the "./generate-route-path-builder.function" file here to reduce cyclic import

iOvergaard (dismissed):

PR #642, opened 2023-04-04
Feature/router slot 2.0.0

iOvergaard:
This upgrades our internal build of router-slot from 1.6.1 to 2.0.0 with a new feature set, which mainly includes the support for query parameters.

[Check the changeset out on the router-slot fork](https://github.com/umbraco/router-slot/compare/1.6.1...2.0.0).

[The main change is in anchor.ts](https://github.com/umbraco/router-slot/compare/1.6.1...2.0.0#diff-e20332f42de062090a3f49e1bf8bda9cd91099ce5ca5261f6f0d409a9ffe3786)

## Changes on the backoffice
* You may now only import router-slot models and utils from `@umbraco-cms/backoffice/router`
  * There is a new local eslint rule to check for this
* We have some events and elements for the backoffice only on `@umbraco-cms/internal/router`
  * The events can potentially be moved to the router library if we create some interfaces for `UmbRouterSlot`


nielslyngsoe:
Hi @iOvergaard 

Just reviewing this, the code looks good, and what I see works. But the section sidebar is gone. So there is a bit to investigate.
image

PR #643, opened 2023-04-05
Feature/property editor tags

iOvergaard:
TODO


iOvergaard (commented):
LGTM but cannot really test it against the server before we can create real documents.

loivsen (approved):

PR #649, opened 2023-04-16
Chore: Ensure Umb class prefix

madsrasmussen:
The PR includes an ESLint rule that enforces the use of an "Umb" prefix for all classes. There were a few instances where this convention was not followed, but this update ensures consistent naming for all classes.


nielslyngsoe (approved):
good

PR #650, opened 2023-04-17
Keep other setInitStatus in app.ts

nathanwoulfe:
#647 removed the duplicate `setInitStatus` call in the app element, but we need to keep the one in `connectedCallback` rather than the constructor as we need the OpenAPI configuration to be completed before attempting to fetch the initial status, else we don't have the base URL set.


iOvergaard (approved):
It's fine. Good catch. I'm moving stuff in the authentication branch where everything will happen in the constructor, but will approve this for now because it will be a while before the auth stuff can be merged.

PR #654, opened 2023-04-20
Authentication using OpenID

iOvergaard:
## Description

Adds OpenID authentication using the library `@openid/appauth` which facilities features such as reading the code from the URL and re-hydrating authentication state in local storage, which is necessary when you send the user back and forth to the management API.

Most of the code is run through UmbAppElement, so check that out. 

This PR also adds fixes to the UmbResourceController to control how and which errors are shown. We are now no longer notifying the user if a CancelError occurs, which is a good kind of error.

It also makes sure that we do not use the ProblemDetailsModel when encountering programmatic errors but instead, the errors are thrown. We decided that we want to throw errors and be quite harsh about it because this is something you need to fix. If we for some reason would like to have errors be notified to the user, it is better that we come up with a model of our own.

Lastly, we now also show an error page (new: UmbErrorElement) when an unrecoverable error is encountered (mostly occurs if we cannot reach the backend server) and I built that in to the router, which seems to be running in a better state.

## How to test

This is only testable up against the v13 branch on Umbraco CMS. You need first to decide if you want to test with Vite or directly from the server. This is important since you need to tell the server what hostname your backoffice is being served from (this will be automatically deducted in the future).

### Running directly off the server
1. Checkout v13/dev
2. Run `git submodule update --init`
3. Go to src/Umbraco.Web.UI.Client.New
4. Check out this branch: `git checkout poc/openid-auth`
5. Rebuild the backoffice: `npm ci && npm run build:for:cms`
6. Go back to src/Umbraco.Web.UI.New
7. Start the backend server: `dotnet run`
8. Access https://localhost:44339/umbraco and install Umbraco
9. Check that you are sent to the login page
10. Log in

### Running with Vite
1. Perform steps 1, 2, 6, 7, 8 from before
2. Open this file in an editor: `src/Umbraco.Web.UI.New/appsettings.Development.json`
3. Add this to the Umbraco.CMS section:
```
      "NewBackOffice":{
        "BackOfficeHost": "http://localhost:5173"
      },
```
4. Set Vite to use Umbraco API by copying the ".env" file to ".env.local" and setting the following:
```
VITE_UMBRACO_USE_MSW=off
VITE_UMBRACO_API_URL=https://localhost:44339
```
5. Start the vite server: `npm run dev` in your backoffice folder
6. Check that you are sent to the login page
7. Log in

## What does not work yet

* You cannot log out through the UI
* If your session expires or your token is revoked, you will start getting 401 network errors, which for now only will be shown as a notification - we need to figure out how to send you back to log in
* We do not _yet_ poll to see if the token is still valid or check how long before you are logged out
* The whole thing is running through a class called UmbAuthFlow, which Google has made and I have heavily modified - we need to figure out if we want this to act through a controller instead of being hooked directly in to the UmbAppElement


src/app.ts
```diff
@@ -44,27 +60,68 @@ export class UmbAppElement extends UmbLitElement {
 		},
 	];
 
+	#authFlow: UmbAuthFlow;
 	#umbIconRegistry = new UmbIconStore();
 	#uuiIconRegistry = new UUIIconRegistryEssential();
 	#runtimeLevel = RuntimeLevelModel.UNKNOWN;
+	#isMocking = import.meta.env.VITE_UMBRACO_USE_MSW === 'on';
```
madsrasmussen, line 67: Just a small note for ourselves. I think we should look into moving all mocking logic outside of the app element, so the element doens't know anything about what is serving the data.

The code is already like this, so it is fine to look into this outside of this PR.

madsrasmussen: I have added a TODO

madsrasmussen (approved):
Awesome!

* real CMS flow works as described 🎉 
* mock flow works as described. 🎉

PR #657, opened 2023-04-20
Bugfix: Load plugins from mocked App_Plugins folder

madsrasmussen:
Vite couldn't load the mocked extension beacuse the files were not copied to the public folder. The PR adds the App_Plugin files to be copied to bypass Vites dynamic import limit.

### DIFF
--- a/vite.config.ts
+++ b/vite.config.ts
@@ -9,6 +9,10 @@ export const plugins: PluginOption[] = [
 				src: 'public-assets/icons/*.js',
 				dest: 'icons',
 			},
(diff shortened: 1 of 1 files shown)


nielslyngsoe (approved):

iOvergaard:
Hmm, I removed these because we didn't need them when running on the v13 branch. We'll probably need to split up the plugins then for the two vite configs.

nielslyngsoe:
Interesting. And yes we need to support both cases in some way :-)

PR #668, opened 2023-04-26
Bugfix/logviewer saved search fixes

julczka:
This PR fixes bugs in log viewer and fully implements saved searches funcionality
## Types of changes

## Checklist

### DIFF
--- a/src/backoffice/settings/logviewer/workspace/logviewer.context.ts
+++ b/src/backoffice/settings/logviewer/workspace/logviewer.context.ts
@@ -258,7 +258,9 @@ export class UmbLogViewerWorkspaceContext {
 			return;
 		}
 
-		this.#isLoadingLogs.next(true);
+		const isPollingEnabled = this.#polling.getValue().enabled;
+
+		if (!isPollingEnabled) this.#isLoadingLogs.next(true);
 
 		const skip = (this.currentPage - 1) * 100;
 		const take = 100;
--- a/src/backoffice/settings/logviewer/workspace/logviewer.context.ts
+++ b/src/backoffice/settings/logviewer/workspace/logviewer.context.ts
@@ -89,8 +89,6 @@ export class UmbLogViewerWorkspaceContext {
 	logs = createObservablePart(this.#logs, (data) => data?.items);
 	logsTotal = createObservablePart(this.#logs, (data) => data?.total);
 
-
-
 	#polling = new UmbObjectState<PoolingCOnfig>({ enabled: false, interval: 2000 });
 	polling = createObservablePart(this.#polling, (data) => data);
 
@@ -106,12 +104,7 @@ export class UmbLogViewerWorkspaceContext {
 		this.#repository = new UmbLogViewerRepository(this.#host);
 	}
 
-	async init() {
-		this.validateLogSize();
-	}
-
 	onChangeState = () => {
-
 		const searchQuery = query();
 		let sanitizedQuery = '';
 		if (searchQuery.lq) {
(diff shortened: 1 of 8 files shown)


src/backoffice/settings/logviewer/workspace/logviewer.context.ts
```diff
@@ -101,18 +104,7 @@ export class UmbLogViewerWorkspaceContext {
...
-		this.validateLogSize();
-	}
-
-	reset() {
-		this.#logs.next({ items: [], total: 0 });
-		this.setCurrentPage(1);
-	}
-
 	onChangeState = () => {
-		this.reset();
```
iOvergaard, line 114: I thought it made sense to at least clear the visible logs here when searching for new logs to be sure that if a server errors occurs, then we don't show out-dated logs. What do you think about that?

src/backoffice/settings/logviewer/workspace/views/overview/log-overview-view.element.ts
iOvergaard, line 1: When a saved search is pretty short, it's hard to find the "click target" since it follows the length of the message. Consider if we want the entire line to be clickable:

image

src/backoffice/settings/logviewer/workspace/views/search/components/log-viewer-search-input.element.ts
iOvergaard, line 1: Don't know if caused by this PR, but when I go to the search view, the search bar is so long that I cannot see the date pickers on my smaller screen:

image

iOvergaard: image

iOvergaard (changes requested):
It all seems to work, only found a few quirks.

src/backoffice/settings/logviewer/workspace/logviewer.context.ts
julczka: There is a loading state implemented to "cover" that moment :) With the reset it was bit to flickering. It look slike this now, so there is no risk in seeing "old" logs from before the request 
log-viewer-loading staTE

src/backoffice/settings/logviewer/workspace/views/overview/log-overview-view.element.ts
julczka: done

src/backoffice/settings/logviewer/workspace/views/search/components/log-viewer-search-input.element.ts
julczka: fixed, that was due to some super long wors in the log messages. added the `word-break: break-all;` on this div and it sceles properly now

src/backoffice/settings/logviewer/workspace/logviewer.context.ts
```diff
@@ -262,10 +254,14 @@ export class UmbLogViewerWorkspaceContext {
 	}
 
 	getLogs = async () => {
-		if (!this.#canShowLogs.getValue()) {
+		if (this.#canShowLogs.getValue() === false) {
```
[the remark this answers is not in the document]
julczka: So this works as folowing: 
whenever date range is set we validate the log size. 
if the observable `this.#canShowLogs` is equal to false a [warning is displayed](https://github.com/umbraco/Umbraco.CMS.Backoffice/blob/602a7ea59a7187fa8a5a0735a6dc271af3142e25/src/backoffice/settings/logviewer/workspace/views/overview/log-overview-view.element.ts#L70-L71) instead of saved searches/log messages, so they will not see any logs if that changes value. 
This is added so it just does not make the log request when it cannot show them, I dont know what comes back from the endpoint if therer are to many messages in a given date range, but I guess there is no point in making the request for logs at all if there are to many to show. I'm not sure if I understand your commnet correctly actually.

PR #669, opened 2023-04-26
Feature: Allow picking of Tree Root

madsrasmussen:


nathanwoulfe (approved):
LGTM - tree roots are pickable, routing works as expected 👍

PR #691, opened 2023-05-03
Feature/more property editors

loivsen:
- Label 
- Decimal & Numeric
- Textbox


src/backoffice/core/property-editors/uis/number/property-editor-ui-number.element.ts
```diff
@@ -1,26 +1,52 @@
...
+		const step = config.find((x) => x.alias === 'step');
+		if (step) this._step = step.value;
+	}
 
 	private onInput(e: InputEvent) {
 		this.value = (e.target as HTMLInputElement).value;
 		this.dispatchEvent(new CustomEvent('property-value-change'));
 	}
 
+	//TODO: UUI-INPUT needs to handle max, min, and step natively in its input field, maybe make a uui-input-number?
```
iOvergaard, line 40: Is this TODO still relevant?

src/backoffice/core/property-editors/uis/label/property-editor-ui-label.element.ts
```diff
@@ -1,8 +1,11 @@
-import { html } from 'lit';
-import { customElement, property } from 'lit/decorators.js';
+import { html, nothing } from 'lit';
+import { customElement, property, state } from 'lit/decorators.js';
 import { UUITextStyles } from '@umbraco-ui/uui-css/lib';
 import { UmbPropertyEditorExtensionElement } from '@umbraco-cms/backoffice/extensions-registry';
 import { UmbLitElement } from '@umbraco-cms/internal/lit-element';
+import type { DataTypePropertyPresentationModel } from '@umbraco-cms/backoffice/backend-api';
+
+type umbracoDataValueType = 'STRING' | 'DECIMAL' | 'DATE/TIME' | 'TIME' | 'INTEGER' | 'BIG INTEGER' | 'LONG STRING';
```
iOvergaard, line 8: Is this saved on the config model anywhere? Can we ensure some more type safety somehow, e.g. can they use the same type variable?

loivsen: Its not from what I can see.
Im not sure if these are needed in the property editor UI either, as it doesn't seem like label even has any inputs in the first place?
The old backoffice has all these different label property editors with the different types but Im not sure if they really do anyting in the frontend at all.

iOvergaard: Looks like it is used to tell the server database what the column type should be. These types should be supported, but not sure if they are stringified like you do here: 
image

iOvergaard: We don't need to know any of this in the UI. It is used only in the configuration of the underlying property editor but has no effect on the UI.

src/backoffice/core/property-editors/uis/value-type/property-editor-ui-value-type.element.ts
```diff
@@ -1,24 +1,53 @@
 import { html } from 'lit';
 import { UUITextStyles } from '@umbraco-ui/uui-css/lib';
-import { customElement, property } from 'lit/decorators.js';
+import { UUISelectEvent } from '@umbraco-ui/uui';
```
iOvergaard, line 3:
```suggestion
import type { UUISelectEvent } from '@umbraco-ui/uui';
```

src/backoffice/core/property-editors/uis/label/property-editor-ui-label.element.ts
```diff
@@ -9,16 +9,17 @@ import { UmbLitElement } from '@umbraco-cms/internal/lit-element';
...
+	@property()
+	description = '';
+
 	@property({ type: Array, attribute: false })
-	public config = [];
+	config = [];
 
 	render() {
-		return html`<div>umb-property-editor-ui-label</div>`;
+		return html`${this.value}`;
```
iOvergaard, line 22: This value could contain anything such as custom HTML and scripts. We should probably test it to see if it prints out HTML correctly or if it is escaped. I know that some extensions save HTML directly in a label. We should probably also figure out if we can scrape it a bit to remove dangerous characters by running it through an HTML parser such as [sanitize-html](https://www.npmjs.com/package/sanitize-html)

iOvergaard (approved):

PR #692, opened 2023-05-03
Feature/property editor date

loivsen:
Property Editor Date Picker
- Created a date-input for property editor date picker.
  - date-input can handle the types: date, time, and datetime-local
  - The input has a min, max, and step property. The min and max can be used to limit the date picker to pick a date within a specific date/time period. step is used for interval between date/time that can be picked.
  - date-input value is always in UTC
  - date-input displays the input value in local time or server time, depends on offsetTime setting.
- config repository created, it returns mock data for servertime offset. The datetime input currently expects the offset to be in minutes and calculates the value & display value depends on that, but this can easily be changed. Currently the mock data returns -120, which is in Denmark, summertime
- Created date-input stories


iOvergaard (approved):
LGTM

PR #695, opened 2023-05-03
Chore: Notification / Modal imports

madsrasmussen:
A PR that removes the responsibility from the backoffice element to attach the modal and notification contexts and elements to the Backoffice element. It is done this way to unsure that any package will be able to append new contexts and elements from an entry point.

As part of this I found out that we didn't allowed slotted elements in the Backoffice element. That is now fixed and the element is now fully flexible.


nathanwoulfe (approved):
Makes sense to me - both elements are rendered correctly, modals/notifications work as before 👍

PR #696, opened 2023-05-03
Chore: Reduce number relative imports to core

madsrasmussen:
A PR to reduce the number of relative imports to the core folder. When we do this and build as javascript modules, we end up with a copy of the code in each module. To prevent this we need to build that specific area as a new module, or include it in an already existing module. In the end we will have import alias's for all modules and can build each module separately where they will import each other.


src/backoffice/core/components/backoffice-frame/backoffice-header.element.ts
```diff
@@ -1,28 +1,25 @@
...
 @customElement('umb-backoffice-header')
 export class UmbBackofficeHeaderElement extends LitElement {
-	
-
 	render() {
 		return html`
 			<div id="appHeader">
 				<uui-button id="logo" look="primary" label="Umbraco" compact>
-					<img src="${logoImg}" alt="Umbraco" />
+					<img src="/umbraco_logomark_white.svg" alt="Umbraco" />
```
iOvergaard, line 14: Did you test this against an Umbraco instance? When running against that, the backoffice is hosted in a directory (/umbraco/) and in my testing, it didn't pick up the configured base URL from vite.cms.config.ts unless you actually import the file first and echo it out hence why images were laid out like this.

tsconfig.json
```diff
@@ -46,7 +46,9 @@
 			"@umbraco-cms/backoffice/collection": ["libs/collection"],
 			"@umbraco-cms/backoffice/tree": ["libs/tree"],
 			"@umbraco-cms/backoffice/section": ["libs/section"],
+			"@umbraco-cms/backoffice/variant": ["libs/variant"],
 
+			"@umbraco-cms/backoffice/core/components": ["src/backoffice/core/components"],
```
iOvergaard, line 51: I like this change, just so you know, the rollup setup for libraries do not support multiple entrypoints at the moment, e.g. `/core/components`, because it only builds against the root index.ts file in each library. So we probably need to do something either with the paths or with the rollup setup later on when we get an actual "core" library.

src/backoffice/core/components/backoffice-frame/backoffice-header.element.ts
madsrasmussen: That's true! We need to do some URL prefixing to get that to work in the "real" CMS. I changed this because it seemed to be some Vite configuration magic, and it breaks when I try to do a "simple" build of the packages as modules. I might be able to fix it with some Rollup configs, but it might also be great to be able to easily prefix paths?

iOvergaard: I think this is the only way to get Vite to prefix paths at least 🤔 but the backoffice-* elements don't belong in a library do they?

Just come to think of it, we haven't worked with anything regarding loading assets from a package yet. You need a way to access the server URL, which is currently exposed through a context token, but haven't really tested it.

tsconfig.json
iOvergaard: I don't think it breaks anything. Merely just a heads up if you wanted to move the "core" folder into our libs folder and expect it to work out of the box :)

PR #698, opened 2023-05-05
Implements UmbDataTypePropertyCollection for easy config access

nathanwoulfe:
Adds `UmbDataTypePropertyCollection`, which extends `Array` to add accessor functions for working with datatype properties:
 - `getValueByAlias<T>(alias: string)` returns the value of the datatype property for the given alias 
 - `getByAlias(alias: string)` returns the datatype property for the given alias

`getValueByAlias` does not type guard, which would be nice but perhaps not essential, as the property editor requesting the datatype property shouldn't be dealing with unknown config (eg we know the `maxNumber` config is a number, so will be assigned into a numeric variable, so the type-guard would be redundant.

I've updated instances of `Array<DataTypePropertyPresentationModel>` to `UmbDataTypePropertyCollection` throughout (this is the majority of changes in this PR), to ensure the application still builds, then have implemented in `UmbPropertyEditorUIMultiUrlPickerElement` as an example.

If we're happy with this approach, I can update the remaining property editors. It's really a convenience helper for developers, but could easily be extended with additional functionality if we start seeing more repetitive boilerplate in property editors.


iOvergaard:
Hi @nathanwoulfe, looks good, but there are a lot of file conflicts. I believe some of the property editors were updated and require the new config type. Would you fix it, please?

nathanwoulfe:
@iOvergaard should be better now - I've whipped through and updated all property editor UIs.

src/backoffice/core/components/workspace-property/workspace-property.element.ts
```diff
@@ -90,7 +91,7 @@ export class UmbWorkspacePropertyElement extends UmbLitElement {
 	 * @default ''
 	 */
 	@property({ type: Object, attribute: false })
-	public set config(value: DataTypePropertyPresentationModel[]) {
+	public set config(value: UmbDataTypePropertyCollection) {
 		this._propertyContext.setConfig(value);
```
nielslyngsoe, line 95: I dont think this should be changed. This would still be of the previous type. As you didn't change the value parsed to this element.
That is the raw data given to the context, and first turned into a UmbDataTypePropertyCollection when send to the property.
So this should be reverted.

libs/extensions-registry/interfaces/property-editor-ui-extension-element.interface.ts
```diff
@@ -1,6 +1,6 @@
-import { DataTypePropertyPresentationModel } from '@umbraco-cms/backoffice/backend-api';
+import { UmbDataTypePropertyCollection } from '@umbraco-cms/backoffice/data-type';
 
 export interface UmbPropertyEditorExtensionElement extends HTMLElement {
 	value: unknown;
-	config: DataTypePropertyPresentationModel[];
+	config: UmbDataTypePropertyCollection;
```
nielslyngsoe, line 5: I think we should make the config optional. So you don't need to spin up an empty UmbDataTypePropertyCollection. And additionally, you would know when you got the config (not existing vs. getting the object) :-)

src/backoffice/core/property-editors/uis/toggle/property-editor-ui-toggle.element.ts
```diff
@@ -26,18 +26,11 @@ export class UmbPropertyEditorUIToggleElement extends UmbLitElement implements U
 	_showLabels?: boolean;
 
 	@property({ type: Array, attribute: false })
```
nielslyngsoe, line 28: I would say we can remove the Type here. As its being used for attribute value conversion. As attribute is disabled its not needed anymore :-)

nielslyngsoe: This applied to multiple places, i just didn't want to write a comment on all of them. :-) I hope that's cool

libs/data-type/data-type-property-collection.class.ts
```diff
@@ -0,0 +1,30 @@
+import { DataTypePropertyPresentationModel } from '@umbraco-cms/backoffice/backend-api';
+
+/**
+ * Extends Array to add utility functions for accessing data type properties
+ * by alias, returning either the value or the complete DataTypePropertyPresentationModel object
+ */
+export class UmbDataTypePropertyCollection extends Array<DataTypePropertyPresentationModel> {
```
nielslyngsoe, line 7: I love it, fantastic!

src/backoffice/core/components/property-type-based-property/property-type-based-property.element.ts
```diff
@@ -93,7 +93,7 @@ export class UmbPropertyTypeBasedPropertyElement extends UmbLitElement {
 			this._dataTypeObserver = this.observe(
 				await this._dataTypeRepository.byId(dataTypeId),
 				(dataType) => {
-					this._dataTypeData = dataType?.values || [];
+					this._dataTypeData = new UmbDataTypePropertyCollection(dataType?.values || []);
```
nielslyngsoe, line 96: Maybe there is a detail I don't see here, but wouldn't UmbDataTypePropertyCollection fallback to a empty array if the given value is 'undefined'.

aka. i don't think you need the ` || [] `here

nielslyngsoe (commented):
I @nathanwoulfe great to get your eyes on the Developer Experience. I love this. I just added a few comments here and there. I hope you would have a look at those :-) Thanks

And keep it coming!

src/backoffice/core/property-editors/uis/toggle/property-editor-ui-toggle.element.ts
nathanwoulfe: There's a stack of places in fact - 80+ where we declare type but attribute is false. I can clean those up in a separate PR rather than mixing in to this one.

libs/data-type/data-type-property-collection.class.ts
```diff
@@ -0,0 +1,30 @@
...
+	constructor(args: Array<DataTypePropertyPresentationModel> = []) {
+		super(...args);
+	}
+
+	private _getByAlias(alias: string) {
```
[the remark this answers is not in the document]
nathanwoulfe: Yup, that makes sense, have updated 👍

src/backoffice/core/components/property-type-based-property/property-type-based-property.element.ts
nathanwoulfe: I'll double check, but I do remember something exploding if I didn't include the empty array fallback.

nathanwoulfe: All works fine without the fallback 👍

libs/extensions-registry/interfaces/property-editor-ui-extension-element.interface.ts
nathanwoulfe: Makes sense to me - there are a few places (mainly in WIP property editors) where we have a default empty array. I'll leave those for cleaning up when the editors are built, but have changed config to optional, so the prop doesn't need to exist at all.

nielslyngsoe (approved):
Looks good

src/backoffice/core/property-editors/uis/toggle/property-editor-ui-toggle.element.ts
nielslyngsoe: cool :-)

PR #704, opened 2023-05-09
Feature: Tree Picker Modal

madsrasmussen:
This PR adds a "global" TreePicker Modal that can be used for any default tree picking action.

When opening the modal you can pass a `treeAlias` to the modal and it will render the tree registered for that alias. 

To avoid having to do this every time we open a Tree Picker I have now allowed to pass default data through a `ModalToken` That means we can set the **treeAlias** as part of the token.

With this change you don't have to create any modal elements or register any manifests (other than a tree) unless you want to do something that the default Tree Picker shouldn't support.

As part of this PR I have deleted all the duplicated code we had for all the tree pickers.

**How to set up a token for a Tree Picker.** 
You set the `Umb.Modal.TreePicker` as the `modalAlias`. Pass the `treeAlias` as the modal default data.

```ts
export const UMB_DATA_TYPE_PICKER_MODAL = new UmbModalToken<UmbDataTypePickerModalData, UmbDataTypePickerModalResult>(
  'Umb.Modal.TreePicker',
  {
    type: 'sidebar',
    size: 'small',
  },
  {
    treeAlias: 'Umb.Tree.DataTypes',
  }
);
```


nielslyngsoe (approved):
Looks good

PR #706, opened 2023-05-10
Chore: Remove Vite variables from App element 

madsrasmussen:
Unfortunately Vite doesn't support `importmaps` out of the box (yet), so we will have to test the setup with another dev server. We currently use `@web-dev-server` though the test runner, where `importmaps` are supported.

The app is already build so it is possible to change the server, but we have references to `Vite` inside the app element that needs to live outside of the implementation.

The PR moves the variables outside the app element.


src/app/app.element.ts
```diff
@@ -1,104 +1,102 @@
...
-	private backofficePath = import.meta.env.DEV ? '/' : '/umbraco';
+	// for some reason it is not possible to use private fields for objects.
+	_config: UmbAppConfig = {
+		serverUrl: window.location.origin,
+		backofficePath: '/umbraco',
+		bypassAuth: false,
+	};
+
+	@property({ type: Object, attribute: false })
+	set config(value: UmbAppConfig) {
```
iOvergaard, line 34: Previously I kept these as separate properties with simple values so that they could be set from the backend Razor template in Light DOM individually, e.g.
```
<umb-app backoffice-path="/umbraco"></umb-app>
```

But I don't think I can do that after this change. Is it problematic to keep them as separate properties?

madsrasmussen: Hm, that's a good point. I made a combined config because it is easier to know when to initialize the app. If we have more properties I had to check across. 

Just thinking about it, maybe it could be solved with Lis updated callback? 🤔

iOvergaard: Can't we just assume that any properties are initialized on the connected callback? I mean, either they are set through an environment variable or hardcoded into an attribute on the element, right?

madsrasmussen: I have updated to use the connectedCallback 👍 . Can you have a second look?

iOvergaard: LGTM

src/app/app.element.ts
```diff
@@ -31,7 +29,7 @@ export class UmbAppElement extends UmbLitElement {
 	 * @remarks This is the base URL of the Umbraco server, not the base URL of the backoffice.
 	 */
 	@property({ type: String })
-	private serverUrl = import.meta.env.VITE_UMBRACO_API_URL ?? '';
+	serverUrl? = window.location.origin;
```
iOvergaard, line 32: I would say the default should just keep being `''` since it is being prepended to API calls and we probably just want to keep them relative in CMS land.

```suggestion
	serverUrl = '';
```

src/app/app.element.ts
```diff
@@ -40,65 +38,72 @@ export class UmbAppElement extends UmbLitElement {
 	 */
 	@property({ type: String })
 	// TODO: get from server config
-	private backofficePath = import.meta.env.DEV ? '/' : '/umbraco';
+	backofficePath? = '/umbraco';
```
iOvergaard, line 41: I think we can assume it is not potentially undefined

```suggestion
	backofficePath = '/umbraco';
```

iOvergaard, line 41: I think we can assume it is not potentially undefined

```suggestion
	backofficePath = '/umbraco';
```

src/app/app.element.ts
```diff
@@ -40,65 +38,72 @@ export class UmbAppElement extends UmbLitElement {
...
+
+	/**
+	 * Bypass authentication.
+	 * @type {boolean}
+	 */
+	// TODO: this might not be the right solution
+	@property({ type: Boolean })
+	bypassAuth? = false;
```
iOvergaard, line 49:
```suggestion
	bypassAuth = false;
```

iOvergaard (approved):

PR #711, opened 2023-05-15
Bugfix: Browser compatible MSW

madsrasmussen:


iOvergaard:
How come this works under Vite but not web-dev-server?

madsrasmussen:
I am not sure what module resolution Vite runs all files through, but they are doing something magical. It is this bug #https://github.com/mswjs/msw/issues/1384 we are running into. Hopefully esm support will be possible one day.

Until then it doesn't make a big difference. It is only code we use internally.

iOvergaard (approved):
LGTM

PR #721, opened 2023-05-16
Chore: Organize folders in apps and packages

madsrasmussen:


nielslyngsoe:
New structure is great and everything seems to still work as expected.

PR #729, opened 2023-05-19
Chore/UI bits and pieces

nathanwoulfe:
Some UI-focussed bits and pieces I found today
- `custom-properties.css` wasn't loading
- workspace footer buttons were rendered in reverse order
- language dropdown wasn't full width in sidebar
- header in tree context menu needed styling
- links from the translation dashboard had the wrong url
- used css vars for constants for layout element dimensions


src/apps/app/app.element.ts
```diff
@@ -1,4 +1,6 @@
 import '@umbraco-ui/uui-css/dist/uui-css.css';
+import '../../shared/css/custom-properties.css';
```
iOvergaard, line 2: Seems like you added this as a tsconfig path alias, so consider using that alias here.

src/packages/core/components/section/section-sidebar-context-menu/section-sidebar-context-menu.element.ts
```diff
@@ -101,16 +101,25 @@ export class UmbSectionSidebarContextMenuElement extends UmbLitElement {
...
 				top: 0;
-				width: 300px;
+				width: var(--umb-section-sidebar-layout-width);
 				border: none;
 				border-left: 1px solid var(--uui-color-border);
 				border-right: 1px solid var(--uui-color-border);
 				background-color: var(--uui-color-surface);
 			}
+
+			#action-modal h3 {
```
iOvergaard, line 115: Consider targeting a class rather than all h3 elements so people can opt into this styling

src/packages/core/components/tree/context-menu/tree-context-menu.service.ts
```diff
@@ -78,16 +78,25 @@ export class UmbTreeContextMenuServiceElement extends UmbLitElement {
...
 				top: 0;
-				width: 300px;
+				width: var(--umb-section-sidebar-layout-width);
 				border: none;
 				border-left: 1px solid var(--uui-color-border);
 				border-right: 1px solid var(--uui-color-border);
 				background-color: var(--uui-color-surface);
 			}
+
+			#action-modal h3 {
```
iOvergaard, line 92: Looks like all styling is duplicated here as well as [src/packages/core/components/section/section-sidebar-context-menu/section-sidebar-context-menu.element.ts](src/packages/core/components/section/section-sidebar-context-menu/section-sidebar-context-menu.element.ts) for `#action-modal`. Can we do something about that?

iOvergaard (changes requested):

src/packages/core/components/tree/context-menu/tree-context-menu.service.ts
nathanwoulfe: Yeah, noticed that at the time, I'll see what can be done to clean it up

src/packages/core/components/section/section-sidebar-context-menu/section-sidebar-context-menu.element.ts
nielslyngsoe: Should not be a problem here cause of Shadow-DOM. I'll go ahead and get these corrections merged and then we can always revisit the h3 selector. :-)

src/packages/core/components/tree/context-menu/tree-context-menu.service.ts
nielslyngsoe: I cannot find the usage of these files anywhere, looks like legacy — so I have removed them.

iOvergaard (approved):

PR #740, opened 2023-05-30
no ** redirects and re-render routes.

nielslyngsoe:
No '**' redirects as routes are registered over time(async), instead, this comes with a new build of the router-slot. Which allows for routes to appear async, re-rendering everytime routes are set until it finds a match.


iOvergaard (approved):
LGTM

PR #752, opened 2023-06-01
fix storybook build

iOvergaard:
* Moved monaco-editor to its own external module
* Added build for monaco's workers so they land up in a place where they can be imported and made sure the syntax is the same as that from vite, so storybook build does not break
* Made sure Monaco's CSS can be imported dynamically with rollup CSS plugin


nielslyngsoe (approved):
Changes looks good

julczka (approved):

PR #760, opened 2023-06-05
Feature/logout

iOvergaard:
## Changes
* Add export for UmbAuthFlow
* Add export for UmbAppContext
* Set up user modal to log out by deleting token and redirecting the user to the root

## Others
* Also added example of how to work with our access token from outside now that it's generally available to extension authors (it wasn't before because UmbAuthFlow was hidden)
  * Added UmbAuthFlow with an interface because we do not necessarily want to expose UmbAuthFlow directly in the future when we have a current user context
* Added UmbAppContext to extension authors where they among other things can get the server URL for manual fetch requests


src/shared/auth/auth-flow.ts
```diff
@@ -32,6 +33,8 @@ import {
 
 const requestor = new FetchRequestor();
 
+const TOKEN_RESPONSE_NAME = 'umb:tokenResponse';
```
nielslyngsoe, line 36: Could we provide a name here that is a bit more specific, like userAuthTokenResponse ?

iOvergaard: Done

nielslyngsoe (approved):
Code looks good :-)

PR #763, opened 2023-06-06
Feature/general styling

JesmoDev:
Fixes a bunch of styling issues in the backoffice.
Also makes following changes to the umb-body-layout:

Adds default padding to the main and header containers (can be removed with `main-no-padding` or  `header-no-padding` attributes)

Added attributes: 
* `header-transparent`: removes the background color and border on the header. It also alters the top padding on the main container to make it sit nicely with an empty header. This also ads a shadow when scrolling.
* `main-no-padding`: removes all padding in the main container (takes precedence over any other attribute)
* `header-no-padding`: removes all padding in the header container (takes precedence over any other attribute)
* `header-fit-height`: sets the height of the header to fit-content


src/packages/settings/logviewer/workspace/views/search/components/log-viewer-message.element.ts
```diff
@@ -230,6 +236,30 @@ export class UmbLogViewerMessageElement extends UmbLitElement {
...
+				border-radius: var(--uui-border-radius);
+				background-color: var(--uui-color-background);
+				margin-bottom: var(--uui-size-space-3);
+			}
+
+			pre {
+				border-left: 4px solid #d42054;
+				color: #303033;
+				display: block;
+				font-family: Lato, Helvetica Neue, Helvetica, Arial, sans-serif;
```
nielslyngsoe, line 256: Dont we have some variable from UUI that we can use here?

nielslyngsoe: I see we don't.. lets keep this for now.

nielslyngsoe: I made a AB task for it for the UUI to export.

nielslyngsoe (approved):
Looks good, THANKS!

PR #767, opened 2023-06-06
Feature/document-type-crud

nielslyngsoe:
New APIs + Document Type Save and Update implementation.


src/mocks/handlers/document-type.handlers.ts
```diff
@@ -9,9 +9,12 @@ export const handlers = [
...
 		if (!data) return;
 
-		umbDocumentTypeData.insert(data);
+		// TODO: This is something that is missing in the Full model, but which we need to for the tree model. This should be fixed in the Full model.
+		data.parentId ??= null;
 
-		return res(ctx.status(200));
+		const created = umbDocumentTypeData.insert(data);
+
+		return res(ctx.status(200), ctx.set({'location': '/header/location/id/'+created.id}));
```
iOvergaard, line 17: Should the header value be such a weird URL-ish string? I believe the backend controller will send back the URL of the Management API that you need to call to get the new item?

iOvergaard (commented):
I can create document types and edit them.

I get an error if I try to delete a type:
```
Uncaught (in promise) TypeError: this.repository.requestItems is not a function
    at UmbTrashEntityAction.execute (trash.action.ts:27:42)
```

When I create a new type, the browser URL does not reflect the real path because it gets two times `/umbraco/umbraco/` prepended to it. I guess there's something where the redirect should not start with a `/` perhaps.

The property editors are not being rendered in the document type workspace. It's just a grey box. I don't know if that is within the scope of this feature?

I can only add one tab to a document type, but edit/delete works for that tab.

I can't rename groups, nor can I add more than 1 group. Probably working as intended for now.

I guess the code is merge-able, so it should be up to you how much of this you wish to fix now versus later, Niels.

nielslyngsoe:
Deleting does not work cause the action is trash and not Delete. And that action is missing. I might remove the trash option for now.

Url thing sounds like a bug, I'll have a look.

Rendering just a grey box for the property editor is fully on purpose for now.

New tabs are just named 'New' and currently tabs merge based on the name. I guess it should either prevent getting/setting an existing name or take more values into consideration when merging containers.

Groups names have been left out on purpose so far, but I see that it feels off. Should be a rather simple one, I'll note it down.

src/mocks/handlers/document-type.handlers.ts
nielslyngsoe: I've written a comment above.

iOvergaard (approved):
LGTM

PR #778, opened 2023-06-12
Update translation

JesmoDev:
Updates Create so that it works and redirects the user to the newly created trasnlation


nielslyngsoe (approved):
Now it works on the server, but we still miss to make this work with our mock data.

PR #780, opened 2023-06-16
Feature/general styling v2

JesmoDev:
Added a pre slot to the umb-body-layout that sits before the header slot. This slot is currently used for the icon picker and back navigation arrows:
image image


nielslyngsoe:
I have wrote some comments/thoughts in danish to Jesper on Slack :-)

PR #783, opened 2023-06-19
Chore/naming of pickers and inputs

loivsen:
**Renaming some elements:**
- input-document-picker is now input-document
- input-document-type-picker is now input-document-type
- input-color-picker is now input-color
- input-media-picker is now input-media
- date-input is now input-date to match the rest of inputs

**Things to notice:**
The original assignment was renaming the inputs from for example input-document-picker to document-input (removing the picker name, and moving input to the back of the name), however in the meanwhile a lot of the inputs have been renamed to starting with input rather than ending with input. For now, I have continued this trend (starting with input), but it can be changed if we rather want to end with it.

The input-language-picker has **not** been renamed, as it seems to have some picker logic within the file (picker context) so Im not sure if the name is fitting or not.


iOvergaard (approved):
LGTM

PR #786, opened 2023-06-22
use the asPromise() method for generating await init promises

nielslyngsoe:
Cleans up old ways of making a init promise, instead using the asPromise() method for generating await init promises.

### DIFF
--- a/src/packages/core/picker-input/picker-input.context.ts
+++ b/src/packages/core/picker-input/picker-input.context.ts
@@ -49,7 +49,7 @@ export class UmbPickerInputContext<ItemType extends ItemResponseModelBaseModel> 
 		this.selectedItems = this.#itemManager.items;
 
 		this.#init = Promise.all([
-			this.#itemManager.init,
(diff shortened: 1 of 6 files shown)


iOvergaard (changes requested):
The build is complaining about an illegal accessor for a private property

src/packages/core/picker-input/picker-input.context.ts
```diff
@@ -49,7 +49,7 @@ export class UmbPickerInputContext<ItemType extends ItemResponseModelBaseModel>
 		this.selectedItems = this.#itemManager.items;
 
 		this.#init = Promise.all([
-			this.#itemManager.init,
+			this.#itemManager.#init,
```
iOvergaard, line 52:
```suggestion
			this.#itemManager.init,
```

iOvergaard (changes requested):

iOvergaard (approved):

PR #790, opened 2023-06-23
TinyMCE Toolbar Configuration

iOvergaard:
## Types of changes

## Checklist


nathanwoulfe:
@iOvergaard this looks grand - will pull the branch and take it for a spin. Wondering too if we can use the toolbar meta to automatically register plugins in the editor rather than relying on the plugin to do it internally... Would mean a little less convention in the plugins, and allow us to handle registration consistently.

nathanwoulfe (approved):
Works beautifully. Will no pursue auto registration of plugins - it's possible for a plugin to have multiple toolbar buttons, so we can't have an interface defining the toolbar action method, which would allow auto-registration. Instead, can leave it to the implementor to decide how they want their plugin rendered.

iOvergaard:
@nathanwoulfe 
> it's possible for a plugin to have multiple toolbar buttons, so we can't have an interface defining the toolbar action method
Glad we landed on the same conclusion then. There are so many things you can register with a Tiny plugin, so adding only an extra extension type for a toolbar button is quite limiting, and also very difficult to extract the types for the different callbacks (I went down that route already).

bjarnef:
@iOvergaard maybe the "Font size" and "Font select" should use some custom icons to differ from "Text color" which TinyMCE has an icon for in toolbar.

Font size on websites typical has a small/large letter: 
https://www.flaticon.com/free-icon/decrease-font-size_14631
https://icon-icons.com/icon/font-size/99704

Font select could maybe just be an "A" (or "T").
Some suggestions here:
https://fontawesome.com/search?q=font&o=r

I think it would be great the icons represent the same icons in toolbar when possible, but some commands doesn't have toolbar icons. In that case we could use a custom icon.

The "Source code editor" in the ace editor in Umbraco, but it is also possible to enable the "Code" command (the default code editor in TinyMCE), which doesn't seem to be added by default. In older versions of Umbraco both was registered.

image

Still possible to enable both - in that case I guess it is fine to use same icon for both code editors (unless we could use a slightly different icon for the ace code editor).

iOvergaard:
@bjarnef 
Yes you are totally right.

Regarding code, I was a little in doubt. The current backoffice does some rewrites of the command to seem to fit it into Umbraco's own "ace" command. There's also a premium code editor for Tiny that some might use, so maybe we should be really careful about replacing Tiny's own commands. But we already replace them for stuff like "link" and "image" so it's a hot mess.

We also have to care about migrating from older versions of Umbraco, so can't just rename things, but we can certainly add new commands/toolbar buttons.

PR #791, opened 2023-06-26
Chore/UI styles

nathanwoulfe:
Noticed when scrolling content or datatype views in vertical layout the editor was visible behind the header. These changes add a background to the header and a pseudo-element to mask overflow.

Also tightens whitespace in `workspace-property-layout` to ensure even whitespace above and below all instances (reduces padding when element is in a content node as the containing box adds padding-Y).

Finally, tweaks TinyMCE styles to match editor borders from UUI - ie squared corners, color, border width.


nielslyngsoe:
Hi @nathanwoulfe I really love your focus on the details and love your effort to bring everything to a production ready quality.
I do though have a bit of comments on the way these have been carried out. And I hope you will appreciate my corrections, as learning into how we separate concerns with Web Components, more specifically Shadow DOM.

I hope my comment in the code will survive though I will be correcting the code as well. :-)

src/packages/core/workspace/workspace-property-layout/workspace-property-layout.element.ts
```diff
@@ -80,18 +86,28 @@ export class UmbWorkspacePropertyLayoutElement extends LitElement {
 				grid-column: span 2;
 			}
 
+			:host > div:last-child {
```
nielslyngsoe, line 89: This selector feels unsafe to me.
When reading it, one does not easily capture what its suppose to do. Therefor is some one day comes a head and adds another div then this would break.

My solution on this one is to add an ID to the div, like #editorColumn.

src/packages/core/workspace/workspace-property-layout/workspace-property-layout.element.ts
```diff
@@ -80,18 +86,28 @@ export class UmbWorkspacePropertyLayoutElement extends LitElement {
...
+					margin-top:0;
 				}
 			}
 
 			:host(:last-of-type) {
 				border-bottom: none;
 			}
 
+			/* container has 18px padding, reduce here to ensure consistency */
 			:host-context(umb-variantable-property:first-of-type) {
```
nielslyngsoe, line 105: I can see this :host-context was here already, but I would not recommend using such.

With the perspective of separation of concerns, then we do not want web-components to be aware about other web-components. Especially not linking them by element name.

Looking at the problem I would say its a implementation problem within `umb-document-workspace-view-edit-tab`.

Here we have the ability to overwrite the padding of uui-box, via an CSS Custom Prop. in this way:

`
uui-box {
	--uui-box-default-padding: 0 var(--uui-size-space-5);
}
`

src/packages/core/workspace/workspace-property-layout/workspace-property-layout.element.ts
```diff
@@ -102,7 +118,22 @@ export class UmbWorkspacePropertyLayoutElement extends LitElement {
 				position: sticky;
 				top: var(--uui-size-space-4);
 				height: min-content;
-				z-index: 2;
+				z-index: 1050;
```
nielslyngsoe, line 121: May I ask why you want this super high z-index number?

I know its typical CSS before shadow DOM to do so. But I would really like us to avoid such. So please go with the lowest one that does what we need.

Maybe another time, add a comment to describe the value. Is it because it must go on top of something?

nielslyngsoe: Ah, I just realized its because it's sticky in the small-mode/mobile version.

Thats a mistake, we do not want the label to float with. cause that creates al sorts of trouble if the editor it self had an action bar that was sticky.

In the case of the RTE, it would make more sense to have the toolbar following the screen. :-)

So I will get rid of this.

nielslyngsoe (approved):
Please read my comments, see the changes I made, and check if you are happy with the PR.

I disabled auto-merge so you can approve it this way. And I have approved the PR. So Just merge if you are happy :-)

nathanwoulfe:
Thanks Niels, that all makes sense - appreciate your holistic view of how this all works together 😄

PR #792, opened 2023-06-27
Remove tokens in INSTALL state

iOvergaard:
## Description

Due to needs from the backend, we should not store any tokens on the installer, since they will be invalid anyway.

## Changes

* extract auth initialization to another method and only invoke it if we are not in INSTALL state
* clear any cached tokens if we are in INSTALL state


nielslyngsoe (approved):
Code looks good

PR #795, opened 2023-06-29
Property Editor Structure and Types

nielslyngsoe:
* Correct folder name to match import map.
* Make property editor schema alias fallback to Umbraco.JSON if not defined.
* Make pathname for dashboard and section view optional. instead it will be generated from the label/name


src/packages/core/section/section-views/section-views.element.ts
```diff
@@ -141,16 +142,17 @@ export class UmbSectionViewsElement extends UmbLitElement {
 		return this._dashboards.length > 0
 			? html`
 					<uui-tab-group slot="header" id="dashboards">
-						${this._dashboards.map(
-							(dashboard) => html`
+						${this._dashboards.map((dashboard) => {
+							const dashboardName = dashboard.meta.label ?? dashboard.name;
+							const dashboardPath =
+								'dashboard/' + dashboard.meta.pathname ? dashboard.meta.pathname : umbracoPath(dashboardName);
```
iOvergaard, line 148: I think `umbracoPath()` is used for our msw handlers to prepend the management API path to those, so wouldn't the path look like: `/umbraco/api/v1/.....` if they didn't provide a fallback?

src/packages/core/section/section-views/section-views.element.ts
```diff
@@ -141,16 +147,17 @@ export class UmbSectionViewsElement extends UmbLitElement {
 		return this._dashboards.length > 0
 			? html`
 					<uui-tab-group slot="header" id="dashboards">
-						${this._dashboards.map(
-							(dashboard) => html`
+						${this._dashboards.map((dashboard) => {
+							const dashboardName = dashboard.meta.label ?? dashboard.name;
+							const dashboardPath =
```
iOvergaard, line 152: Looks like we are constructing this path more than once. This could be prone to errors. Have you considered making a common function to construct the dashboardPath?

src/packages/core/section/section-views/section-views.element.ts
```diff
@@ -13,6 +13,7 @@ import {
 import { createExtensionElement } from '@umbraco-cms/backoffice/extension-api';
 import { UmbLitElement } from '@umbraco-cms/internal/lit-element';
 import { UmbObserverController } from '@umbraco-cms/backoffice/observable-api';
+import { pathFolderName } from 'src/shared/utils/path-folder-name.function.js';
```
iOvergaard, line 16: Action fails on this line. The import path should use an alias.

iOvergaard (changes requested):

src/packages/core/section/section-views/section-views.element.ts
nielslyngsoe: Wups, Corrected :-)

src/packages/core/section/section-views/section-views.element.ts
nielslyngsoe: Yeah, for sure. I have made two methods for it.

We might come up with some smarter methods, enabling similar for multiple other implementations.

As I think manifests, only should be defining pathnames, if the developer likes to be specific. otherwise, we should just generate it from data. So I think we will see more of this in the future. :-)

iOvergaard (approved):

PR #816, opened 2023-07-24
Feature/rename to dictionary

iOvergaard:
Rename the folder holding the dictionary in Umbraco to "dictionary" to avoid clashing with the future translation/localization package.


nielslyngsoe (approved):
Much better, I'm very happy with the change.

PR #817, opened 2023-07-25
Feature/localization part 1

iOvergaard:
## Description

This is step 1 of the RFC for localization: https://github.com/umbraco/Umbraco.CMS.Backoffice/discussions/818

**Storybook:** [There is a new documentation page which you can see here](https://ambitious-stone-0033b3603-817.westeurope.1.azurestaticapps.net/?path=/docs/guides-localization--docs).

## What changed

**New element: umb-localize**
```javascript
render() {
  return html`<umb-localize key="general_close"></umb-localize>`
}
```

**New context: UmbLocalizationContext**
Has two methods `localize` and `localizeMany` that return an observable with the localized values of the requested keys.

## Types of changes

## Motivation and context

This allows UI developers to add localized strings to their UI, which is one of the success criteria for migrating to the new Backoffice.

## How to test?

Add a new UI element (or amend an existing one) and test by one of these scenarios:

**Test the render function (should render "Close" in the browser)**
```javascript
render() {
  return html`<umb-localize key="general_close"></umb-localize>`
}
```

**Test the UmbLocalizeController**
```javascript
render() {
  return html`<uui-button .label=${this.localize.term('general_close')}></uui-button>`
}
```

## Screenshots

Changing the user language updates a few strings:

https://github.com/umbraco/Umbraco.CMS.Backoffice/assets/752371/d7f775d7-b805-433d-943b-c70b7351d4e1

## Checklist


src/libs/localization-api/localization.context.ts
```diff
@@ -0,0 +1,64 @@
...
+	 * @param fallbackLanguageIsoCode The fallback language to use (example: 'en-us', default: 'en-us')
+	 */
+	setLanguage(languageIsoCode: string, fallbackLanguageIsoCode?: string) {
+		this.#translationRegistry.register(languageIsoCode, fallbackLanguageIsoCode);
+	}
+
+	/**
+	 * Localize a key.
+	 * If the key is not found, the fallback is returned.
+	 * If the fallback is not provided, the key is returned.
```
ronaldbarendse, line 31: The code below returns an empty string instead of the key, so the documentation should be updated to:
```suggestion
	 * If the fallback is not provided, an empty string is returned.
```

Same for the `localizeMany` documentation...

src/libs/localization-api/localization.context.ts
```diff
@@ -0,0 +1,64 @@
...
+	 *
+	 * @description This method combines the results of multiple calls to localize.
+	 * @param keys The keys to localize. The keys are case sensitive.
+	 * @example localizeMany(['general_close', 'general_logout']).subscribe((values) => {
+	 * 	console.log(values[0]); // 'Close'
+	 * 	console.log(values[1]); // 'Log out'
+	 * });
+	 * @see localize
+	 */
+	localizeMany(keys: string[]) {
```
ronaldbarendse, line 59: Should this also support fallback values?

src/libs/localization-api/registry/translation.registry.ts
```diff
@@ -0,0 +1,70 @@
...
+
+	constructor(umbExtensionRegistry: UmbBackofficeExtensionRegistry) {
+		this.#extensionRegistry = umbExtensionRegistry;
+	}
+
+	get translations() {
+		return this.#innerDictionary.asObservable();
+	}
+
+	register(userCulture: string, fallbackCulture = 'en-us') {
```
ronaldbarendse, line 21: Wouldn't it make more sense to use a neutral culture (without any country/region, like `en`) or even the invariant culture (.NET uses an [empty string as name](https://learn.microsoft.com/en-us/dotnet/api/system.globalization.cultureinfo.invariantculture?view=net-7.0#remarks) or [`iv` as ISO code](https://learn.microsoft.com/en-us/dotnet/api/system.globalization.cultureinfo.twoletterisolanguagename?view=net-7.0#remarks))?

src/packages/core/extension-registry/models/translations.model.ts
```diff
@@ -0,0 +1,26 @@
...
+	 * @example
+	 * {
+	 *   "general": {
+	 *     "cancel": "Cancel",
+	 *     "close": "Close"
+	 *   }
+	 * }
+	 */
+	translations?: Record<string, Record<string, string>>;
+}
```
ronaldbarendse, line 26: This doesn't allow adding translations for multiple cultures in a single manifest.

src/packages/users/current-user/modals/current-user/current-user-modal.element.ts
```diff
@@ -25,15 +32,16 @@ export class UmbCurrentUserModalElement extends UmbLitElement {
...
-			this.#auth = instance;
-		});
-
 		this.consumeContext(UMB_APP, (instance) => {
 			this.#appContext = instance;
 		});
 
-		this._observeCurrentUser();
+		this.consumeContext(UMB_LOCALIZATION_CONTEXT, (instance) => {
+			instance.localizeMany(['general_close', 'general_logout']).subscribe((values) => {
```
ronaldbarendse, line 40: Wouldn't it be nicer to use [destructuring assignment](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment) for the `values` array, so you don't have to use indexes (that are very prone to typos and make it harder to add new keys before/between existing ones?

ronaldbarendse: If it's possible to provide direct mappings, that would make it even better... Something like:
```ts
instance.localizeMap({
  'general_close': value => this.labelClose = value,
  'general_logout': value => this.labelLogout = value
});
```

iOvergaard:
Thank you for your comments, @ronaldbarendse. I'm taking this back to draft, though, because we are reworking some parts of it to better support Lit/Web Components :)

src/packages/users/current-user/modals/current-user/current-user-modal.element.ts
iOvergaard: Yes, but that is not backwards compatible with the current backoffice. The idea was that people had to rework as little as possible. Although, now we are looking into a different API so things might change and we probably won't need methods such as `localizeMany` moving forward.

src/packages/core/extension-registry/models/translations.model.ts
iOvergaard: This is on purpose. You have to define the culture in your manifest. If you want to define multiple cultures, you simply register another manifest, which you can do even in the same file.

src/libs/localization-api/localization.context.ts
iOvergaard: Probably a good idea, but the old method didn't and we are probably going to rework this anyway.

src/libs/localization-api/registry/translation.registry.ts
iOvergaard: Yes, but we had some trouble with the cultures of the backend, where the "en.xml" file is actually "en-gb", but the default is "en-us". I am currently reworking this into listening on the "lang" attribute in the browser, so makes sense to support both with and without the "region" part of the culture code, but there is a discrepancy with the backend here and not sure how to easily fix that.

ronaldbarendse: Hmm, on the server-side we can use [`CultureInfo.Parent`](https://learn.microsoft.com/en-us/dotnet/api/system.globalization.cultureinfo.parent?view=net-7.0) to get the fallback language and go from e.g. `en-GB` to `en`, so you don't have to manually provide the fallback language when defining the translations. And if it can't find translations within this fallback chain, it can use the app's default culture that's specified in the [`NeutralResourcesLanguageAttribute`](https://learn.microsoft.com/en-us/dotnet/api/system.resources.neutralresourceslanguageattribute?view=net-7.0).

It doesn't seem like browsers have a similar way to get 'parent' cultures, although [`Intl.Locale`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Locale) splits the language, script and region, so you could do the following:
```js
function getParent(tag) {
  var locale = new Intl.Locale(tag);
  if (locale.region && locale.script) return locale.language + '-' + locale.script;
  if (locale.region || locale.script) return locale.language;
  return null;
}

getParent('en-GB'); // en
getParent('en'); // null

getParent('zh-Hans-CN'); // zh-Hans
getParent('zh-Hans'); // zh
getParent('zh'); // null
```

The client side could then also define a similar neutral/default culture (as a single configuration value), in case the preferred culture can't be retrieved from the authenticated user or [`navigator.language`](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/language) (or the first supported language from [`navigator.languages`](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/languages)).

src/external/rxjs/index.ts
```diff
@@ -14,4 +14,6 @@ export {
 	of,
 	lastValueFrom,
 	firstValueFrom,
+	switchMap,
```
nielslyngsoe, line 17: Maybe I'm mistaken, but I'm not sure this is used anymore?

nielslyngsoe: and the throwError.. :-)

src/libs/localization-api/localize.controller.ts
```diff
@@ -0,0 +1,145 @@
...
+			return String(key);
+		}
+
+		if (typeof term === 'function') {
+			return term(...args) as string;
+		}
+
+		if (typeof term === 'string') {
+			if (args.length > 0) {
+				// Replace placeholders of format "%index%" and "{index}" with provided values
```
nielslyngsoe, line 118: I think, New Backoffice, gives u the option to get away from indexed arguments and make named arguments. Could we enable that while still keeping the index ones. :-) ?

nielslyngsoe: Ah, no worries.. we can postpone making that an option.

src/libs/localization-api/registry/translation.registry.ts
```diff
@@ -0,0 +1,69 @@
...
+
+							if (loadedExtension && hasDefaultExport(loadedExtension)) {
+								for (const [dictionaryName, dictionary] of Object.entries(loadedExtension.default)) {
+									this.#addOrUpdateDictionary(innerDictionary, dictionaryName, dictionary);
+								}
+							}
+
+							// Notify subscribers that the inner dictionary has changed.
+							const translation: Translation = {
+								$code: userCulture,
```
nielslyngsoe, line 41: We do not use the $ in the rest of out code, so I think we should just get rid of it here. I see it as confusing when reading the code.

src/libs/observable-api/observer.ts
```diff
@@ -1,11 +1,19 @@
 import { Observable, Subscription, lastValueFrom } from '@umbraco-cms/backoffice/external/rxjs';
 
+export type ObserverCallbackStack<T> = {
+	next: (_value: T) => void;
+	error?: (_value: unknown) => void;
+	complete?: () => void;
+};
+
+export type ObserverCallback<T> = ((_value: T) => void) | ObserverCallbackStack<T>;
```
nielslyngsoe, line 9: Cool :-)

src/packages/core/localization/localize.element.ts
```diff
@@ -0,0 +1,65 @@
...
+		return localizedValue;
+	}
+
+	protected render() {
+		return this.text
+			? html`${this.text}`
+			: this.debug
+			? html`<span style="color:red">${this.key}</span>`
+			: html`<slot></slot>`;
+	}
```
nielslyngsoe, line 58: I think we should to add the style: `display: content` to the host. Or what do you think?

src/packages/core/localization/localize.element.ts
```diff
@@ -0,0 +1,65 @@
...
+
+		return localizedValue;
+	}
+
+	protected render() {
+		return this.text
+			? html`${this.text}`
+			: this.debug
+			? html`<span style="color:red">${this.key}</span>`
+			: html`<slot></slot>`;
```
nielslyngsoe, line 57: I'm not sure I think we should allow the slot here. I think that will be hard to understand how that is composed looking the implementation of umb-localize and then something within that.

But maybe you have a use case you can describe for me?

src/packages/core/localization/manifests.ts
```diff
@@ -0,0 +1,27 @@
+import { ManifestTypes } from '../extension-registry/index.js';
+import { ManifestTranslations } from '../extension-registry/models/translations.model.js';
+
+const translationManifests: Array<ManifestTranslations> = [
+	{
+		type: 'translations',
+		alias: 'Umb.Translations.En_US',
+		weight: -100,
```
nielslyngsoe, line 8: I'm not sure what the weight will achieve. Shouldn't we just get rid of it? :-)

iOvergaard: The idea is that we are sure to load the system languages first since we want any extension languages to override ours. But I'm not sure if the weight is used automatically or not.

src/packages/core/localization/localize.element.ts
iOvergaard: It seemed to work all right without it, but I guess it couldn't hurt.

src/packages/core/localization/manifests.ts
nielslyngsoe: interesting approach — as I was only seeing weight as a sorting.

So I'm not sure the weight would have an influence on that, but I will check. :-)

src/libs/localization-api/manager.ts
```diff
@@ -0,0 +1,73 @@
...
+		const code = t.$code.toLowerCase();
+
+		if (translations.has(code)) {
+			// Merge translations that share the same language code
+			translations.set(code, { ...translations.get(code), ...t });
+		} else {
+			translations.set(code, t);
+		}
+
+		// The first translation that's registered is the fallback
```
nielslyngsoe, line 53: I'm not sure how we can guarantee this. I know we initialize the loading of the fallback language first. But I wouldn't be sure it's the first one to finish, especially if someone comes with a predefined set.

Could we either have a property in the manifest, like `default`? or an argument to parse on to this method.

src/libs/localization-api/registry/translation.registry.ts
nielslyngsoe: Okay, I understand now. I find the code a bit hard to read. And I'm a little bit allergic to that type of just mixing properties together style.

I guess I need to see how its suppose to be utilized, cause currently there is no test or code using for example the `$dir`.

Would be nice to bring an example of how someone would use that value, if there is such case. Otherwise there is ways we can get rid of these two props.

src/libs/localization-api/localize.controller.ts
```diff
@@ -0,0 +1,152 @@
...
+
+	/**
+	 * Gets the host element's language as determined by the `lang` attribute. The return value is transformed to
+	 * lowercase.
+	 */
+	lang() {
+		return `${this.#hostEl.lang || documentLanguage}`.toLowerCase();
+	}
+
+	private getTranslationData(lang: string) {
```
nielslyngsoe, line 93: This method is only being used once, and there we only use two of the returned props. I think we just just get rid of this and copy the code for retrieving primary and secondary into the method below.

src/libs/localization-api/registry/translation.registry.ts
nielslyngsoe: Lets keep it.. it might be useful for someone.. and I don't have better letter than the $

nielslyngsoe (approved):
Really cool, looks good and works well

Issue #800, opened 2023-06-30
[BUG]: Published Status - Snapshot does not warn user


iOvergaard:
**Describe the bug**
Clicking the "Snapshot Internal Cache" button does not warn the user beforehand, but the other buttons do. This is an expensive operation so it is importing that the user confirms the action.

**To Reproduce**
1. Go to Settings -> Published Status
2. Click the "Snapshot Internal Cache" button

**Expected behavior**
Expectation is that a warning box is shown to the user to confirm.

**Screenshots**

https://github.com/umbraco/Umbraco.CMS.Backoffice/assets/752371/9091ed6d-31d0-468a-ba84-5ffde4017246

PR #820, opened 2023-08-01
add warning dialog to internal cache button

bkclerke:
fix for #800 - add warning dialog to internal cache button on the settings published status dashboard. 

NOTE: Current CMS functionality does not include a warning dialog for this action so the dialog text is new.

2023-07-31 20_30_23-Umbraco - Personal - Microsoft​ Edge


iOvergaard (approved):
LGTM! We will go through the translations once we retrofit the UI elements with localization.

PR #827, opened 2023-08-03
Feature/isolate localization api

iOvergaard:
## Changes
- Ensure that libs/localization-api only contains generic API types and that the tests for those are generic as well including a generic test for a controller host element.
- Add new entry @umbraco-cms/backoffice/localization that holds the instance of the registry with tests focusing on that registry alone
- Ensure that tests around UmbLocalizeElement act more as integrations test between the 3 vectors of the localization API: UmbLocalizeController, UmbTranslationRegistry, and the extension registry since the controller host already tests the direct controller.

The main benefit of these changes is that API stuff is isolated and can be run without the backoffice (including the tests!) and that we only test the functionality between the extension registry and the translation registry once (and only in the backoffice test suite!).

Bonus: Also fixed a bug where the non-regional language actually overwrote the regional language instead of being put in there as a secondary language.


nielslyngsoe (approved):
Good! :-)

PR #838, opened 2023-08-23
Feature/document type groups

loivsen:
## Document Type Groups

- Properties can now be deleted. The delete button will appear when focusing/hovering the property editing field
- Properties that are inherited cannot be deleted, but will show where are inherited from, and has a link for a fast editing (modal route) the document type they belong to.
- Properties have tags that shows a little bit of info: Their datatype, if they vary by culture, and if label above is set to true.
- Clicking on "Add property" now opens a modal for creating a property, and will only create a new property if this is submitted.

### DIFF
--- a/src/packages/core/modal/common/property-settings/property-settings-modal.element.ts
+++ b/src/packages/core/modal/common/property-settings/property-settings-modal.element.ts
@@ -153,7 +153,7 @@ export class UmbPropertySettingsModalElement extends UmbModalBaseElement<
 		this._customValidationOptions.forEach((option) => {
 			option.selected = option.value === regEx;
 		});
-		console.log(this._customValidationOptions);
+		//console.log(this._customValidationOptions);
 
 		this._returnData.validation!.regEx = regEx ?? null;
 		this.requestUpdate('_returnData');
--- a/src/packages/documents/document-types/workspace/views/design/document-type-workspace-view-edit-properties.element.ts
+++ b/src/packages/documents/document-types/workspace/views/design/document-type-workspace-view-edit-properties.element.ts
@@ -12,6 +12,7 @@ import {
 	PropertyTypeModelBaseModel,
 } from '@umbraco-cms/backoffice/backend-api';
 import { UMB_WORKSPACE_CONTEXT } from '@umbraco-cms/backoffice/workspace';
+import { UMB_PROPERTY_SETTINGS_MODAL, UmbModalRouteRegistrationController } from '@umbraco-cms/backoffice/modal';
 const SORTER_CONFIG: UmbSorterConfig<DocumentTypePropertyTypeResponseModel> = {
 	compareElementToModel: (element: HTMLElement, model: DocumentTypePropertyTypeResponseModel) => {
 		return element.getAttribute('data-umb-property-id') === model.id;
@@ -84,6 +85,9 @@ export class UmbDocumentTypeWorkspaceViewEditPropertiesElement extends UmbLitEle
 	@state()
 	_ownerDocumentTypes?: DocumentTypeResponseModel[];
 
+	@state()
+	protected _modalRouteNewProperty?: string;
+
 	constructor() {
 		super();
 
@@ -96,6 +100,19 @@ export class UmbDocumentTypeWorkspaceViewEditPropertiesElement extends UmbLitEle
 			this._propertyStructure = propertyStructure;
 			this.#propertySorter.setModel(this._propertyStructure);
 		});
+
+		// Note: Route for adding a new property
+		new UmbModalRouteRegistrationController(this, UMB_PROPERTY_SETTINGS_MODAL)
+			.addAdditionalPath('new-property')
+			.onSetup(() => {
+				return {};
+			})
+			.onSubmit((result) => {
+				this.#addProperty(result);
+			})
+			.observeRouteBuilder((routeBuilder) => {
+				this._modalRouteNewProperty = routeBuilder(null);
+			});
 	}
 
 	connectedCallback(): void {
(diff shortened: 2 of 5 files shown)


src/packages/documents/document-types/workspace/views/design/document-type-workspace-view-edit-property.element.ts
```diff
@@ -124,21 +187,7 @@ export class UmbDocumentTypeWorkspacePropertyElement extends UmbLitElement {
 							.value=${this.property.name}
 							@input=${this.#onNameChange}></uui-input>
 						<!-- TODO: should use UUI-LOCK-INPUT, but that does not fire an event when its locked/unlocked -->
```
nielslyngsoe, line 189: This comment is related to the uui-input so I would say it should be moved into the render method.

src/packages/documents/document-types/workspace/views/design/document-type-workspace-view-edit-property.element.ts
```diff
@@ -40,11 +49,36 @@ export class UmbDocumentTypeWorkspacePropertyElement extends UmbLitElement {
...
 
+	@state()
+	protected _editDocumentTypePath?: string;
+
+	@property()
+	public get modalRoute() {
+		return this._modalRoute;
+	}
+
+	@property({ type: String, attribute: 'data-owner-document-type-id' })
```
nielslyngsoe, line 68: I would say we can remove the word 'data' from these two. Makes it easier to read.

src/packages/documents/document-types/workspace/views/design/document-type-workspace-view-edit-property.element.ts
```diff
@@ -72,26 +119,42 @@ export class UmbDocumentTypeWorkspacePropertyElement extends UmbLitElement {
...
+		};
+
+		const modalHandler = this._modalManagerContext?.open(UMB_CONFIRM_MODAL, Message);
+
+		modalHandler
+			?.onSubmit()
+			.then(() => {
+				this.dispatchEvent(new CustomEvent('property-delete'));
+			})
+			.catch(() => {
```
nielslyngsoe, line 153: I don't think we need this catch part. So we can just remove it.

src/packages/core/modal/common/property-settings/property-settings-modal.element.ts
```diff
@@ -153,7 +153,7 @@ export class UmbPropertySettingsModalElement extends UmbModalBaseElement<
 		this._customValidationOptions.forEach((option) => {
 			option.selected = option.value === regEx;
 		});
-		console.log(this._customValidationOptions);
+		//console.log(this._customValidationOptions);
```
nielslyngsoe, line 156: Just remove this line completely, no need for these console.logs(), I think its left over from some debugging.

src/packages/documents/document-types/workspace/views/design/document-type-workspace-view-edit-properties.element.ts
```diff
@@ -77,48 +82,89 @@ export class UmbDocumentTypeWorkspaceViewEditPropertiesElement extends UmbLitEle
...
-				(workspaceContext as UmbDocumentTypeWorkspaceContext).structure
+				(workspaceContext as UmbDocumentTypeWorkspaceContext).structure,
 			);
 		});
 		this.observe(this._propertyStructureHelper.propertyStructure, (propertyStructure) => {
 			this._propertyStructure = propertyStructure;
 			this.#propertySorter.setModel(this._propertyStructure);
 		});
+
+		// Note: Route for adding a new property
```
nielslyngsoe, line 104: Very nice approach

src/packages/documents/document-types/workspace/views/design/document-type-workspace-view-edit-properties.element.ts
```diff
@@ -77,48 +82,89 @@ export class UmbDocumentTypeWorkspaceViewEditPropertiesElement extends UmbLitEle
...
-	async #onAddProperty() {
-		const property = await this._propertyStructureHelper.addProperty(this._containerId);
-		if (!property) return;
+	connectedCallback(): void {
+		super.connectedCallback();
+		const doctypes = this._propertyStructureHelper.getOwnerDocumentTypes();
+		if (!doctypes) return;
+		this.observe(doctypes, (documents) => {
+			this._ownerDocumentTypes = documents;
+		});
```
nielslyngsoe, line 124: Since connectedCallback can happen multiple times, we should add an identifier to this observe. its a third argument like , 'observeOwnerDocumentTypes')

src/packages/documents/document-types/workspace/views/design/document-type-workspace-view-edit-properties.element.ts
nielslyngsoe: I have added the scaffold data that was missing here. :-)

nielslyngsoe:
Very good, I have added a few minor comments to change. When that's done then it should be good to get merged :-)

Please also notice I've added some code to retrieve the property scaffolds data for the property creator modal. Check if you think that is done in the right way.

Thanks

src/packages/documents/document-types/workspace/views/design/document-type-workspace-view-edit-property.element.ts
loivsen: When there is no catch, an error appears in the console when pressing cancel when trying to delete.

loivsen: It doesn't make an actual difference other than just removing the console error 😊

nielslyngsoe: Oh, okay. makes good sense. I'll have a look if we can avoid that.

PR #850, opened 2023-08-30
Feature/Stylesheets

julczka:
## Types of changes

## Checklist


src/mocks/data/utils.ts
```diff
@@ -80,3 +80,9 @@ export const createTextFileItem = (item: any): TextFileResponseModelBaseModel =>
 	name: item.name,
 	content: item.content,
 });
+
+export const createItem = (item: any): any => ({
```
nielslyngsoe, line 84: This is a bit too lose name in my opinion. What type of item is this?

Plus I don't seem to spot where it begin used? so could you confirm what purpose it has? :-) Thanks

julczka: So this item mathches this interface https://github.com/umbraco/Umbraco.CMS.Backoffice/blob/dd443a58013869713047d6cc8d594677588f6658/src/external/backend-api/src/models/ScriptItemResponseModel.ts#L8
Since I am not really sure what type of item this is I am renaming it to the server model

nielslyngsoe:
I get some issues when running this locally with Mocked data.

image

There are a few end-points that do not respond.

I also get this issue from Vite when running the dev server:

image

Would you be able to have a look at that, thanks :-)

nielslyngsoe:
It's like the router of the Stylesheet workspace is not connected properly, cause when switching around I stay on the same tab. though Url does not specify it anymore?

Also switching around after a switch between the two seems broken?

Also when saving I get an error. BTW. I'm testing on mocked data.

I'm also feeling like missing something on the RTE DataType, but maybe that's another feature?
image

julczka:
> 



> It's like the router of the Stylesheet workspace is not connected properly, cause when switching around I stay on the same tab. though Url does not specify it anymore?

It is a bug in the workspace tabs - url does not change in the tabs when you switch to a different item from the tree. I was told to leave that

nielslyngsoe (approved):
Great, I'm happy to get it in now :-)

PR #858, opened 2023-09-05
backoffice provided UmbTextStyle

nielslyngsoe:
To enable Backoffice control of styles in packages, we should provide our own default for styles. For now it just wraps the uui-text one.


iOvergaard:
Would you clarify why we import the text styles in all those components?

nielslyngsoe:
@iOvergaard My answer will be 'No' :-P  — I also tried to clean up a ton of components as part of this PR.

iOvergaard (approved):

PR #859, opened 2023-09-06
Bugfix: keep d.ts files for external libs

madsrasmussen:
We were missing the type declaration files for the external libs, which made it problematic to work with typescript. Ex. A package developers wants to extend some element from the UI Library.

The problem was that we deleted the entire external folder (including d.ts. files) when preparing the external libs for the browser through rollup.

I have made a small change that only removes the non browser compatible .js files, before creating the correct .js files. This leaves all the d.ts files behind and typescript is able to pick them up.

I have done a small prepack test of the final output which you can see in the screenshots below.

Be aware, if you set up a custom tsconfig.json file. `moduleResolution` should be set to `bundler` to support the exports field in the backoffice npm package.

Screenshot 2023-09-06 at 10 14 16

Screenshot 2023-09-06 at 10 12 06

### DIFF
--- a/src/rollup.config.js
+++ b/src/rollup.config.js
@@ -3,42 +3,51 @@ import commonjs from '@rollup/plugin-commonjs';
 import { nodeResolve } from '@rollup/plugin-node-resolve';
 import css from 'rollup-plugin-import-css';
 import webWorkerLoader from 'rollup-plugin-web-worker-loader';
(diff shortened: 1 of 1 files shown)


nielslyngsoe (commented):
The code looks good, I'll hand it over to @iOvergaard  to do another review.

iOvergaard (approved):

iOvergaard:
Looks good. Approved.

PR #863, opened 2023-09-07
Bugfix: Return entity type from workspace context method

madsrasmussen:
Currently, the user group workspace editor can't render a user group because it doesn't return anything from the getEntityType method. The result is that the workspace context never gets consumed.

This PR quick fixes the problem, and returns the 'user-group' entityType from the workspace context. Ideally, we should get the entity type from the User Group Repository, because a workspace could consist of multiple entity types, but I think we need to change this in many places.


nielslyngsoe (approved):
Beautiful

PR #872, opened 2023-09-11
profiling reflecting correct status

loivsen:
The profiling toggle will now not toggle in case of API error, to reflect the current status.
Also added localizations.

### DIFF
--- a/src/packages/settings/dashboards/performance-profiling/dashboard-performance-profiling.element.ts
+++ b/src/packages/settings/dashboards/performance-profiling/dashboard-performance-profiling.element.ts
@@ -1,5 +1,5 @@
-import { UmbTextStyles } from "@umbraco-cms/backoffice/style";
-import { css, html, customElement, state } from '@umbraco-cms/backoffice/external/lit';
+import { UmbTextStyles } from '@umbraco-cms/backoffice/style';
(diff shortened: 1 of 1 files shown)


src/packages/settings/dashboards/performance-profiling/dashboard-performance-profiling.element.ts
```diff
@@ -1,88 +1,68 @@
...
 		const { error } = await tryExecuteAndNotify(
 			this,
-			ProfilingResource.putProfilingStatus({ requestBody: { enabled: !this._profilingStatus } })
+			ProfilingResource.putProfilingStatus({ requestBody: { enabled: !this._profilingStatus } }),
 		);
 
-		if (!error) {
-			this._profilingStatus = !this._profilingStatus;
-		}
+		error ? this.#setToggle(this._profilingStatus) : this.#setToggle(!this._profilingStatus);
```
iOvergaard, line 41: Please use the `if (error) {}` notation. It makes the code easier to read ☺️

iOvergaard (changes requested):

src/packages/settings/dashboards/performance-profiling/dashboard-performance-profiling.element.ts
loivsen: done :)

PR #875, opened 2023-09-11
feature: Move app context into app module

madsrasmussen:
I have moved the app.context.js into the app module to keep dependencies close to each other. The App Context is now exposed through the app module. It removes the context module at the same time which could lead to confusion between the context module and context-API module

These changes also need to be updated in the razor templates: see PR: https://github.com/umbraco/Umbraco-CMS/pull/14792


iOvergaard (changes requested):
Looks good! But I think you need to update the exports of `package.json` as well if it is publically accessible. Otherwise people won't get typings for the module.

madsrasmussen:
@iOvergaard so close. It is updated now.

iOvergaard (approved):

PR #877, opened 2023-09-11
Bugfix: Save doctype throws error

madsrasmussen:
This PR cleans up saving a content type. We had a bunch of temporary hacks because of missing functionality on the server. I have now been able to remove those.


nielslyngsoe (approved):
So happy to get this corrected!

PR #878, opened 2023-09-11
Feature: Delete Document Type

madsrasmussen:
This PR implements the Delete document type entity action. 

It adds the correct methods, data sources + missing stores to reuse the generic Delete entity action.


nielslyngsoe (approved):
looks good

PR #879, opened 2023-09-13
Feature: Document Recycle Bin 1. iteration

madsrasmussen:
This PR implements the first version of the Document Recycle Bin. It is now possible to Trash a document and see it has been moved to the recycle bin.

There are still a couple of improvements we need to make. 
* The stores aren't currently up to date, which means that sometimes a refresh is needed to see the updated recycle bin because of a cached tree result. The document will be removed from the document tree though.
* It is not possible to completely delete a document or empty the recycle bin. We are missing endpoints to do that.
* ... and some more :)

We need to look into some connected store system, so we don't have to manually do that through a repository. That will have to be in another PR.

Screenshot 2023-09-15 at 10 35 10


iOvergaard (approved):
LGTM

PR #880, opened 2023-09-13
feat: service configuration urls

iOvergaard:
Due to the backend switching to membership authorization by default, we no longer have the `.well-known` endpoint, so we need to code the paths statically.

This PR should not be merged until the backend is ready. @kjac will know when the right time to do this is.


kjac:
@iOvergaard backend is *finally* ready for this now 😄 merge at will.

kjac (approved):

PR #890, opened 2023-09-20
Feature/dictionary update

loivsen:
Dictionary
- Localizations
- Route URL for creating a new dictionary item `create/parentId`. The id may be `null` if no parent.
- Mock data were lacking a put API call and showed error. Added put for mocking.


src/mocks/handlers/dictionary.handlers.ts
```diff
@@ -115,6 +106,18 @@ export const handlers = [
 		return res(ctx.status(200), ctx.json(saved));
 	}),
 
+	rest.put('/umbraco/management/api/v1/dictionary/:id', async (req, res, ctx) => {
```
iOvergaard, line 109: Please use the function `umbracoPath` for any mock URLs. Check the other handlers to see how it is used.

src/packages/dictionary/dashboards/dictionary/dashboard-localization-dictionary.element.ts
```diff
@@ -124,40 +112,25 @@ export class UmbDashboardTranslationDictionaryElement extends UmbLitElement {
 
 	#filter(e: { target: HTMLInputElement }) {
 		this._tableItemsFiltered = e.target.value
-			? this.#tableItems.filter((t) => t.id.includes(e.target.value))
+			? this.#tableItems.filter((t) => t.id.toLocaleLowerCase().includes(e.target.value.toLocaleLowerCase()))
```
iOvergaard, line 115: This is only an optimization, but consider running any functions outside of the filter method first so you don't have to run them inside a loop for each iteration. Doing that will optimize the execution moving from O(n) to O(1).

In this case, you could store the value of `e.target.value.toLocaleLowerCase()` outside the filter call in a local variable and then refer to that when comparing with `includes()`.

src/packages/dictionary/dictionary/workspace/dictionary-workspace-editor.element.ts
```diff
@@ -38,10 +38,16 @@ export class UmbDictionaryWorkspaceEditorElement extends UmbLitElement {
 		return html`
 			<umb-workspace-editor alias="Umb.Workspace.Dictionary">
 				<div id="header" slot="header">
-					<uui-button href="/section/dictionary/dashboard" label="Back to list" compact>
+					<uui-button
+						href="/section/dictionary/dashboard"
```
iOvergaard, line 42: This url seems to be absolute. I'm not sure it would work inside the CMS, where we depend on links being relative, so they pick up the base URL. Try it out and consider changing it to `section/dictionary/dashboard`.

src/packages/dictionary/dictionary/workspace/views/editor/workspace-view-dictionary-editor.element.ts
```diff
@@ -62,12 +62,11 @@ export class UmbWorkspaceViewDictionaryEditorElement extends UmbLitElement {
 	render() {
 		return html`
 			<uui-box>
-				<p>Edit the different language versions for the dictionary item '<em>${this._dictionary?.name}</em>' below.</p>
-
+				${unsafeHTML(this.localize.term('dictionaryItem_description', this._dictionary?.name || 'unnamed'))}
```
iOvergaard, line 65: Do we have a translation for the word "unnamed" or how did this work in the current backoffice?

iOvergaard (changes requested):
Workspace part looks good, but a few optimizations here and there. Please also make sure it is tested against the real CMS.

src/packages/dictionary/dictionary/workspace/views/editor/workspace-view-dictionary-editor.element.ts
loivsen: I was unable to find any fitting candidate for this.

In the current backoffice, you have to enter a name before you can go to this page. It looks like the name is then "locked in" the translation until you press save.

iOvergaard (approved):

PR #894, opened 2023-09-21
fix stacking of popovers

maudsj:
Dropdowns, color picker etc displaying under the following property editors, resulting in e.g. elements of a dropdown not being selectable anymore

Remove container type styling on host in workspace property layout and remove z-index on headerColumn (labels)

- [x ] Bug fix (non-breaking change which fixes an issue)

## How to test?
Add an extra language in settings, use color picker in content

image

## Checklist

- [x ] I have read the **[CONTRIBUTING](<(https://github.com/umbraco/Umbraco.CMS.Backoffice/blob/main/.github/CONTRIBUTING.md)>)** document.


nielslyngsoe:
Hi @maudsj

Seems to work, but this change breaks the container query — Maybe this is our only option, with the current state of Container Queries. But I would really like us to investigate if there could be another option

How do you see our options?

If nothing else can work for us, we should get rid of the Container Query and then implement another way for this to break into the look with the layout with the label on top and the editor below.

PR #898, opened 2023-09-22
localize document-info

filipbech:
I added type to the <umb-localize element... it should probably be individual elements instead, since the input "key" is weird for the input for dates, numbers etc... Could we reworked later!

### DIFF
--- a/src/packages/core/localization/index.ts
+++ b/src/packages/core/localization/index.ts
@@ -1,3 +1,6 @@
 import './localize.element.js';
+import './localize-date.element.js';
+import './localize-number.element.js';
+import './localize-relative-time.element.js';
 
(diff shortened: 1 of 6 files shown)


bjarnef:
I wonder if it would be better to add some utility functions to format date, number, etc. to a specific culture?

E.g. the [Shoelace](https://shoelace.style/) has some great components:
https://shoelace.style/components/format-date
https://shoelace.style/components/format-number
https://shoelace.style/components/relative-time

I think it may confuse with localization of translation keys?

Maybe this would be better:
`<umb-format-date>`, `<umb-format-number>`, `<umb-relative-time>`

iOvergaard (approved):

PR #900, opened 2023-09-28
property editor dropdown

loivsen:
Property editor dropdown


src/packages/core/components/input-dropdown/input-dropdown-list.element.ts
```diff
@@ -0,0 +1,54 @@
...
+	}
+
+	#onChange(e: UUISelectEvent) {
+		e.stopPropagation();
+		if (e.target.value) this.value = e.target.value;
+		this.dispatchEvent(new CustomEvent('change', { bubbles: true, composed: true }));
+	}
+
+	render() {
+		if (!this.options) return nothing;
```
iOvergaard, line 31: Would it perhaps be better to show an empty select if there are no options, so developers can see this is happening?

src/packages/core/property-editor/uis/dropdown/property-editor-ui-dropdown.element.ts
```diff
@@ -1,22 +1,65 @@
-import { html, customElement, property } from '@umbraco-cms/backoffice/external/lit';
-import { UmbTextStyles } from "@umbraco-cms/backoffice/style";
+import { UUISelectEvent } from '@umbraco-ui/uui';
+import type { UmbPropertyEditorConfigCollection } from '@umbraco-cms/backoffice/property-editor';
+import { html, customElement, property, state } from '@umbraco-cms/backoffice/external/lit';
+import { UmbTextStyles } from '@umbraco-cms/backoffice/style';
 import { UmbPropertyEditorExtensionElement } from '@umbraco-cms/backoffice/extension-registry';
+import '../../../components/input-dropdown/input-dropdown-list.element.js';
```
iOvergaard, line 6: This import path looks off somehow. Do we have an alias for this perhaps?

src/packages/core/components/input-dropdown/input-dropdown-list.element.ts
loivsen: I think thats a good point. I changed it.

src/packages/core/property-editor/uis/dropdown/property-editor-ui-dropdown.element.ts
loivsen: Eh, it's not even needed I just forgot to export the input element.

iOvergaard (approved):

PR #901, opened 2023-09-28
Feature: Modal Events + Observable Value

madsrasmussen:
## Modal events and Observable value

This PR introduces an "official" way of communicating from a modal to any consumers of the context.

In some scenarios, we need to respond to different kinds of events happening within a modal. Until now it has been possible to dispatch events through the modal element within the context. However, it was cumbersome for any consumers to listen for the events because they had to subscribe to the element through the context before it was possible to add listeners.

As part of this PR, I have made the modal context itself an EventTarget, which means we can dispatch and listen to events directly on the modal context. This means that the API becomes easier to use.

**Dispatching events from within a modal**

```ts
export class MyModalElement extends UmbModalBaseElement {
  #onModalAction() {
    this.modalContext?.dispatchEvent(new CustomEvent('action-happened'));
  }
}
```

**Listening for events outside the modal**
```ts
const modalContext = this.#modalManagerContext.open(MY_MODAL);

modalContext.addEventListener('action-happened', (event) => {
  console.log('react on modal action happening');
});

modalContext?.onSubmit().then((value) => {
  console.log('Modal submitted with value: ', value);
});
```

As a bonus, I have also added a modal "value" as part of the modal context. This gives us the possibility to look up the current value when an event happens. Currently, the value is first assembled when the modal is submitted. We will have to update modals going forward to keep the value in the context up to date. Another use case for the modal context value is that the implementor of the modal can subscribe to the value and can thereby create a "real-time"-update experience when values are updated in the dialog. This can be used in any case where we want a preview to be updated as a dialog is filled out.

With this update, we get a closer alignment to how property editor UIs are built too.

**Subscribing to value updates**
```ts
const modalContext = this.#modalManagerContext.open(MY_MODAL);

modalContext.value.subscribe((value) => {
  console.log('real time value updates: ', value);
});
```

**Known issues:**

We currently miss the `close-end` modal event in UUI Library which is fired when a model closing animation is done. I have used the `close` event instead which results in modals closing without the animation. It will be an easy fix, but the issue can first be fixed when the next version of UUI is released.


nielslyngsoe:
Just skimmed the code, especially the implementation changes and how it affected the modal route registrations. But the bindings between the two are near to intact and the changes look good. So well done! I'm looking forward to this optimization/alignment so we get better reactive options with Modals.

nielslyngsoe (approved):
Good, I have now tested and it works as expected

PR #904, opened 2023-09-29
Feature/sort document type tabs

loivsen:
Sorting for document-type tabs.

Features
- The Document Type Workspace Context now has an observable `isSorting` that returns a boolean.
- `isSorting = true` when clicking the Reorder button.
- Containers (tabs and groups) and properties can now be sorted.
- When dragging a container or property, a placeholder is shown in the current slot the sorter controller looks at.

Missing features
- Properties cannot be dragged into another group yet
- Groups cannot be converted to a tab yet
- Should we only be able to drag at the handle icon?
- Filling out the input field of sortOrder does not do anything yet.

Known errors
- Calculation of sortOrder when dragging into a new slot is not always correct
- Sort controller will not drag an item on the first attempt (general issue for SorterController).


iOvergaard:
@loivsen Your code looks good but there is a build error. Not sure if it is related to the changes you made. Would you make sure, please?

loivsen:
> @loivsen Your code looks good but there is a build error. Not sure if it is related to the changes you made. Would you make sure, please?

@iOvergaard I fixed the issue.

iOvergaard (approved):

PR #914, opened 2023-10-02
DX Improvement: Correct users import map to user

nielslyngsoe:
Note this changes the import map and folder of `/users/` to `/user/`

This also renames the UmbPropertyEditorExtensionElement to UmbPropertyEditorUiElement

Plus removing a test `headerApp`

## Types of changes

## Checklist


package.json
```diff
@@ -65,9 +65,9 @@
 		"./partial-view": "./dist-cms/packages/templating/partial-views/index.js",
 		"./stylesheet": "./dist-cms/packages/templating/stylesheets/index.js",
 		"./template": "./dist-cms/packages/templating/templates/index.js",
-		"./user-group": "./dist-cms/packages/users/user-groups/index.js",
-		"./current-user": "./dist-cms/packages/users/current-user/index.js",
-		"./users": "./dist-cms/packages/users/users/index.js",
+		"./user-group": "./dist-cms/packages/user/user-groups/index.js",
+		"./current-user": "./dist-cms/packages/user/current-user/index.js",
+		"./users": "./dist-cms/packages/user/user/index.js",
```
madsrasmussen, line 70:
```suggestion
		"./user": "./dist-cms/packages/user/user/index.js",
```

madsrasmussen (commented):
I think we should consider a new name for our root package folders, so they make it more clear that they can consist of different areas and modules. So instead of having `user/user` we should come up with something different for the first folder. 

All I can come up with at the moment is "user-management" but it is long and not that clever. I also thought about shortening management to "user-mgmt".

The same should go for the document, media, member, etc. packages that have the same issue with a subfolder with the same name.

We also need to make sure that we get the importmaps in the razor templates updated when we change paths to modules.

nielslyngsoe:
@madsrasmussen agree, but I'm neither sure about the naming.

I have fixed the merge conflict and merged your suggestion. So I hope you will have another look :-)

madsrasmussen (approved):
I will merge this now. There is one weird test complaining about imports from the backend-api module. The same error happens on the main branch, so it has not been introduced as part of these changes and we can investigate in another PR.

PR #919, opened 2023-10-05
Enable api js and element js for the same extension

nielslyngsoe:
Work in progress


src/packages/core/event/executed.event.ts
```diff
@@ -1,5 +1,5 @@
 export class UmbExecutedEvent extends Event {
```
madsrasmussen, line 1: Should we rename the class too?

src/packages/documents/documents/entity-bulk-actions/manifests.ts
```diff
@@ -11,28 +11,30 @@ const entityActions: Array<ManifestEntityBulkAction> = [
...
 		meta: {
 			label: 'Copy',
 			repositoryAlias: DOCUMENT_REPOSITORY_ALIAS,
-			api: UmbDocumentCopyEntityBulkAction,
-		},
-		conditions: {
-			entityType,
 		},
+		conditions: [{
+			alias: 'Umb.Condition.CollectionEntityType',
```
madsrasmussen, line 35: Is it "dangerous" to call this condition "collectionEntityType", if the bulk actions can be used in other places than collections in the future?

The entity actions are so generic, that they in theory can be used anywhere.

madsrasmussen (commented):
In general, everything looks good to me. I have left two smaller comments. 

Do we have some documentation we need to update with these changes?

Update:
We will update docs for the next preview release, so we don't have documentation now that doesn't work for the current preview release.

src/packages/core/event/executed.event.ts
madsrasmussen: This has been updated

PR #920, opened 2023-10-06
Feature/markdown editor

loivsen:
Simple markdown editor using Monaco
TODO: Actions, Preview.


iOvergaard (approved):
LGTM and great that you figured out how to call `loadCodeEditor()`. I would suggest we figure out how to pass the options to `<umb-code-editor>` as a property object directly in the render() function next, so you don't have to wait for the actual editor to load.

PR #925, opened 2023-10-11
Improvement: User spring cleaning - Part 1

madsrasmussen:
First part of the User Clean up task


src/packages/core/collection/collection.context.ts
```diff
@@ -30,107 +29,119 @@ export class UmbCollectionContext<ItemType, FilterModelType extends UmbCollectio
...
-			}
-		);
+		this.entityType = entityType;
+		this.host = host;
+
+		const currentUrl = new URL(window.location.href);
+		this.collectionRootPathname = currentUrl.pathname.substring(0, currentUrl.pathname.lastIndexOf('/'));
+
+		this.init = Promise.all([
+			new UmbObserverController(
```
nielslyngsoe, line 49: I know we still dont have the Context Base, but, ideally a Context is a Controller.
So make the CollectionContext extends the Controller base like this: `extends UmbBaseController`

This will give you the ability to do`this.observe()`

But even more importantly, it will make the Context a ControllerHost, meaning it owns the references to its sub controllers. meaning cleanup/destroying will be much more sharp. Enabling taking the CollectionContext out of service, without destroying the origin element. That might not make any difference in this case, but I want us to aim for that architecture :-)

src/packages/core/collection/collection.context.ts
```diff
@@ -30,107 +29,119 @@ export class UmbCollectionContext<ItemType, FilterModelType extends UmbCollectio
...
+				async (repositoryManifest) => {
+					if (repositoryManifest) {
+						const result = await createExtensionApi(repositoryManifest, [this.host]);
+						this.repository = result as UmbCollectionRepository;
+						this.requestCollection();
+					}
+				},
+			).asPromise(),
+
+			new UmbObserverController(
```
nielslyngsoe, line 61: Then this can become this.observe as well :-)

src/packages/core/collection/collection.context.ts
```diff
@@ -30,107 +29,119 @@ export class UmbCollectionContext<ItemType, FilterModelType extends UmbCollectio
...
 
-	public setSelection(value: Array<string>) {
-		if (!value) return;
-		this.#selection.next(value);
+	/**
+	 * Sets the current selection.
+	 * @param {Array<string>} selection
+	 * @memberof UmbCollectionContext
+	 */
+	public setSelection(selection: Array<string>) {
```
nielslyngsoe, line 91: We do have a `UmbSelectionManagerBase`, could it make sense to use that as part of this context. So we instantiate that on a prop called `selection` ?  Just an idea, if you think it could make sense?

src/packages/core/collection/collection.context.ts
```diff
@@ -143,11 +154,48 @@ export class UmbCollectionContext<ItemType, FilterModelType extends UmbCollectio
...
+
+	#setCurrentView() {
+		const currentUrl = new URL(window.location.href);
+		const lastPathSegment = currentUrl.pathname.split('/').pop();
+		const views = this.#views.getValue();
+		const viewMatch = views.find((view) => view.meta.pathName === lastPathSegment);
+
+		/* TODO: Find a way to figure out which layout it starts with and set _currentLayout to that instead of [0]. eg. '/table'
+			For document, media and members this will come as part of a data type configuration, but in other cases "users" we should find another way. 
+			This should only happen if the current layout is not set in the URL.
```
nielslyngsoe, line 194: Maybe we could hardcode the one we look for? Just in the case for Users... so that looks for the grid view? Base on something that we find relatively stable.. then we might later make that configurable or smarter in some way?

src/packages/core/collection/collection.element.ts
```diff
@@ -1,68 +1,41 @@
...
-import './collection-selection-actions.element.js';
-import './collection-toolbar.element.js';
-
 @customElement('umb-collection')
 export class UmbCollectionElement extends UmbLitElement {
 	@state()
 	private _routes: Array<UmbRoute> = [];
 
-	@state()
-	private _selection?: Array<string> | null;
```
nielslyngsoe, line 20: really good to get this away from the element, perfect 👍

src/packages/user/user/collection/user-collection.element.ts
```diff
@@ -1,59 +1,24 @@
...
-			redirectTo: 'grid',
-		},
-	];
-
-	connectedCallback(): void {
-		super.connectedCallback();
-		this.provideContext(UMB_COLLECTION_CONTEXT, this.#collectionContext);
+	constructor() {
+		super();
+		this.provideContext(UMB_COLLECTION_CONTEXT, this.collectionContext);
```
nielslyngsoe, line 14: As well as making the CollectionContext a Controller then it should also become self providing :-) so removing this line.

nielslyngsoe: We also need to figure out if it makes sense to have a generic element for Collections, cause we have "workspace-view-collection.element" but maybe it not good to have one generic.. and BTW. If someone needs their own custom, then it might be user collection.. so I don't know, just a perspective.

src/packages/core/modal/token/change-password-modal.token.ts
```diff
@@ -1,9 +1,19 @@
...
+	requireOldPassword?: boolean;
 }
 
-export const UMB_CHANGE_PASSWORD_MODAL = new UmbModalToken<UmbChangePasswordModalData>('Umb.Modal.ChangePassword', {
-	type: 'dialog',
-});
+export interface UmbChangePasswordModalValue {
+	newPassword: string;
+	confirmPassword: string;
+	oldPassword?: string;
```
nielslyngsoe, line 11: Could this just be one password?  on open its the current, on submit it's the new.
Then its the dialogs responsibility to handle the `confirmPassword`? So that's just UI of the modal, that the outsides does not know about?

src/packages/user/user/collection/user-collection.element.ts
nielslyngsoe: Oh, I see you did that with the umb-collection element, right?

src/packages/user/user/components/user-document-start-node/user-document-start-node.element.ts
```diff
@@ -0,0 +1,53 @@
...
+	#itemRepository = new UmbDocumentRepository(this);
+
+	protected async firstUpdated(): Promise<void> {
+		if (this.ids.length === 0) return;
+		const { data } = await this.#itemRepository.requestItems(this.ids);
+		this._displayValue = data || [];
+	}
+
+	render() {
+		if (this.ids.length < 1)
```
nielslyngsoe, line 24: I do prefere always having { }, it makes it more clear what the intentions are :-)

nielslyngsoe: and in this case I would then prefer an else statement for the next part as that would make it easier to read that its one or the other :-)

src/packages/user/user/components/user-media-start-node/user-media-start-node.element.ts
```diff
@@ -0,0 +1,53 @@
...
+	#itemRepository = new UmbMediaRepository(this);
+
+	protected async firstUpdated(): Promise<void> {
+		if (this.ids.length === 0) return;
+		const { data } = await this.#itemRepository.requestItems(this.ids);
+		this._displayValue = data || [];
+	}
+
+	render() {
+		if (this.ids.length < 1)
```
nielslyngsoe, line 24: same with {} and else statement

src/packages/user/user/conditions/user-allow-delete-action.condition.ts
```diff
@@ -0,0 +1,29 @@
...
+} from '@umbraco-cms/backoffice/extension-api';
+
+export class UmbUserAllowDeleteActionCondition extends UmbUserActionConditionBase {
+	constructor(args: UmbConditionControllerArguments<UmbConditionConfigBase>) {
+		super(args);
+	}
+
+	async onUserDataChange() {
+		// don't allow the current user to delete themselves
+		if (!this.userData || !this.userData.id || (await this.isCurrentUser())) {
```
nielslyngsoe, line 15: Cool :-)

src/packages/user/user/modals/resend-invite/resend-invite-to-user-modal.element.ts
```diff
@@ -0,0 +1,70 @@
...
+		if (!isValid) return;
+
+		const formData = new FormData(form);
+		const message = formData.get('message') as string;
+
+		const { error } = await this.#userRepository.resendInvite({
+			message,
+		});
+
+		debugger;
```
nielslyngsoe, line 29: Remember to remove this one ;-)

src/packages/user/user/repository/user-repository-base.ts
```diff
@@ -0,0 +1,32 @@
+import { UMB_USER_STORE_CONTEXT_TOKEN, UmbUserStore } from './user.store.js';
+import { UMB_USER_ITEM_STORE_CONTEXT_TOKEN, UmbUserItemStore } from './item/user-item.store.js';
+import type { UmbControllerHostElement } from '@umbraco-cms/backoffice/controller-api';
+import { UmbContextConsumerController } from '@umbraco-cms/backoffice/context-api';
+import { UMB_NOTIFICATION_CONTEXT_TOKEN, UmbNotificationContext } from '@umbraco-cms/backoffice/notification';
+
+export class UmbUserRepositoryBase {
```
nielslyngsoe, line 7: I'm not sure if we have made Repositories a Controller anywhere jet. But I would say that would be the right for this one as well.. So `new UmbContextConsumerController(this.host, ` can be changed into `this.consumeContext(...`

src/packages/user/user/workspace/components/user-workspace-profile-settings/user-workspace-profile-settings.element.ts
```diff
@@ -0,0 +1,138 @@
...
+			} else {
+				// If users language code did not fit any of the options. We will create an option that fits, named unknown.
+				// In this way the user can keep their choice though a given language was not present at this time.
+				this.languages.push({
+					value: currentUserLanguageCode ?? 'en-us',
+					name: currentUserLanguageCode ? `${currentUserLanguageCode} (unknown)` : 'Unknown',
+					selected: true,
+				});
+			}
+		});
```
nielslyngsoe, line 83: I know its annoying. But I havent found a better way jet. We need to give a unique string or Symbol to these observers so they replace each other if setup twice...  like this: `this.observe(..., ..., 'mySomethingObserver')`

src/shared/auth/auth.context.ts
```diff
@@ -68,4 +68,9 @@ export class UmbAuthContext implements IUmbAuth {
 	signOut(): Promise<void> {
 		return this.#authFlow.signOut();
 	}
+
+	async isUserCurrentUser(userId: string): Promise<boolean> {
+		const currentUser = await firstValueFrom(this.currentUser);
```
nielslyngsoe, line 73: Is this different than `this.#currentUser.getValue()`    I see its async, but wouldn't this just give you undefined. Or would it wait for the first real value?

nielslyngsoe (changes requested):
Really good clean up, I have made a new notes and thoughts

src/packages/core/collection/components/collection-view-bundle.element.ts
```diff
@@ -0,0 +1,95 @@
...
+			this.#collectionContext = context;
+			if (!this.#collectionContext) return;
+			this._collectionRootPathname = this.#collectionContext.collectionRootPathname;
+			this.#observeViews();
+			this.#observeCurrentView();
+		});
+	}
+
+	#observeCurrentView() {
+		this.observe(this.#collectionContext!.currentView, (view) => {
```
nielslyngsoe, line 36: note about missing ControllerAlias, to make observe replace it self if this method is called again.

src/packages/core/collection/components/collection-view-bundle.element.ts
```diff
@@ -0,0 +1,95 @@
...
+			this._currentView = view;
+		});
+	}
+
+	#observeViews() {
+		this.observe(this.#collectionContext!.views, (views) => {
```
nielslyngsoe, line 42: note about missing ControllerAlias, to make observe replace it self if this method is called again.

src/packages/user/user/collection/views/grid/user-grid-collection-view.element.ts
```diff
@@ -2,37 +2,51 @@ import { getDisplayStateFromUserStatus } from '../../../../utils.js';
...
+	#userGroupCollectionRepository = new UmbUserGroupCollectionRepository(this);
 
 	constructor() {
 		super();
 
-		//TODO: Get user group names
-
 		this.consumeContext(UMB_COLLECTION_CONTEXT, (instance) => {
 			this.#collectionContext = instance as UmbUserCollectionContext;
 			this.observe(this.#collectionContext.selection, (selection) => (this._selection = selection));
```
nielslyngsoe, line 31: note about missing ControllerAlias, to make observe replace it self if this method is called again.

src/packages/user/user/conditions/user-allow-action-base.condition.ts
```diff
@@ -0,0 +1,50 @@
...
+	protected userData?: UmbUserDetail;
+
+	constructor(args: UmbConditionControllerArguments<UmbConditionConfigBase>) {
+		super(args.host);
+		this.config = args.config;
+		this.#onChange = args.onChange;
+
+		this.consumeContext(UMB_WORKSPACE_CONTEXT, (context) => {
+			const userContext = context as UmbUserWorkspaceContext;
+			this.observe(userContext.data, (data) => {
```
nielslyngsoe, line 25: note about missing ControllerAlias, to make observe replace it self if this method is called again.

src/packages/user/user/workspace/components/user-workspace-access-settings/user-workspace-access-settings.element.ts
```diff
@@ -0,0 +1,122 @@
...
+	private _user?: UserResponseModel;
+
+	#userWorkspaceContext?: typeof UMB_USER_WORKSPACE_CONTEXT.TYPE;
+
+	constructor() {
+		super();
+
+		this.consumeContext(UMB_USER_WORKSPACE_CONTEXT, (instance) => {
+			this.#userWorkspaceContext = instance;
+			this.observe(this.#userWorkspaceContext.data, (user) => (this._user = user));
```
nielslyngsoe, line 23: note about missing ControllerAlias, to make observe replace it self if this method is called again.

src/packages/user/user/workspace/components/user-workspace-info/user-workspace-info.element.ts
```diff
@@ -0,0 +1,138 @@
...
+	private _userInfo: Array<UmbUserWorkspaceInfoItem> = [];
+
+	#userWorkspaceContext?: typeof UMB_USER_WORKSPACE_CONTEXT.TYPE;
+
+	constructor() {
+		super();
+
+		this.consumeContext(UMB_USER_WORKSPACE_CONTEXT, (instance) => {
+			this.#userWorkspaceContext = instance;
+			this.observe(this.#userWorkspaceContext.data, (user) => {
```
nielslyngsoe, line 25: note about missing ControllerAlias, to make observe replace it self if this method is called again.

src/packages/user/user/workspace/components/user-workspace-profile-settings/user-workspace-profile-settings.element.ts
```diff
@@ -0,0 +1,138 @@
...
+		super();
+
+		this.consumeContext(UMB_AUTH, (instance) => {
+			this.#authContext = instance;
+			this.#observeCurrentUser();
+		});
+
+		this.consumeContext(UMB_USER_WORKSPACE_CONTEXT, (instance) => {
+			this.#userWorkspaceContext = instance;
+			this.observe(this.#userWorkspaceContext.data, (user) => (this._user = user));
```
nielslyngsoe, line 35: note about missing ControllerAlias, to make observe replace it self if this method is called again.

src/packages/user/user/workspace/components/user-workspace-profile-settings/user-workspace-profile-settings.element.ts
```diff
@@ -0,0 +1,138 @@
...
+		const target = event.composedPath()[0] as UUISelectElement;
+
+		if (typeof target?.value === 'string') {
+			this.#userWorkspaceContext?.updateProperty('languageIsoCode', target.value);
+		}
+	}
+
+	#observeCurrentUser() {
+		if (!this.#authContext) return;
+		this.observe(this.#authContext.currentUser, async (currentUser) => {
```
nielslyngsoe, line 49: note about missing ControllerAlias, to make observe replace it self if this method is called again.

src/packages/user/user/workspace/user-workspace.context.ts
```diff
@@ -26,19 +26,31 @@ export class UmbUserWorkspaceContext
...
 
 	async load(id: string) {
-		const { data } = await this.repository.requestById(id);
+		const { data, asObservable } = await this.repository.requestById(id);
 		if (data) {
 			this.setIsNew(false);
 			this.#data.update(data);
 		}
+
+		this.observe(asObservable(), (user) => this.onUserStoreChanges(user));
```
nielslyngsoe, line 35: note about missing ControllerAlias, to make observe replace it self if this method is called again.

src/packages/core/collection/collection.context.ts
madsrasmussen: Great! This is now updated

src/packages/user/user/modals/resend-invite/resend-invite-to-user-modal.element.ts
madsrasmussen: It is gone now :D

src/shared/auth/auth.context.ts
madsrasmussen: It waits for the first real value. A small shortcut so you don't have to set up an observer etc.

src/packages/core/collection/collection.context.ts
madsrasmussen: I have updated the code so all logic now comes from the SelectionManager. I can't figure out what the best practice is. Should we talk directly with the selection manager or implement methods in the class that talk to the selection manager? I have currently gone with the last solution because then each implementation can pick and choose the features. But it also adds a lot more code. 

Any good ideas?

src/packages/user/user/repository/user-repository-base.ts
madsrasmussen: I have made a RepositoryBase now that extends the BaseController. The code for all user repositories has been updated. Now we just have to go through all the repos. I think we will have to do that in the next PR.

src/packages/core/modal/token/change-password-modal.token.ts
madsrasmussen: Ah, yes... We will have to send the old password back too.  It is needed when we change the password for the current user.

src/packages/user/user/collection/user-collection.element.ts
madsrasmussen: There is one generic `UmbCollectionElement` and the user collections extend that. I then overwrite the renderToolbar method to add the custom filters. I am trying to move as much back into the generic element that makes sense. 

I think we should have an extension something called a "CollectionAction". In most cases, this will be the create action, but for users, it will be both create and invite.

I will also add logic for pagination to the generic context/element.

madsrasmussen: I have made the collection context self providing. Thanks 👍

src/packages/core/collection/collection.context.ts
madsrasmussen: I have added a task for this, so we can come back to it.

nielslyngsoe (approved):
Yes, looks good. And good to see that the SelectionManager could replace the custom code :-)

PR #926, opened 2023-10-11
Feature/markdown editor preview

loivsen:
## Description
Markdown editor preview

- Provides a preview of the markdown editor.
- Using Marked for turning markdown into HTML
- Using sanitize-html to ensure nothing dangerous is inserted
- Added missing property config overlaySize.

## Types of changes


package.json
iOvergaard, line 1: You need to add the new exports at the top of this file in the "exports" section as well.

iOvergaard: All externals are automatically exported by the `*`. Disregard this message ☺️

PR #932, opened 2023-10-17
Fix missing toolbar icons shown in configuration of richtext editor

bjarnef:
## Description

Added missing icons shown at toolbar in configuration of richtext editor.

image

The `tinymce` doesn't contain a "default" property, but `IconManager` directly. Not sure why, but when when output to console in render function, it seems to log twice.

**Before**

image

**After**

image

## Types of changes

## Checklist


bjarnef:
@iOvergaard @loivsen seeems like an oversight?

nielslyngsoe:
@bjarnef looks good to me, there is though a TypeScript issue. I assume TinyMCE is not TypeScript, if that's true then I recommend casting you import to `any`. otherwise, try to see what you can do to get the right type.

bjarnef:
@nielslyngsoe ahh, I see the failed test.

I noticed elsewhere in code where it use `type` instead.
https://github.com/umbraco/Umbraco.CMS.Backoffice/blob/62c3fc89aeec41b6b163baaaee63f0721698bb90/src/packages/core/components/input-tiny-mce/input-tiny-mce.handlers.ts#L1

nielslyngsoe (approved):
Cool, thanks!

src/packages/core/property-editor/uis/tiny-mce/config/toolbar/property-editor-ui-tiny-mce-toolbar-configuration.element.ts
```diff
@@ -6,7 +6,7 @@ import { firstValueFrom } from '@umbraco-cms/backoffice/external/rxjs';
 import { UmbPropertyEditorConfigCollection } from '@umbraco-cms/backoffice/property-editor';
 import { tinymce } from '@umbraco-cms/backoffice/external/tinymce';
 
-const tinyIconSet = tinymce.default?.IconManager.get('default');
+const tinyIconSet = tinymce.IconManager.get('default');
```
nielslyngsoe, line 9: I see, then the solution is to do this (basically saying "TypeScript I know better than you...")
```suggestion
const tinyIconSet = (tinymce as any).IconManager.get('default');
```

bjarnef:
@nielslyngsoe is there a reason elsewhere in code it use

https://github.com/umbraco/Umbraco.CMS.Backoffice/blob/527936fcbdbe061fbb17cb81e0d194ddf81565ff/src/packages/core/components/input-tiny-mce/tiny-mce-plugin.ts#L2

nielslyngsoe:
@bjarnef when we apply type to the import statement it basically means that we do not want a real import, we just want to borrow the type in the TypeScript compile check. :-) So if the code is actually used, then we should not append type to the import statement :-)

iOvergaard (approved):
LGTM

PR #936, opened 2023-10-19
Multinode Treepicker

loivsen:
## Description

Property editor for the MultiNode Treepicker.
As well as StartNode picker (used in the configuration)

**Note:** Needs update when Media picker and  Member picker are ready (multinode picker) and MediaTypes picker is ready (for startnode picker)

## Types of changes


loivsen:
Now that dynamics path coming in to replace XPath, we maybe need something in the picker when the type is set to "content". Originally it shows an XPath option.

src/mocks/data/user-group.data.ts
```diff
@@ -50,6 +50,13 @@ export const data: Array<UserGroupResponseModel> = [
 		documentStartNodeId: 'all-property-editors-document-id',
 		permissions: [UMB_USER_PERMISSION_DOCUMENT_CREATE, UMB_USER_PERMISSION_DOCUMENT_DELETE],
 	},
+	{
```
madsrasmussen, line 53: Is this user group used for anything? maybe it is just leftover test data?

src/packages/core/components/input-start-node/input-start-node.element.ts
```diff
@@ -0,0 +1,94 @@
+import { UmbInputDocumentElement } from '@umbraco-cms/backoffice/document';
```
madsrasmussen, line 1: It looks like all this input does is select between different content types. Maybe we could rename this to input-content-type instead? Then it could also be used for other things that needs to select between content types.

I also think this component should go into the content-type folder inside the core folder. You can see an example of how we have components in the culture folder.

src/packages/core/components/input-start-node/input-start-node.element.ts
```diff
@@ -0,0 +1,94 @@
...
+
+	@property({ type: Array })
+	options: Array<Option> = [];
+
+	#onTypeChange(event: UUISelectEvent) {
+		this.startNodeType = event.target.value as StartNode['type'];
+
+		// Clear others
+		this.startNodeId = '';
+		this.dispatchEvent(new CustomEvent('change'));
```
madsrasmussen, line 28: You can use the UmbChangeEvent here instead of a custom event.

It will be something like: `this.dispatchEvent(new UmbChangeEvent());` The event class is imported from `@umbraco-cms/backoffice/event`

src/packages/core/components/input-start-node/input-start-node.element.ts
```diff
@@ -0,0 +1,94 @@
...
+			case 'media':
+				return this.renderTypeMedia();
+			case 'member':
+				return this.renderTypeMember();
+			default:
+				return 'No type found';
+		}
+	}
+
+	renderTypeContent() {
```
madsrasmussen, line 56: Can we make these private methods?

src/packages/core/components/input-tree/input-tree.element.ts
```diff
@@ -0,0 +1,130 @@
...
+	}
+	public get value(): string {
+		return super.value as string;
+	}
+
+	selectedIds: Array<string> = [];
+
+	#onChange(event: CustomEvent) {
+		this.value = (event.target as UmbInputDocumentElement).selectedIds.join(',');
+		this.dispatchEvent(new CustomEvent('change', { bubbles: true, composed: true }));
```
madsrasmussen, line 70: The same goes for this event. You can use our own `UmbChangeEvent`. Be aware that our event is not composed, so you might have to check if relied on this event being composed.

src/packages/core/components/input-tree/input-tree.element.ts
```diff
@@ -0,0 +1,130 @@
+import { css, html, customElement, property } from '@umbraco-cms/backoffice/external/lit';
```
madsrasmussen, line 1: I think we should put this component into the tree module as we will do with the `content type input`. You will find a tree folder inside the core package.

src/packages/core/components/input-tree/input-tree.element.ts
```diff
@@ -0,0 +1,130 @@
...
+	}
+
+	constructor() {
+		super();
+	}
+
+	render() {
+		switch (this.type) {
+			case 'content':
+				return html`<umb-input-document
```
madsrasmussen, line 80: You can simplify this code a bit because all of the properties are the same. If you programmatically create the correct input element, based on the type, then you can set the properties once on the element afterward. Something along the lines of:

```
let inputElement = null;
if (type === 'content') {
 inputElement = new UmbInputDocument();
} else if...

inputElement.selectedIds = this.selectedIds;

// insert your element into the dom
```

src/packages/core/property-editor/uis/tree-picker/config/start-node/property-editor-ui-tree-picker-start-node.element.ts
```diff
@@ -1,20 +1,54 @@
...
+
+	#onChange(event: CustomEvent) {
+		const target = event.target as UmbInputStartNodeElement;
+
+		this.value = {
+			type: target.startNodeType,
+			id: target.startNodeId,
+		};
+
+		this.dispatchEvent(new CustomEvent('property-value-change'));
```
madsrasmussen, line 44: If you feel like it, you could create a new custom event, like we have for the change event, and use it here instead of the hardcoded value. I think there might be other places where we also hardcode the event type, so it could also be in a new PR too.

src/packages/core/property-editor/uis/tree-picker/manifests.ts
```diff
@@ -15,9 +15,19 @@ const manifest: ManifestPropertyEditorUi = {
 			properties: [
 				{
 					alias: 'startNode',
-					label: 'Start node',
+					label: 'Node type',
 					description: '',
 					propertyEditorUiAlias: 'Umb.PropertyEditorUi.TreePicker.StartNode',
+					config: [
```
madsrasmussen, line 21: If we change the input-start-node to input-content-type then I think we can hardcode these values inside the input element instead. Then we don't have to pass this config value all the way down to the input element. It might simplify the code a little bit.

madsrasmussen (changes requested):

madsrasmussen:
> Now that dynamics path coming in to replace XPath, we maybe need something in the picker when the type is set to "content". Originally it shows an XPath option.

Let's add this as part of a new PR.

src/mocks/data/user-group.data.ts
loivsen: Yea it might be, I cant remember if it was used for anything.

src/packages/core/components/input-start-node/input-start-node.element.ts
loivsen: sounds good

src/packages/core/property-editor/uis/tree-picker/config/start-node/property-editor-ui-tree-picker-start-node.element.ts
loivsen: Sure, i will do it after this PR gets completed, also a PR for dynamic paths.

madsrasmussen (approved):

PR #938, opened 2023-10-20
Partial Media Types update

JesmoDev:
Partial media type update


nielslyngsoe (approved):
Cool, I know its work in progress but lets get this in so it does not gets de-railed.

PR #941, opened 2023-10-24
Feature/decouple resource controller

iOvergaard:
In order for extension developers to use the UmbResourceController and tryExecute* functions, we need to avoid using the `instanceof` operator, since their API errors and cancelable promises will not be an instance of the internal classes in the Backoffice.


nielslyngsoe (approved):
Good 👍

PR #942, opened 2023-10-24
Feature/isolate auth flow

iOvergaard:
Make it more obvious what you get from the `UMB_AUTH` context and how to use it by explaining that it is related to the Management API only.

The main change is that `performWithFreshTokens` has been renamed to `getLatestToken` and added some documentation to highlight that it is only useful for the management API.

We've also removed an internal auth class from the exports to avoid letting outsiders access that directly. All auth work should be interacted with through UmbAuthContext only, which is available as the token `UMB_AUTH`.


nielslyngsoe (approved):
Looks really good 👍

PR #947, opened 2023-10-30
localization for telemetry

loivsen:
Localization for telemetry dashboard.

### DIFF
--- a/src/packages/settings/dashboards/telemetry/dashboard-telemetry.element.ts
+++ b/src/packages/settings/dashboards/telemetry/dashboard-telemetry.element.ts
@@ -46,7 +46,7 @@ export class UmbDashboardTelemetryElement extends UmbLitElement {
 			this,
 			TelemetryResource.postTelemetryLevel({
 				requestBody: { telemetryLevel: this._telemetryFormData },
-			})
+			}),
 		);
 
 		if (error) {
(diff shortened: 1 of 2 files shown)


iOvergaard (approved):

iOvergaard (changes requested):
Seems to be a test error where it can't tell that `umb-localize` is a real text. How can we fix this? Can we configure the tests to accept `umb-localize` as real text, or should we use the direct `${this.localize(...)}` method instead?

image

src/packages/settings/dashboards/telemetry/dashboard-telemetry.element.ts
```diff
@@ -113,26 +105,16 @@ export class UmbDashboardTelemetryElement extends UmbLitElement {
 	render() {
 		return html`
 			<uui-box class="uui-text">
-				<h1 class="uui-h2">Consent for telemetry data</h1>
+				<h1 class="uui-h2"><umb-localize key="analytics_consentForAnalytics"></umb-localize></h1>
```
iOvergaard, line 108: Ahh, I think it could work if you include a fallback text here, so the tester can see it:

```suggestion
				<h1 class="uui-h2"><umb-localize key="analytics_consentForAnalytics">Consent for telemetry data</umb-localize></h1>
```

iOvergaard (approved):

PR #961, opened 2023-11-02
eslint: add rule to detect erroneous external imports

iOvergaard:
This new rule ensures that we won't accidentally make imports from external libraries elsewhere. The only allowed places to do so are in tests, stories, and `src/external/*` modules.

We won't notice these errors in the dev-server, because Vite knows how to handle them, but it WILL become a problem in the compiled output used in the CMS.

BAD:
```js
import { html } from 'lit';
```

GOOD:
```js
import { html } from '@umbraco-cms/backoffice/external/lit';
```


.eslintrc.json
```diff
@@ -45,6 +45,7 @@
 				"local-rules/umb-class-prefix": "error",
 				"local-rules/prefer-static-styles-last": "warn",
 				"local-rules/ensure-relative-import-use-js-extension": "error",
+				"local-rules/ensure-no-external-imports": "error",
```
madsrasmussen, line 48: can we call the rule something like "app no direct node_module imports", so the name doesn't collide with the naming of our "external"-module?

madsrasmussen (changes requested):
Really great to get this rule in place! You found some weird imports here an there. 

I only have one request regarding the rule name. It can be misunderstood because we call our own module "external".

.eslintrc.json
iOvergaard: Great idea! Here is what Github Copilot suggests, what do you think? 😅

> GitHub Copilot
You could consider naming the rule something like enforce-umbraco-external-imports. This name clearly communicates the purpose of the rule (to enforce imports from the @umbraco-cms/backoffice/external module) and should not conflict with the name of your external module.

madsrasmussen: Brilliant 🎉 😄

nielslyngsoe (approved):
Looks good to me :-)

PR #965, opened 2023-11-06
Feature: Use native Object.groupBy

madsrasmussen:
Browsers now ship an `Object.groupBy` method that does the same job as the one we have used from lodash. I have changed our code to use the native version instead and removed the export of the groupBy function from lodash.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/groupBy


bjarnef:
@madsrasmussen I haven't had a deeper look at native `groupBy` yet, but besides `Object.groupBy()` it seems there are `Map.groupBy()` which accepts any data type as a key: https://dmitripavlutin.com/javascript-array-group/

image

nielslyngsoe (approved):

PR #970, opened 2023-11-07
Isolate RxJS

nielslyngsoe:
Isolate RxJS inside our State Objects to avoid confusion or unforeseen manipulation of the data.

## Types of changes

## Checklist


src/shared/auth/auth.context.ts
```diff
@@ -2,27 +2,26 @@ import { IUmbAuth } from './auth.interface.js';
...
-	#host;
 	#authFlow;
 
 	constructor(host: UmbControllerHostElement, serverUrl: string, redirectUrl: string) {
-		this.#host = host;
+		super(host)
 		this.#authFlow = new UmbAuthFlow(serverUrl, redirectUrl);
 
-		this.isLoggedIn.subscribe((isLoggedIn) => {
+		this.observe(this.isLoggedIn, (isLoggedIn) => {
```
iOvergaard, line 24: It seems this part needs to forward the inner observable to the observer controller, or else it fails because observer.ts tries to call `.subscribe()` directly on whatever is passed in here. I don't know if this is the correct approach, but it works exactly in this case:

```suggestion
		this.observe(this.isLoggedIn.asObservable(), (isLoggedIn) => {
```

iOvergaard: Another approach could be to make observer controller understand the `UmbBasicState` object. Or an approach could even be to simply implement the missing `subscribe()` method in UmbBasicState, but that seems a tad too advanced for that class...?

iOvergaard (changes requested):
It works with one small change, but I don't know if my suggestion is too simplistic:

src/shared/auth/auth.context.ts
nielslyngsoe: Currently it's a bit of an edge developer experience improvement, I think for clarification about what it is that you are doing. I preferred the developer calling `asObservable()`. Leading to the developer taking action on wether they need to full thing to be observable or if they can declare an observable on parts of it. Also make sure they keep the source(state object) as a private var and make the observable public.

This also makes the subscribe method more hidden, which is a good thing, as I would like devs to use our Observe method(controller) so we can ensure that it gets destroyed once an element or controller is taken out of action.

Your idea is a solution if it's shown to be something people struggle with, but I think we need to see that first before we start maintaining something like this. Without it, we have a clearer interface to maintain going forward — which can become a strength if we see any trouble with States in the future.

iOvergaard (approved):

PR #979, opened 2023-11-09
Improvement: Split Current User from auth

madsrasmussen:
This PR separates the Auth Context into an `AuthContext` - responsible for user authentication. And a `CurrentUserContext` - responsible for retrieving and interacting with data for the Authenticated User.

The PR also does the first round of cleanup in the App Element, to further split responsibility into smaller areas:

`UmbSeverConnection`: Check the current Run Time Level - Install, Upgrade, etc of the server.
`UmbDebugContextController`: Listen for Context Debug events.

There is still some code in the app.element.ts that should be moved around, but it will be part of another clean-up round.

With these changes, we also fixed a bug where the current user didn't get loaded correctly with mocked data.

**What to test:**
* The current user gets loaded with mocked data
* The current user gets loaded with data from the CMS


src/packages/user/manifests.ts
```diff
@@ -17,4 +17,10 @@ export const manifests = [
 	...currentUserManifests,
 	...userPermissionManifests,
 	...modalManifests,
+	{
+		name: 'User Management Entry Point',
+		alias: 'Umb.EntryPoint.UserManagement',
+		type: 'entryPoint',
```
nielslyngsoe, line 23: Could we use a bundle instead?  Maybe it's just me but I find the Bundle easier to decode(read), as I then know it's not doing any custom code. It's just registering some more manifests.

nielslyngsoe: Actually, Just expends a manifest.js here, like done with the above manifests.

nielslyngsoe (approved):
Restructure looks good, I did make a few comments, But I assume you will look at those before merging :-)

src/packages/user/manifests.ts
madsrasmussen: I have registered it as a globalContext. Then the entry point is gone :)

src/packages/user/current-user/current-user.context.ts
```diff
@@ -0,0 +1,72 @@
...
+	constructor(host: UmbControllerHost) {
+		super(host);
+
+		this.consumeContext(UMB_AUTH_CONTEXT, (instance) => {
+			this.#authContext = instance;
+			this.#observeIsAuthorized();
+		});
+
+		// TODO: revisit this. It can probably be simplified
+		this.observe(umbLocalizationRegistry.isDefaultLoaded, (isDefaultLoaded) => {
```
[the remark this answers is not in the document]
madsrasmussen: I will leave this as is. Then we can revisit it later.

PR #983, opened 2023-11-13
Improvement: Data Type Repository split

madsrasmussen:
Our Repositories are growing in size and we need to split them into multiple repositories based on responsibility.

Merges together with: https://github.com/umbraco/Umbraco-CMS/pull/15199


loivsen (changes requested):
The entity action delete uses `requestItems` which is from your new item repository, it currently seems that the action gets the detail repository.

Otherwise great updates 🥇

madsrasmussen:
> The entity action delete uses `requestItems` which is from your new item repository, it currently seems that the action gets the detail repository.
> 
> Otherwise great updates 🥇

Good catch! I will get it fixed

loivsen (approved):

PR #989, opened 2023-11-15
Feature: File System File Updates + Folder Repository Base

madsrasmussen:
## Feature: File System File Updates + Folder Repository Base

Merges together with: https://github.com/umbraco/Umbraco-CMS/pull/15521

A massive PR that lays the foundation to work with data from mocks to applications across different server models. It introduces the "unique" property so we can reuse the same UI components across different server data models. The main idea is that the server model only reaches the data source where it gets mapped to a client-side model.

Here are some highlights of the changes:

### Mock Databases
We now have two types of Mock databases. Entity Mock Database  + File System Mock Database. Each mock database type comes with a range of data managers: Detail, list, tree, etc. We can introduce more when needed. This means that we don't have to duplicate logic for each of the silos.

**Entity helpers**
https://github.com/umbraco/Umbraco.CMS.Backoffice/tree/feature-Folder-Repository-Base/src/mocks/data/entity

**File System helpers**
https://github.com/umbraco/Umbraco.CMS.Backoffice/tree/feature-Folder-Repository-Base/src/mocks/data/file-system

**Example of a mock db class**
https://github.com/umbraco/Umbraco.CMS.Backoffice/blob/feature-Folder-Repository-Base/src/mocks/data/stylesheet/stylesheet.db.ts

## Mock data
I have introduced a "Mock Model". Instead of having a detail model, tree item model, etc, we now have one big Mock Model. It combines all the other server models. The benefit of this is we don't have to maintain the same data across different models.

As part of the Mock data, I have started omitting the "type" field from the server models. We know that this one will be removed at one point, so it makes sense to start removing it now. We currently use `entityType` in the client models instead.

https://github.com/umbraco/Umbraco.CMS.Backoffice/blob/feature-Folder-Repository-Base/src/mocks/data/stylesheet/stylesheet.data.ts#L11

## Client-side Models.
I have started introducing client-side models for the file system silos + a couple of entity silos. This includes unique and entityType fields. See an example here: 

https://github.com/umbraco/Umbraco.CMS.Backoffice/blob/feature-Folder-Repository-Base/src/packages/templating/stylesheets/types.ts

## Folders
I have added a common folder modal + entity actions to support folders to any tree. They rely on a dataset with a unique field. This means that these can be used when a data set has been mapped to client models. The common folder logic includes creating, reading, updating, and deleting folders. Some trees don't support the renaming of folders yet. In that case, we just don't add a rename entity action.

https://github.com/umbraco/Umbraco.CMS.Backoffice/tree/feature-Folder-Repository-Base/src/packages/core/tree/folder

## File System Files (scripts, stylesheets, partial views, static files)
I have made a run-through of the code related to these files and cleaned up a bit. The same has happened on the server, so this PR includes new server models.

**Server File Path to unique:**
The server uses a path field as the identifier of a file. This field gets mapped to a URL-friendly unique. I have created this helper class to do that:
https://github.com/umbraco/Umbraco.CMS.Backoffice/blob/feature-Folder-Repository-Base/src/packages/core/server-file-system/server-path-unique-serializer.ts

**Stylesheet rules:**
Before this PR stylesheet rules were extracted and inserted server side. This is now a client-side job because it is based on a UI we invented and doesn't need to happen on the server. I have created this helper class to work with stylesheet rules. It is used in the Stylesheet workspace and Rich Text Editor.

https://github.com/umbraco/Umbraco.CMS.Backoffice/blob/feature-Folder-Repository-Base/src/packages/templating/stylesheets/utils/stylesheet-rule-manager.ts

## New Response Header in Management API
The Management API will return a new response header `Umb-Generated-Resource`. It will return the identifier of a newly created resource. We need to use this in the client, because we can use it directly with the auto-generated resources instead of the Absolute URL from the `Location` header.

### What to test:
**Scripts:**
* Load, create, update, and delete a script
* Create and delete a script folder

**Stylesheets**
* Load, create, update delete a stylesheet
* Update the stylesheet through the "Rule Editor" and the "Code Editor"
* Check that you can select stylesheets as part of the Tiny MCE config
* Check that you can select the style formats in the Tiny MCE property editor
* Create and delete a stylesheet folder

**Partial Views**
* Load, create, update, and delete a partial view
* Create and delete a partial view folder

**Document/Media Structure**
I have had to start adding the Client side models for Media Types, Document Types, and Documents as I ran into trouble with the media type folders and the reusable folder code. This resulted in updates to the structure manager. It was mainly around typescript models but also changes from `id` -> `unique` for the "outer" models. Please check that the doc type + media type editor still works. And check that a document loads the structure correctly.


nielslyngsoe (approved):
Blind approval to get it in to avoid merge conflicts. Seen in perspective of the PRs size.

Issue #997, opened 2023-11-18
[BUG]: Broken Storybook


enkelmedia:
**Describe the bug**
The Storybook that is referenced in the documentation and included in the repository looks like this:

![image](https://github.com/umbraco/Umbraco.CMS.Backoffice/assets/1782524/d27d6fc9-e66f-4b22-8256-721c27ae74c7)

Does not matter where I click I get the same error.

https://apidocs.umbraco.com/v14/ui?path=/docs/guides-context-api--docs

The version of Storybook that is public here: https://apidocs.umbraco.com/v14/ui?path=/docs/guides-context-api--docs seems to be older (probably preview003) but the one in the main-branch had another issue that this PR fixes.

iOvergaard:
This should be fixed implicitly when we release the next preview of v14, which we hope to do soon. If we find an opportunity to do so, we can probably fix it in-place. We'll see what comes first.

enkelmedia:
It's possible to fetch down the latest main-branch locally and `npm run storybook` to get it up and running while waiting for the next preview.

I'll close this =D

iOvergaard:
The main storybook should work now as well. We are currently in the process of preparing preview004 and Storybook has been updated as part of that: https://apidocs.umbraco.com/v14/ui/

PR #998, opened 2023-11-18
Fix import in preview.js to make Storybook work

enkelmedia:
Fixes this issue: https://github.com/umbraco/Umbraco.CMS.Backoffice/issues/997

The refactoring of `UmbDocumentTypeStore` to `UmbDocumentTypeDetailStore` made Storybook throw errors.

### DIFF
--- a/.storybook/preview.js
+++ b/.storybook/preview.js
@@ -13,7 +13,7 @@ import { UMB_MODAL_MANAGER_CONTEXT_TOKEN, UmbModalManagerContext } from '../src/
 import { UmbDataTypeTreeStore } from '../src/packages/core/data-type/tree/data-type.tree.store';
 import { UmbDocumentStore } from '../src/packages/documents/documents/repository/document.store';
 import { UmbDocumentTreeStore } from '../src/packages/documents/documents/repository/document.tree.store';
-import { UmbDocumentTypeDetailStore } from '../src/packages/documents/document-types/repository/detail/document-type-detail.store.ts';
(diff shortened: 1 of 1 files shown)


.storybook/preview.js
```diff
@@ -13,7 +13,7 @@ import { UMB_MODAL_MANAGER_CONTEXT_TOKEN, UmbModalManagerContext } from '../src/
 import { UmbDataTypeTreeStore } from '../src/packages/core/data-type/tree/data-type.tree.store';
 import { UmbDocumentStore } from '../src/packages/documents/documents/repository/document.store';
 import { UmbDocumentTreeStore } from '../src/packages/documents/documents/repository/document.tree.store';
-import { UmbDocumentTypeStore } from '../src/packages/documents/document-types/repository/document-type.store';
+import { UmbDocumentTypeDetailStore } from '../src/packages/documents/document-types/repository/detail/document-type-detail.store.ts';
```
iOvergaard, line 16:
```suggestion
import { UmbDocumentTypeDetailStore } from '../src/packages/documents/document-types/repository/detail/document-type-detail.store';
```

iOvergaard (approved):
I think the build of Storybook, that you reference in the issue is a lot older and is caused by a different import, that we already fixed. But it is good to get this import fixed as well, as that surely will cause the same problem in the future.

enkelmedia:
@iOvergaard You are right, the issue that was fixed in the PR is not the same as the issue in the screenshot. After applying the fix in the PR Storybook worked on the latest code in main.

PR #1001, opened 2023-11-20
Feature/file upload temp

loivsen:
upload


src/packages/core/components/input-upload-field/input-upload-field.element.ts
```diff
@@ -1,13 +1,15 @@
+import { UmbId } from '../../index.js';
```
iOvergaard, line 1: I think it is better to import from the alias here, no?

```suggestion
import { UmbId } from '@umbraco-cms/backoffice/id';
```

src/packages/core/components/input-upload-field/input-upload-field.element.ts
```diff
@@ -99,11 +127,14 @@ export class UmbInputUploadFieldElement extends FormControlMixin(UmbLitElement)
...
-		return html`${this.#renderFiles()} ${this.#renderDropzone()}`;
+		return html`<div id="wrapper">${this.#renderFilesWithPath()} ${this.#renderFilesUploaded()}</div>
+			${this.#renderDropzone()}${this.#renderButtonRemove()}`;
 	}
 
+	//TODO When the property editor gets saved, it seems that the property editor gets the file path from the server rather than key/id.
+	// This however does not work when there is multiple files. Can the server not handle multiple files uploaded into one property editor?
 	#renderDropzone() {
-		if (!this.multiple && this._currentFiles.length) return nothing;
+		if (!this.multiple && (this._currentFiles.length || this._filePaths.length)) return nothing;
```
iOvergaard, line 137: Did you test with an array as you suggested when we spoke last, or perhaps the server is meant to receive one file at a time?

As we talked about, I think it is on purpose you get the URL to the file, because the file is only uploaded into the filesystem on the server thus having no unique ID, which is a feature of a database otherwise. The URL is both the unique identifier as well as the public path to the file meaning, you can load the file into an `` tag for instance.

src/packages/core/components/input-upload-field/input-upload-field.element.ts
```diff
@@ -116,21 +147,41 @@ export class UmbInputUploadFieldElement extends FormControlMixin(UmbLitElement)
...
 	}
 
 	#handleRemove() {
-		// Remove via endpoint?
-		this._currentFiles = [];
+		this._filePaths = [];
+		const uniques = this._currentFiles.map((item) => item.unique) as string[];
+		this.#manager.remove(uniques);
+
+		this.dispatchEvent(new CustomEvent('change', { bubbles: true }));
```
iOvergaard, line 184: I think we have an `UmbChangeEvent` somewhere, could we use that here?

src/packages/core/temporary-file/temporary-file-manager.class.ts
```diff
@@ -0,0 +1,74 @@
...
+
+		if (!queue.length && this.getIsReady()) return;
+
+		this.#isReady.next(false);
+
+		queue.forEach(async (item) => {
+			if (item.status !== 'waiting') return;
+
+			const { error } = await this.#temporaryFileRepository.upload(item.unique, item.file);
+			await new Promise((resolve) => setTimeout(resolve, (Math.random() + 0.5) * 1000)); // simulate small delay so that the upload badge is properly shown
```
iOvergaard, line 57: If this is for testing purposes, you can use the `delay()` function in the MSW handler to simulate delays much like this.

iOvergaard (changes requested):
I like the UI of the uploader and excellent use of the loader circle tied up with actual upload progress. Definitely an improvement over the current backoffice! I also like to see more `repeat` over `map` in our Lit templates - nice job on that!

I think we need to resolve the TODOs left here before merging, though. I would like us to get a good understanding of temporary files on the server and make sure that this upload component and temporary file context support our use cases. I've left a few comments to that purpose down below.

src/packages/core/temporary-file/temporary-file-manager.class.ts
loivsen: For UX in general. I think it would be nicer if it was shown for at least one second. Uploads that are pretty much instant just looks like a blink.

PR #1007, opened 2023-11-21
Feature/user filters styling

JesmoDev:
Use popover container in user collection header and adds styling to make it look decent (not final design)

Add popover layout component to get consistent popover container styling

image


madsrasmussen (approved):
Looks good to me 👍

PR #1013, opened 2023-11-22
Feature/swap to popover container - part 1

JesmoDev:
Replaces all instances of the old popover element with the new popover-container element
also combines umb-dropdown and umb-button-with-dropdown into umb-dropdown.

Niels:
This PR also updates TypeScript as we needed latest features to be typed, as well it performes a format to match latest.


nielslyngsoe:
Hi @JesmoDev 

Are you using a special build of the `uui-popover-container`, cause when i run this, it looks like this: 
image

nielslyngsoe (approved):
Cool, works well. :) still a few visual things, but we will correct that as we go :-)

PR #1017, opened 2023-11-24
POC: Connected stores

madsrasmussen:
The PR introduces the concept of connected stores. It is a helper class to keep stores in sync. When using this class it is only needed to keep the detail store up to date, and then all stores depending on that model will get updates when changes happen and can respond accordingly.

The code is still in a POC state because we are missing information on the detailed model to keep everything in the tree in sync.


nielslyngsoe (approved):
Looks good

PR #1021, opened 2023-11-26
Adjust Image Cropper handle (focal point)

bjarnef:
## Description

This PR adjust focal point (handle) in Image Cropper to be more consistent with handle in Color Area.
https://github.com/umbraco/Umbraco-CMS/issues/16672

It moves the focal point based on `x` and `y` coordinates in the container on mouse click or touch.
Furthermore it moves in the handle / focal point on keyboard arrows, while holding down Shift key allows to fine-tune position.

## Types of changes

## Motivation and context

To re-use some existing logic and make the UI/UX consistent.

https://github.com/umbraco/Umbraco.CMS.Backoffice/assets/2919859/a9d80fce-ed13-4a5d-8fc4-e458c69054ce

## Checklist


src/packages/core/components/input-image-cropper/image-cropper-focus-setter.element.ts
```diff
@@ -1,4 +1,5 @@
-import { clamp } from './mathUtils.js';
+//import { clamp } from './mathUtils.js';
+import { drag, clamp } from '@umbraco-ui/uui-base/lib/utils';
```
bjarnef, line 2: Currently backoffice doesn't have a `drag` method and `clamp` is a copy from what already exists in UI library.
Should the clamp function be removed from backoffice repository or just re-use functions from UI library?

iOvergaard: No, you should be able to re-export them from the `external/uui` library and import it directly here:
```js
import { drag, clamp } from '@umbraco-cms/backoffice/external/uui';
```

bjarnef: I have updated this in https://github.com/umbraco/Umbraco.CMS.Backoffice/pull/1021/commits/c2c7b48e7e5c0682861fda5e46841255f32e50d7

bjarnef:
There a minor issue when resetting focal point, where it also need to reset the coords like it does on image loaded here:
https://github.com/umbraco/Umbraco.CMS.Backoffice/pull/1021/files#diff-edf1088805f907d36474f517086a2f1a7b8c0f07f1aeaa66d6d1861da99a6d12R72-R74

src/packages/core/components/input-image-cropper/image-cropper-focus-setter.element.ts
```diff
@@ -8,19 +9,28 @@ import {
...
 
 @customElement('umb-image-cropper-focus-setter')
 export class UmbImageCropperFocusSetterElement extends LitElement {
 	@query('#image') imageElement!: HTMLImageElement;
 	@query('#wrapper') wrapperElement!: HTMLImageElement;
 	@query('#focal-point') focalPointElement!: HTMLImageElement;
 
+	@state() private isDraggingGridHandle = false;
+
+	@state() private coords = { x: 0, y: 0 };
```
bjarnef, line 26: Any better way to track `x` and `y` coords? The `onMove()` function returns `x` and `y` based on the rendered image size (not natural size, but client size).

It seems focal point `left` and `top` properties are readonly - so I couldn't set there directly.

But I think it is fine the track `x` and `y` coords and calculate this to updated focal point style in `#setFocalPoint()`.

Maybe the logic to update `focalPointElement` style should be wrapped in a private function, since it does that a few times.

```
this.focalPointElement.style.left = `calc(${left * 100}% - ${this.#DOT_RADIUS}px)`;
this.focalPointElement.style.top = `calc(${top * 100}% - ${this.#DOT_RADIUS}px)`;
```

bjarnef:
@iOvergaard @JesmoDev please have a look 😎

nielslyngsoe:
Hi @bjarnef looks good.

I also see that you are trying to fix a few other things, like keyboard navigation etc. Thats cool, would it be possible to finish it of in near future?

Otherwise, I would recommend splitting your PRs up. Small PRs that are easy to review can sometimes be really good as they are fast to review and we can then get them merged in. In this way, we can keep the list of open PRs to a minimum. (as I was up for merging this, but its not done in its current state, just the change of the focus point look) :-)

bjarnef:
@nielslyngsoe it should be ready to review, but any feedback are welcome 😉
I am still learning by doing 🤭😅😎

bjarnef:
Do we have any way to know when reset focal point happens? It seems `firstUpdated()` happens on init and `updated()` each time focal point is moved, but we only want to update coords to center when focal point reset is triggered.

bjarnef:
I found a workaround to update coords if focal point is updated/reset to center using the reset button.
https://github.com/umbraco/Umbraco.CMS.Backoffice/pull/1021/commits/11666ac4e506924e5c308cac709f0fb8bf728364

nielslyngsoe:
@bjarnef yes, that is one solution. But the one we mainly use is to append setter and getter methods to the property itself. You should be able to fairly simple find a lot of solutions doing so. A typical one is the `value` and `config` property on any Property Editor UI.

bjarnef:
@nielslyngsoe I tried the following instead. https://github.com/umbraco/Umbraco.CMS.Backoffice/pull/1021/commits/03bcee207030af7fb2ba3dd4c30dd423329590f0

A bit similar to the `UmbInputImageCropperElement` here:
https://github.com/umbraco/Umbraco.CMS.Backoffice/blob/11efe83172c254e4027702a920fc6bd3cfe41a8a/src/packages/core/components/input-image-cropper/input-image-cropper.element.ts#L19

However to make `#resetCoords()` working it need to happen after image has loaded and CSS applied here to get the rendered size. 
https://github.com/umbraco/Umbraco.CMS.Backoffice/pull/1021/files#diff-edf1088805f907d36474f517086a2f1a7b8c0f07f1aeaa66d6d1861da99a6d12R92

Or call `#setFocalPointStyle()` but it requires `focalPointElement` to be found via query which seems to happen after property value setter.

bjarnef:
@JesmoDev this probably conflicts a bit with changes in https://github.com/umbraco/Umbraco.CMS.Backoffice/pull/1052

This PR makes the Image Cropper handle works using keyboard navigation and use same logic as color picker handle.

I could use some help with implementation of the property as @nielslyngsoe suggested - see last commit.

bjarnef:
@iOvergaard do you know how we can solve the last part?
I think it would be great to use same handle as in Color Picker and make it accessible as well.

bjarnef:
@iOvergaard @nielslyngsoe could you have another review at this?
Try moving focal point around and also navigate it with keyboard using arrow keys and eventually hold shift key to fine adjust.

Also while focal point not in center navigate reset button, hit enter and after focal point is moved reset to center, focus focal point again using keyboard and moving it should move it from center.

bjarnef:
@iOvergaard I fixed the tests as well.

There's a lot of `@typescript-eslint/no-explicit-any` warnings though.

image

bjarnef:
@nielslyngsoe could you have another look on this? 😁😎

bjarnef:
@iOvergaard sorry for pinging you again, but anything preventing us from re-using the existing `drag` logic used in color picker in Image Cropper focal point as well? 🤗

iOvergaard:
Hi @bjarnef 

Just wanted to get back to you. We would like to look into this functionality, but we are getting pretty close to the deadline for 14.0, so it will probably have to wait until after release. Meanwhile, I see a lot of out-commented code in your PR - do you need us to look further into it or are you going to update it?

bjarnef:
Hi @iOvergaard 

Yes, the commented code can be removed. When testing it seemed to work well to me and it use using the same logic as color picker instead of handle the handle different in Image Cropper :)

The only issue I had previous was resetting the handle, when clicking the reset button, but this should also have been fixed.

Only thing left may be trigger the `UmbChangeEvent`, which I noticed recently was refactored, when handle position change, but it isn't crucial.

bjarnef:
@iOvergaard I have fixed the merge conflicts.

A few things are noticed:
- `drag` function seems not to avaiable from `@umbraco-cms/backoffice/external/utils`, but `clamp` is available from `@umbraco-cms/backoffice/utils`, but it seems to be redundant from UI library.

- `wrapperElement` and `focalPointElement` isn't `HTMLImageElement` but `HTMLElement`

bjarnef:
I think it revolved this issues. However on initial change of focal post (click or drag) to value is undefined and thus return an console error.
It seems to log twice where the second time it has the initial value of focal point.

bjarnef:
@iOvergaard any update on this? Can I help with anything to close this?

iOvergaard:
> @iOvergaard any update on this? Can I help with anything to close this?

@bjarnef Going to fix the `clamp` issue in UUI for the upcoming 1.9 at least. Then we'll need to have another look at this PR. I don't know off the top of my head if we have to fix the media picker as well, which now also has a focal point.

bjarnef:
@iOvergaard @leekelleher I have updated this with latest changes. If media picker has a focal point I guess it would work, but haven't tested.

I guess I would need to run CMS project on same time to test we media? Is it possible the media library to have mock image similar to image cropper?

bjarnef:
@iOvergaard currently I have backoffice in a separate location.

It is possible to change path in `.gitmodules`:
```
[submodule "src/Umbraco.Web.UI.Client"]
	path = src/Umbraco.Web.UI.Client
	url = https://github.com/umbraco/Umbraco.CMS.Backoffice.git
```

or clone backoffice repo directly to `Umbraco.Web.UI.Client` folder?

iOvergaard:
@bjarnef The `Umbraco.Web.UI.Client` folder _is_ a clone of the backoffice. You can go into the folder and run git commands, e.g. `git checkout main`.

If you want to use your fork there, you can add it as a remote, e.g. `git remote add bjarnef https://github.com/bjarnef/Umbraco.CMS.Backoffice.git` and then `git checkout bjarnef/feature/image-cropper-handle`

bjarnef:
@iOvergaard thanks. I also needed to fetch from upstream/remote before checking out the branch.

It works great with focal point in media picker as well.
A few adjustments:

- Added top margin about actions buttons below image cropper, so if focal point is a bottom it doesn't overlap buttons.
- Only crosshair cursor on wrapper, when focal point is visible.
- Cancel drag/click in wrapper (area) when focal point is hidden.

bjarnef:
I included label property to the types, but I seems the `label` property is missing from endpoint at local crops in media picker property value?

image

I noticed it was shown at Image Cropper in media section.

image

but then I noticed it actually has e.g. "Square" as alias and the label was undefined.

image

Anyway shouldn't the endpoints map the label as well?

src/packages/media/media/components/input-image-cropper/image-cropper-focus-setter.element.ts
```diff
@@ -9,36 +12,48 @@ export class UmbImageCropperFocusSetterElement extends LitElement {
...
 
 	@property({ attribute: false })
-	focalPoint: UmbImageCropperFocalPoint = { left: 0.5, top: 0.5 };
+	get focalPoint() {
+		return this.#focalPoint;
+	}
+	set focalPoint(value) {
+		this.#focalPoint = value;
+		this.#onFocalPointUpdated();
+	}
```
iOvergaard, line 31: These need to be moved around according to Lit 3 warnings:

```suggestion
	set focalPoint(value) {
		this.#focalPoint = value;
		this.#onFocalPointUpdated();
	}
	get focalPoint() {
		return this.#focalPoint;
	}
```

src/packages/media/media/components/input-image-cropper/image-cropper-focus-setter.element.ts
```diff
@@ -9,36 +12,48 @@ export class UmbImageCropperFocusSetterElement extends LitElement {
...
+
+	#focalPoint: UmbImageCropperFocalPoint = { left: 0.5, top: 0.5 };
 
 	// TODO: [LK] Temporary fix; to be reviewed.
 	@property({ type: Boolean })
 	hideFocalPoint = false;
+	
+	@property({ type: Boolean, reflect: true }) disabled = false;
```
iOvergaard, line 39:
```suggestion
	@property({ type: Boolean, reflect: true })
	disabled = false;
```

src/packages/media/media/components/input-image-cropper/image-cropper-focus-setter.element.ts
```diff
@@ -9,36 +12,48 @@ export class UmbImageCropperFocusSetterElement extends LitElement {
...
 
 	@property({ type: String })
 	src?: string;
 
-	#DOT_RADIUS = 6 as const;
+	#DOT_RADIUS = 8 as const;
 
 	override disconnectedCallback() {
```
iOvergaard, line 46: This method can now be removed

src/packages/media/media/components/input-image-cropper/image-cropper-focus-setter.element.ts
```diff
@@ -9,36 +12,48 @@ export class UmbImageCropperFocusSetterElement extends LitElement {
...
 	protected override updated(_changedProperties: PropertyValueMap<any> | Map<PropertyKey, unknown>): void {
 		super.updated(_changedProperties);
 
 		if (this.hideFocalPoint) return;
 
-		if (_changedProperties.has('focalPoint') && this.focalPointElement) {
-			this.focalPointElement.style.left = `calc(${this.focalPoint.left * 100}% - ${this.#DOT_RADIUS}px)`;
-			this.focalPointElement.style.top = `calc(${this.focalPoint.top * 100}% - ${this.#DOT_RADIUS}px)`;
+		if (_changedProperties.has('focalPoint') && this.focalPoint) {
+			this.#setFocalPointStyle(this.focalPoint.left, this.focalPoint.top);
```
iOvergaard, line 56: Since there is now a setter for `focalPoint`, consider removing the `updated` hook and integrating it into the that setter.

src/packages/media/media/components/input-image-cropper/image-cropper-focus-setter.element.ts
```diff
@@ -79,73 +93,145 @@ export class UmbImageCropperFocusSetterElement extends LitElement {
...
+				@mousedown=${this.#handleGridDrag}
+        			@touchstart=${this.#handleGridDrag}>
+				<img id="image" @keydown=${() => nothing} src=${this.src} alt="" />
+				<span id="focal-point"
+					class=${classMap({
+						'focal-point--dragging': this.isDraggingGridHandle,
+						'hidden': this.hideFocalPoint
+					})}
+					tabindex=${ifDefined(this.disabled ? undefined : '0')}
+					aria-label="Focal Point"
```
iOvergaard, line 232: We need a localization key for this label, perhaps in the "visuallyHiddenTexts" category.

src/packages/media/media/components/input-image-cropper/image-cropper-focus-setter.element.ts
iOvergaard, line 1: Could we use this element as the focal point for the Media Picker?

src/packages/media/media/components/input-image-cropper/image-cropper-focus-setter.element.ts
```diff
@@ -9,36 +12,48 @@ export class UmbImageCropperFocusSetterElement extends LitElement {
...
 
 	@query('#wrapper')
-	wrapperElement?: HTMLImageElement;
+	wrapperElement?: HTMLElement;
 
 	@query('#focal-point')
-	focalPointElement!: HTMLImageElement;
+	focalPointElement!: HTMLElement;
+
+	@state() private isDraggingGridHandle = false;
```
iOvergaard, line 20:
```suggestion
	@state()
	private isDraggingGridHandle = false;
```

src/packages/media/media/components/input-image-cropper/image-cropper-focus-setter.element.ts
```diff
@@ -9,36 +12,48 @@ export class UmbImageCropperFocusSetterElement extends LitElement {
...
+
+	@state() private coords = { x: 0, y: 0 };
```
iOvergaard, line 22:
```suggestion
	@state()
	private coords = { x: 0, y: 0 };
```

src/packages/media/media/components/input-image-cropper/image-cropper-focus-setter.element.ts
iOvergaard: Should also be renamed to `_isDraggingGridHandle` so it's ignored in the debug element.

src/packages/media/media/components/input-image-cropper/image-cropper-focus-setter.element.ts
iOvergaard: Should also be renamed to `_coords` so it's ignored in the debug element.

src/packages/media/media/components/input-image-cropper/image-cropper-focus-setter.element.ts
iOvergaard, line 1: These suggestions are not part of this PR, but thought it would make sense to fix:

It seems that everything in `update()` could be moved to a setter for `src`.

And `firstUpdated` only sets a CSS variable, which we could probably move directly to the `styles` property since it only sets a static pixel value.

src/packages/media/media/components/input-image-cropper/image-cropper-focus-setter.element.ts
```diff
@@ -79,73 +93,145 @@ export class UmbImageCropperFocusSetterElement extends LitElement {
...
+		const top = clamp((y / height), 0, 1);
+
+		this.#coordsToFactor(x, y);
+
+		this.dispatchEvent(
+			new CustomEvent('change', {
+				detail: { left, top },
+				bubbles: false,
+				composed: false,
+			}),
```
iOvergaard, line 128: Consider creating a custom event class for this to stay in line with other custom events.

iOvergaard (changes requested):

src/packages/media/media/components/input-image-cropper/image-cropper-focus-setter.element.ts
iOvergaard: I was confused, it is already used there.

iOvergaard:
@bjarnef 
> The crops should probably also show Label instead of Alias on the right as in old backoffice. It seems the array of crops in UmbImageCropperPropertyEditorValue type didn't have a label property as in configuration.

Yes, you are right. We need to look that up in the property editor, which we already did, but didn't use for stored crops. I have fixed that in #2145. Thanks for your discovery here!

src/packages/media/media/components/input-image-cropper/image-cropper-focus-setter.element.ts
bjarnef: Yeah, this was just how it already works, so didn't change that part. But a custom event class is probably better.

bjarnef: @iOvergaard should the custom event be added here?
https://github.com/umbraco/Umbraco.CMS.Backoffice/tree/97f256a195d8301bcc68ba21ed879240cd0d663d/src/packages/core/event

src/packages/media/media/components/input-image-cropper/image-cropper-focus-setter.element.ts
bjarnef: So move `#setFocalPointStyle()` insider setter for `focalPoint`?

src/packages/media/media/components/input-image-cropper/image-cropper-focus-setter.element.ts
bjarnef: Does it need to extends `UmbLitElement` instead of `LitElement` to access localize?

```
aria-label="${this.localize.term('visuallyHiddenTexts_focalPoint')}"
```

If just "Focal Point" we can eventually use `general_focalPoint`?

bjarnef:
@iOvergaard feel free to make additional commits to this PR :) 

I noticed some other behavioural oddities.

- Change media item allow me to select a folder
- Without focal point hidden and dragging image it doesn't seem preview thumbnail updates.
- Changing current selected crop reset zoom position which wasn't the case in old backoffice https://github.com/umbraco/Umbraco-CMS/issues/16821
- I can't navigate crops using keyboard https://github.com/umbraco/Umbraco-CMS/issues/16820

src/packages/media/media/components/input-image-cropper/image-cropper-focus-setter.element.ts
iOvergaard: Nah, I would put it in the components/input-image-cropper folder as well and call it something like `UmbImageCropChange`. From there you can export it through the index.ts file.

Also be sure to add the crops directly to the event class rather than use `detail`.

src/packages/media/media/components/input-image-cropper/image-cropper-focus-setter.element.ts
iOvergaard: Yes and yes.  You need UmbLitElement, which it should have been in the first place. Actually made the same change to UmbLitElement over in #2145.

src/packages/media/media/components/input-image-cropper/image-cropper-focus-setter.element.ts
iOvergaard: Yes, I thought if it just needs to react on the changes to `focalPoint`, then it may as well be in the setter of that and we can remove this whole method.

iOvergaard:
@bjarnef All good proposals. Would you add them as separate issues, please?

> Change media item allow me to select a folder (it wasn't possible in old backoffice)

This one we are aware of, we currently have no way of detecting if a media item is actually a folder - all media items can contain subitems, but obviously does not make sense to select one if it itself doesn't have an `umbracoFile`. We need more advanced checks for that.

src/packages/media/media/components/input-image-cropper/image-cropper-focus-setter.element.ts
bjarnef: Okay, makes sense.. I will have a look when https://github.com/umbraco/Umbraco.CMS.Backoffice/pull/2145 has been merged :)

src/packages/media/media/components/input-image-cropper/image-cropper-focus-setter.element.ts
bjarnef: I called it `UmbFocalPointChangeEvent` as it trigget when focal point is set (click or drag):
https://github.com/umbraco/Umbraco.CMS.Backoffice/pull/1021/commits/f1449ddab26802a46750e4e101e78b7cdca41e78

not which event this one is?

image

Do we want `top` and `left` properties like in `UmbProgressEvent`?
https://github.com/umbraco/Umbraco.CMS.Backoffice/blob/97f256a195d8301bcc68ba21ed879240cd0d663d/src/packages/core/event/progress.event.ts#L7

Perhaps there is an type containing `top` and `left` properties? and if it anytime later need additional properties?

iOvergaard: There is the `UmbFocalPointModel` which would be suitable to use.

bjarnef: I replaced the custom event in image crop as well with `UmbImageCropChangeEvent`.

So we have these two:

- `UmbImageCropChangeEvent`
- `UmbFocalPointChangeEvent`

bjarnef:
@iOvergaard can you have another look at latest change. I think there are other enhancements, we can fix in other PRs.
I added two new custom events, but I think currently these introduce breaking changes and they need to have event name `change`? But if developers already implemented this it probably wouldn't work reacting on `CustomEvent` anyway.

src/packages/media/media/components/input-image-cropper/image-cropper-focus-setter.element.ts
bjarnef: I have updated this and a few other labels.

src/packages/media/media/components/input-image-cropper/image-cropper-focus-setter.element.ts
bjarnef: This has been updated as well.

bjarnef:
@iOvergaard can you review this with the latest changes. It would be great to close this one.

bjarnef:
@iOvergaard did you see the latest comments/ commits? It would be great to finish this one 😊

PR #1023, opened 2023-11-27
Bugfix: Modal wait for close animation

madsrasmussen:
The `close-end` event was removed by mistake from UI Library and we had to implement a temporary solution and listen to the `close` event instead. The `close-end` event i now back, and this PR updates the code to listen to that instead.

The result is that modals now show the full slide-out animation before being removed from the DOM.


nielslyngsoe (approved):
cool

PR #1029, opened 2023-11-28
Bugfix: Render User Group Ref icon

madsrasmussen:
We were using the wrong attribute name to pass the icon name into an `uui-icon` element. This PR updates the code to use the correct attribute.

How to test
* Make sure that a user group ref renders an icon. (Test the user group picker on a user)


nielslyngsoe (approved):
good

PR #1030, opened 2023-11-28
Improvement: Direct usage of selection manager instead of proxy methods

madsrasmussen:
This PR makes the selection manager public so we can call methods directly on the manager instead of proxying them with methods on the implementing class. This will simplify the interface for that class.

See the tree context interface as an example of how many methods can be removed from the interface:

https://github.com/umbraco/Umbraco.CMS.Backoffice/pull/1030/files#diff-63700e0510ec22c2995269f8338ca4b82d8a0dcb612b06b2dd93e90c621e909aL20

The PR also adds tests for the Selection Manager:

https://github.com/umbraco/Umbraco.CMS.Backoffice/pull/1030/files#diff-6f2e37bb4ffdcac7a21718fccc2817722cfee34cae23b7b6d702f7d775f3a17bR1


src/shared/utils/selection-manager/selection.manager.ts
```diff
@@ -64,8 +101,11 @@ export class UmbSelectionManager {
 	 * @memberof UmbSelectionManager
 	 */
 	public select(unique: string | null) {
+		if (this.getSelectable() === false) return;
+		if (this.isSelected(unique)) return;
 		const newSelection = this.getMultiple() ? [...this.getSelection(), unique] : [unique];
 		this.#selection.next(newSelection);
+		this._host.getHostElement().dispatchEvent(new UmbSelectionChangeEvent());
```
nielslyngsoe, line 108: Since this is a Controller it self, you can omit `._host` as the getHostElement() is recursive. (Its gets the parent controller that has an element)

src/shared/utils/selection-manager/selection.manager.ts
```diff
@@ -74,8 +114,10 @@ export class UmbSelectionManager {
 	 * @memberof UmbSelectionManager
 	 */
 	public deselect(unique: string | null) {
+		if (this.getSelectable() === false) return;
 		const newSelection = this.getSelection().filter((x) => x !== unique);
 		this.#selection.next(newSelection);
+		this._host.getHostElement().dispatchEvent(new UmbSelectionChangeEvent());
```
nielslyngsoe, line 120: Same here :-)

src/packages/core/modal/common/section-picker/section-picker-modal.element.ts
```diff
@@ -32,17 +32,6 @@ export class UmbSectionPickerModalElement extends UmbModalBaseElement<
 			'umbSectionsObserver';
 	}
 
-	#submit() {
-		this.value = {
-			selection: this.#selectionManager.getSelection(),
```
nielslyngsoe, line 37: I didnt test, just looking at the code, but it seems to me that this is missing now? Or how does this modal set the value of the modal.. well I could have investigated, but it would be good if you had a look at this :-)

madsrasmussen: Good catch. It was removed by mistake. It is back again.

nielslyngsoe (approved):
good :-)

PR #1033, opened 2023-11-29
Feature/couple fixes of errors

loivsen:
## Types of changes

## Checklist

### DIFF
--- a/src/packages/core/components/code-block/code-block.element.ts
+++ b/src/packages/core/components/code-block/code-block.element.ts
@@ -1,4 +1,4 @@
-import { UmbTextStyles } from "@umbraco-cms/backoffice/style";
+import { UmbTextStyles } from '@umbraco-cms/backoffice/style';
 import {
 	css,
 	html,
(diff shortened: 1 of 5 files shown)


src/packages/core/components/alias-input/alias-input.element.ts
```diff
@@ -0,0 +1,112 @@
...
+			this.parentElement.removeEventListener('input', this.#parentChange);
+		}
+	}
+
+	#onToggleAliasLock() {
+		this.locked = !this.locked;
+	}
+
+	#onChange(e: UUIInputEvent) {
+		super.value = e.target.value as string;
```
iOvergaard, line 58: The current class is in itself a form control with a setter for the value, but you are setting the value only on the super element thereby bypassing the checks you have made in the "set value" function. Is that intended?

src/packages/core/components/alias-input/alias-input.element.ts
```diff
@@ -0,0 +1,112 @@
...
+		this.dispatchEvent(new UmbChangeEvent());
+	}
+
+	#parentChange = (e: Event) => {
+		if (!this.pristine) return;
+		super.value = (e as UUIInputEvent).target.value as string;
```
iOvergaard, line 64: Bypass of `this.value` here as well as intended?

src/packages/templating/templates/components/alias-input/alias-input.ts
```diff
@@ -6,15 +6,11 @@ import { UmbChangeEvent } from '@umbraco-cms/backoffice/event';
...
-        <input placeholder="Enter alias..." .value=${this.value} ?disabled=${!this.isOpen} @input=${
-					this.#setValue
-				}></input>
-
-        `;
+		return html`!
+			<uui-button compact @click=${this.#handleClick} label="unlock alias input">
+				<uui-symbol-lock .open=${this.isOpen}></uui-symbol-lock>
+			</uui-button>
+			<input placeholder="Enter alias..." .value=${this.value} ?disabled=${!this.isOpen} @input=${this.#setValue} /> `;
```
iOvergaard, line 13: Is this perhaps meant to use the new umb-alias-input element?

iOvergaard (changes requested):
Good to see more translation implementations, and looks like a good round of cleanup here. Just left a few comments/questions.

PR #1034, opened 2023-11-29
Feature: Collection Default View

madsrasmussen:
This PR introduces an `UmbCollectionViewManager` to simplify the CollectionContext class. It keeps track of all registered collection views and returns matching routes. You can pass a configuration to the CollectionViewManager with a defaultViewAlias and that will be used as the fallback route if the URL doesn't match any of the other routes.

How to use:

```js
const manager = new UmbCollectionViewManager(host, { defaultViewAlias: 'Umb.CollectionView.Alias' });
manager.routes.subscribe(routes =>  // pass to router );
```

The manager is part of the CollectionContext and can be reached like this:
```js
const context = new UmbCollectionContext(host, { defaultViewAlias: 'Umb.CollectionView.Alias' });
```

_Subscribe to views:_
```js
context.view.views.subscribe()
```

_Subscribe to routes:_
```js
context.view.routes.subscribe()
```

**How to test:**
* Make sure that you can navigate to each collection view via URL
-.../users/collection/grid
-.../users/collection/table

* Make sure that the user's collection selects the default view. The default view for users is set to the grid. Navigate between sections. When going into the user section it should end up on the grid view.
-Change the default view to the table. Change the const `UMB_COLLECTION_VIEW_USER_GRID` to `UMB_COLLECTION_VIEW_USER_TABLE` https://github.com/umbraco/Umbraco.CMS.Backoffice/blob/feature/collection-default-view/src/packages/user/user/collection/user-collection.context.ts#L12


madsrasmussen:
@iOvergaard Good point. I have added some test notes in the description. Let me know if it doesn't make sense.

iOvergaard (approved):

PR #1036, opened 2023-11-29
Feature: UI Culture Input Element

madsrasmussen:
This PR adds a UI culture input element.

To make the UI culture selection reusable, and to simplify the user components I have moved the logic for UI culture selection into its own element. It works just like all the other input elements.

During the changes, I noticed that the value always was tied to the current user, so all users would get the same value. I have updated this too, so the value now comes from the user that is open in the UI like you would expect.

At some point, we also want to allow the logged-in user to change their own UI culture through the current user dialog. That task will be easier with the new input element.

Usage:
```html
<umb-ui-culture-input value="en-us" @change=""></umb-ui-culture-input>
```

How to test:
* Make sure it is possible to change and persist the UI culture for any user


nielslyngsoe (approved):
Looks good

PR #1044, opened 2023-12-04
Refactoring of Variant Context (Property dataset)

nielslyngsoe:
Make it possible for anyone to make use of our Property Editors UI.

**Background story:**
Currently its not that simple, because a Property Editor UI might request information about its environment. This has been solved by Variant Context — which is now a requirement for any Property Editors, which makes it harder to implement a Property Editor UI.

The Variant Context brings Name, information about all the Properties of this set and VariantId. This means:

It is possible for Property Editors to retrieve the name of the Variant without awareness about a variantID. (Same integration no matter variant or none variant environment)

It is possible for Property Editors to utilise the value of other properties. And without knowing about the variant configuration. So the code to do so is the same for the variant and the invariant scenario. In other words the resposnsibility has been moved to the Variant Context. — Property Editors written without this in mind will work in both cases.

**Problem:**
But to enable anyone to start use Property Editor UIs, then they have to create a Property Datatset Context, so here comes the Property Dataset Element. ( and a base implementation of such context)

**Solution**
Basic Variant Context (maybe name should be re-thought).
This holds a name and a property-map (A array of properties alias and value)

So if you like to implement one or more Property Editor UIs. you can now do this:

```
const dataset = [
	{
		alias: 'myAlias',
		value: 'testValue',
	},
];

<umb-property-dataset name='dataset name' .value=${dataset}>
	<umb-property alias="myAlias" property-editor-ui-alias="Umb.PropertyEditorUi.TextBox"></umb-property>
	...
</umb-property-dataset>
```

In this way you just need to implement the <umb-property-dataset>. And then either use observables or listening to the 'change' event and getting the data based on this.

A minor detail, but notice how this enables the UI to be hidden, and we still have the full data set available for manipulations and observations. This is useful looking at tabs or just conditional props.

**Content of PR:**
- A dev mode for tests
- Refactoring of variant-context, workspace-property and workspace-property-layout.
- Example that demoes the property-dataset with properties.
- New import map called `@umbraco-cms/backoffice/property`

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to change) (**renaming of files**)

## Checklist


src/packages/core/property/property-dataset/property-dataset.element.ts
```diff
@@ -0,0 +1,121 @@
...
+
+/**
+ *  @element umb-property-dataset
+ *  @description - Element for hosting a property dataset. This is needed for umb-property to work.
+ *  @slot default - Slot for rendering content within.
+ */
+@customElement('umb-property-dataset')
+export class UmbPropertyDatasetElement extends UmbLitElement {
+	// A take on only firing events when the value is changed from the outside.
+	#silentOnce = true;
```
madsrasmussen, line 15: What is the use case of this? It goes against the nature of events that it is when changes happen inside a system to notify the outside. The outside should already know what the value is. 

Maybe we can find a different approach to solve the use case.

src/packages/core/workspace/workspace-property-dataset/invariant-workspace-property-dataset-context.ts
```diff
@@ -0,0 +1,65 @@
...
+	#workspace: WorkspaceType;
+
+	name;
+
+	// default data:
+
+	getVariantId() {
+		return UmbVariantId.CreateInvariant();
+	}
+	getType() {
```
madsrasmussen, line 29: I have started to use the name `entityType` explicitly instead of just `type`. Would it be ok to do the same here so the prop is `getEntityType`? I have noticed it has removed some confusion in other places in the code because `type` is used a lot for many different things.

src/packages/documents/documents/property-dataset-context/document-property-dataset-context.token.ts
```diff
@@ -0,0 +1,13 @@
+import { UmbPropertyDatasetContext } from '@umbraco-cms/backoffice/property';
+import { UMB_DOCUMENT_ENTITY_TYPE } from '../entity.js';
+import type { UmbDocumentPropertyDataContext } from './document-property-dataset-context.js';
+import { UmbContextToken } from '@umbraco-cms/backoffice/context-api';
+
+export const IsDocumentVariantContext = (
+	context: UmbPropertyDatasetContext,
+): context is UmbDocumentPropertyDataContext => context.getType() === UMB_DOCUMENT_ENTITY_TYPE;
```
madsrasmussen, line 8: regarding the rename to `entityType`. I think this here is a good example where it can be confusing what kind of `type` you get. It could have been some sort of context type.

examples/dashboard-with-property-dataset/dataset-dashboard.ts
```diff
@@ -0,0 +1,75 @@
...
+		this.data = (e.target as UmbPropertyDatasetElement).value;
+		this.requestUpdate('data', oldValue);
+	}
+
+	render() {
+		return html`
+			<uui-box class="uui-text">
+				<h1 class="uui-h2" style="margin-top: var(--uui-size-layout-1);">Dataset Example</h1>
+				<umb-property-dataset .value=${this.data} @change=${this.#onDataChange}>
+					<umb-property
```
madsrasmussen, line 26: Would it be possible to set a value prop on the property element so it could work without the dataset context? Or how much chaos would that create? :D

madsrasmussen: I know it is not the purpose of the example. It would be to "simplify" the simple use case of a property element.

src/packages/core/data-type/workspace/data-type-workspace.context.ts
```diff
@@ -146,8 +145,46 @@ export class UmbDataTypeWorkspaceContext
 		return this._configDefaultData?.find((x) => x.alias === alias)?.value;
 	}
 
-	createVariantContext(host: UmbControllerHost): UmbDataTypeVariantContext {
-		return new UmbDataTypeVariantContext(host, this);
+	createPropertyDatasetContext(host: UmbControllerHost) {
```
madsrasmussen, line 148: If I understand this correctly would it then be possible to make this as a "sub context" or use the property-dataset element to achieve the same?

src/packages/core/property/property-dataset/property-dataset.element.ts
nielslyngsoe: Yeah, i was a bit torn by this one. Cause we usually only fire events when the user did some interaction. But in this case we are too far from interaction for me to know wether its interaction or just caused by usual data change. So this is a fix, so when setting data via example the props it does not trigger an update. In other words, If i (as the implementation) put new data, I dont want a event that there is new data? Maybe lets talk about this in a huddle?

src/packages/core/workspace/workspace-property-dataset/invariant-workspace-property-dataset-context.ts
nielslyngsoe: Yeah cool. And good perspective.  This is just transferred from the existing interface, so i'm all in for changing this.

src/packages/core/data-type/workspace/data-type-workspace.context.ts
nielslyngsoe: We had a good talk and I'm moving the creation of this context to the editor implementation. Not a sub context, but a more local implementations.

src/packages/core/property/property-dataset/property-dataset.element.ts
nielslyngsoe: We have implemented some corrections from a huddle talk we had.

PR #1046, opened 2023-12-06
Bugfix/revisit modals

nielslyngsoe:
Some modals was broken, so I finished off the work on Modal Value.

This means a modal now accepts a config and a value (before there where part of one) and then the value is the one returned as well. This gives modals some base feature to work with the value. As well the value can be observed from outsiders if they need to keep up to date with the changes.

## Types of changes

## Checklist


iOvergaard (approved):
This fixes at least the document type editor and adding new properties / selecting existing configurations. LGTM.

PR #1054, opened 2023-12-11
Make notifications always be on top of the page

JesmoDev:
This PR makes notifications alway be on top of the page by using the popover API.
Before, notifications would be hidden behind modals, sidebars, dialogs and so on.

Explanation:
This solution is a little hacky, as the only way for content be at the very top of the #top-layer is to be the most recently added element. So  we close and instantly open the popover (notification container) when a new notification is added. The closing and opening happens in the same frame/tick, so it is not visible.

## Types of changes

## Checklist

### DIFF
--- a/src/packages/core/components/backoffice-notification-container/backoffice-notification-container.element.ts
+++ b/src/packages/core/components/backoffice-notification-container/backoffice-notification-container.element.ts
@@ -56,7 +56,6 @@ export class UmbBackofficeNotificationContainerElement extends UmbLitElement {
 		UmbTextStyles,
 		css`
 			#notifications {
-				overflow: auto;
 				top: 0;
 				left: 0;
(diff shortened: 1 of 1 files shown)


madsrasmussen:
I think the “correct” solution to the issue is to utilize the context API and have some sort of notification container in the modal layer. If the notification should still show in the lower right corner or it should be something different for modals I don’t know.

JesmoDev:
The modal container just holds the modals, but they all go to the #top-layer when they are opened. Which means that they will go on-top of anything except for items sent to the #top-layer after. So the only way to make notifications sit on top is by sending them to the #top-layer after any modals. 

So an alternative to this would be to send each individual notification to the #top-layer when it's opened.

nielslyngsoe (approved):
good

PR #1056, opened 2023-12-11
Feature: User avatar

madsrasmussen:
This PR finalizes the user avatar upload feature.

How to test:
* Make sure you can upload an avatar for a user that is not you
* Make sure you can upload an avatar for your user. Check that the current user avatar in the top right corner updates.
* Make sure that the avatar in the user grid + user table is updated.


src/packages/user/current-user/current-user-header-app.element.ts
```diff
@@ -45,14 +51,70 @@ export class UmbCurrentUserHeaderAppElement extends UmbLitElement {
...
+				url: isRelativeUrl ? serverUrl + avatarScale1 : avatarScale1,
+			},
+			{
+				scale: '2x',
+				url: isRelativeUrl ? serverUrl + avatarScale2 : avatarScale2,
+			},
+			{
+				scale: '3x',
+				url: isRelativeUrl ? serverUrl + avatarScale3 : avatarScale3,
+			},
```
bjarnef, line 89: @madsrasmussen maybe rename `scale` to `density` if it is directly related to the `srcset` attribute as in current backoffice. Instead of 1x, 2x, 3x (DPR - device pixels ratio) it could also have used `w` based on viewport size, e.g. 600w, 1000w etc.

nielslyngsoe: Hi Bjarne, good feedback,  I think renaming scale to density sounds good. I would recommend making a PR with that change, then I will happily approve it :-)

This one is getting merged for now.

nielslyngsoe (approved):
Looks good

PR #1057, opened 2023-12-12
Remove create button from language collection row

JesmoDev:
Removes the create button in the entity action dropdown on the language collection so that there is only remove.
Later we will find a way to only show a dropdown menu if there is more than one action


madsrasmussen (approved):
Looks good to me 👍

PR #1058, opened 2023-12-12
element types only filter

loivsen:
Document Type Picker can now filter on isElement


src/packages/core/tree/tree.context.ts
```diff
@@ -49,6 +49,8 @@ export class UmbTreeContextBase<TreeItemType extends UmbTreeItemModelBase>
 	public repository?: UmbTreeRepository<TreeItemType>;
 	public selectableFilter?: (item: TreeItemType) => boolean = () => true;
 
+	public filter?: (item: TreeItemType) => boolean = () => true;
```
nielslyngsoe, line 52: Couldn't we use the already existing `selectableFilter`

I think, without having tested, that it would be perfect for the case. Cause we dont want to hide items, we just want to prevent the user from picking these.

loivsen: We can do that too. I thought it would better to hide the elements but perhaps just disabling them is fine.

PR #1060, opened 2023-12-12
Bugfix: Dont do full page reload when creating language from the entity action button

JesmoDev:
Fixes an issue where the create entity action would do a full page refresh.

## Types of changes

## Checklist


madsrasmussen:
I would love it if we could try and investigate why the href solution doesn't work. I think we have the same solution in other places that don't do a full-page reload.

Can you investigate what we do when creating docs, data types etc.?

JesmoDev:
I got this solution from documents.
Document types uses a href on the menu item in the modal that get's opened:
`href=${´section/settings/workspace/document-type/create/${this.data?.parentKey || 'null'}´}`

I can only find usage of the `getHref()` function in the language create action and in the docs `entity-actions.mdx`

iOvergaard (approved):

PR #1064, opened 2023-12-14
fix: workspace router re-initialization

iOvergaard:
make sure workspaces reinitializes all main routes when they change

some workspaces such as data type had a problem when you switched from one datatype to another that it wouldn't refresh the sub-routes


madsrasmussen (approved):
Looks good to me

PR #1065, opened 2023-12-15
Server models: propertyEditorAlias => editorAlias

iOvergaard:
update new models to account for propertyEditorAlias => editorAlias


leekelleher (approved):
Looks good to me.

PR #1083, opened 2023-12-20
Tree Picker: "Allow items of type" filter picker

leekelleher:
Added `umb-property-editor-ui-tree-picker-filter` to filter content-types from the selected start-node.

Observes the DataType workspace for changes to the `startNode` configuration value, updates the content-type picker (e.g. documents, media or members) accordingly.

Note: I have included a reference to the `<umb-member-type-input>` components, _(which does not exist yet),_ but it will in due course, _(the assumption is that we follow the same naming conventions)._


src/packages/core/property-editor/uis/tree-picker/config/filter/property-editor-ui-tree-picker-filter.element.ts
```diff
@@ -0,0 +1,95 @@
...
+import { UmbLitElement } from '@umbraco-cms/internal/lit-element';
+
+@customElement('umb-property-editor-ui-tree-picker-filter')
+export class UmbPropertyEditorUITreePickerFilterElement extends UmbLitElement implements UmbPropertyEditorUiElement {
+	constructor() {
+		super();
+
+		this.consumeContext(UMB_DATA_TYPE_WORKSPACE_CONTEXT, (instance) => {
+			const workspace = instance;
+			this.observe(workspace.data, (data) => {
```
nielslyngsoe, line 17: Hi @leekelleher :-)

In this case you can use: `workspace.propertyValueByAlias()` and in that way only listen for that specific value. This will lower the amount of executions on this callback. In other words be more performant.

Ideally we always provide Observables that can be specific to most use cases.

Just a bit of background: :-)
In above case the `.data` property is all data, and will be triggered every time there is one change to it.
And then we have several Observables ( offen made via a helper method called .asObservablePart(...) ) which then only triggers if that specific part changes.

src/packages/core/property-editor/uis/tree-picker/config/filter/property-editor-ui-tree-picker-filter.element.ts
```diff
@@ -0,0 +1,95 @@
+import { UMB_DATA_TYPE_WORKSPACE_CONTEXT } from '../../../../../data-type/workspace/data-type-workspace.context.js';
+import { UmbDocumentTypeInputElement } from '../../../../../../documents/document-types/components/document-type-input/document-type-input.element.js';
+import { UmbMediaTypeInputElement } from '../../../../../../media/media-types/components/media-type-input/media-type-input.element.js';
+import { StartNode } from '@umbraco-cms/backoffice/components';
+import type { UmbPropertyEditorUiElement } from '@umbraco-cms/backoffice/extension-registry';
+import { customElement, html, property } from '@umbraco-cms/backoffice/external/lit';
+import { UmbTextStyles } from '@umbraco-cms/backoffice/style';
+import { UmbLitElement } from '@umbraco-cms/internal/lit-element';
+
+@customElement('umb-property-editor-ui-tree-picker-filter')
```
nielslyngsoe, line 10: I had a really hard time getting the context/meaning of this name, so I think it would be good to give it some focus.

My usual go about this is to mention some different perspectives, to let you into my thinking:

So this one can pick a Type, can so far it can only pick content-types. We have a few other property editor that can pick something, we usually name them Picker.
So I was thinking something along this line:

`umb-property-editor-ui-treepicker-content-type-picker`.

( I put treepicker in there as this is very specific to this one case )

Let me know what you think — We might also need to rename the StartNode one os they are aligned :-)

src/packages/core/property-editor/uis/tree-picker/config/filter/property-editor-ui-tree-picker-filter.element.ts
```diff
@@ -0,0 +1,95 @@
...
+export class UmbPropertyEditorUITreePickerFilterElement extends UmbLitElement implements UmbPropertyEditorUiElement {
+	constructor() {
+		super();
+
+		this.consumeContext(UMB_DATA_TYPE_WORKSPACE_CONTEXT, (instance) => {
```
nielslyngsoe, line 15: Actually, just to be super duper correct, then there is no need to utilize the Workspace in this case.

Not that the perspective I will mention now does not have any effect on this case, but I wanted to let you know, for your general knowledge of this:

So we have a property-dataset-context, which is more local than the workspace. Its generic and will always be present when we utilise property editor UIs. (Where a Workspace is not necessary, especially the specific version (Data-type workspace))

So again, this does not make any difference since this Property Editor UIs is very specific to this specific Data-Type.

But if you used the Property Dataset Context instead, then this Property Editor UI could be used in any other scenario, as long as there is another property with the alias StartNode (that matches that data-structure)

let me know if you like these insights or if its just noise :-)

src/packages/core/property-editor/uis/tree-picker/config/filter/property-editor-ui-tree-picker-filter.element.ts
```diff
@@ -0,0 +1,95 @@
...
+					const startNode = property.value as StartNode;
+					if (startNode.type) {
+						this.sourceType = startNode.type;
+					}
+				}
+			});
+		});
+	}
+
+	@property({ type: Array })
```
nielslyngsoe, line 29: Can I recommend/wish to put the properties in the top of the class.. above the constructor? :-) 😬

src/packages/core/property-editor/uis/tree-picker/config/filter/property-editor-ui-tree-picker-filter.element.ts
```diff
@@ -0,0 +1,95 @@
...
+	value?: string[];
+
+	@property({ attribute: false })
+	sourceType: string = 'content';
```
nielslyngsoe, line 33: `sourceType` is a internal property. So you should mark it with `private` then TypeScript will complain if someone tries to access this. (we also have JS private fields, that is marked with # as the first character, but these does not cope with Lit. So we just use TypeScript for these.)

Also since this is internal, we can spare some performance and just make the property a Lit `@state()` thats equivalent to `@property` just internal. (where property comes with features to be changed from the outside, sync with attribute and all that)

src/packages/core/property-editor/uis/tree-picker/config/filter/property-editor-ui-tree-picker-filter.element.ts
```diff
@@ -0,0 +1,95 @@
...
+		super();
+
+		this.consumeContext(UMB_DATA_TYPE_WORKSPACE_CONTEXT, (instance) => {
+			const workspace = instance;
+			this.observe(workspace.data, (data) => {
+				const property = data?.values.find((setting) => setting.alias === 'startNode');
+				if (property) {
+					const startNode = property.value as StartNode;
+					if (startNode.type) {
+						this.sourceType = startNode.type;
```
nielslyngsoe, line 22: If the source type changes, then we need to make sure we reset the value. So we dont carry over ex. Document-Types to the Media-Types picker.

One solution I see to this, would be to not preset the sourceType to content, but have it as undefined initially.

then you could make this code here:

```
    const oldSourceType = this.sourceType;
    this.sourceType = startNode.type;
    // If we has a sourceType before, we can see this as a change and not the initial value, so lets reset the value, so we dont carry over content-types to the new source type.
    if(oldSourceType && oldSourceType !== this.sourceType) {
        this.value = [];
    }
```

nielslyngsoe:
Hi @leekelleher Cool to get to review your PR :-)
I had a look at your code and made a few comments here and there. I also tested it ad it works :-)

leekelleher:
@nielslyngsoe Thank you for feedback, all good insights, and I appreciate the review. I'll update the code accordingly when I'm back after New Year. 👍

src/packages/core/property-editor/uis/tree-picker/config/filter/property-editor-ui-tree-picker-filter.element.ts
leekelleher: @nielslyngsoe Thanks for the tip, I'll try to implement it that way.  Do you know of any existing examples in the codebase? as I haven't been able to find any (yet).

With my initial attempt of observing the entire data-type workspace, I'd copied similar code from the `umb-multiple-color-picker-input`, e.g. Approved Color editor's "Include labels?" toggle.
https://github.com/umbraco/Umbraco.CMS.Backoffice/blob/28f8de57a5833a9a15ded74816d6314ac92be20e/src/packages/core/components/multiple-color-picker-input/multiple-color-picker-input.element.ts#L116-L122

Once I have a working example of observing the `workspace.propertyValueByAlias()`, then we can also update the Approved Color editor configuration accordingly.

src/packages/core/property-editor/uis/tree-picker/config/filter/property-editor-ui-tree-picker-filter.element.ts
leekelleher: @nielslyngsoe Understood. 👍  This helps to answer my previous question (above), https://github.com/umbraco/Umbraco.CMS.Backoffice/pull/1083#discussion_r1439350482.

src/packages/core/property-editor/uis/tree-picker/config/filter/property-editor-ui-tree-picker-filter.element.ts
leekelleher: @nielslyngsoe I agree, I wasn't completely happy with naming this as `-filter`.

> Note, in the codebase `tree-picker` is hyphenated.

I'm wondering if the `-start-node` should become `-source`, _(I'm unsure about also adding `-picker` here too? 🤔),_ so `umb-property-editor-ui-tree-picker-source`, and then the `-filter` could become `umb-property-editor-ui-tree-picker-source-type-picker`.

Feels quite verbose, but yes "naming things is hard" 😅

src/packages/core/property-editor/uis/tree-picker/config/filter/property-editor-ui-tree-picker-filter.element.ts
leekelleher: @nielslyngsoe I've figured out the code, see commit 4f8739294c8f9e6aa2beafdd2336e8b434ba323b for the snippet.

src/packages/core/property-editor/uis/tree-picker/config/filter/property-editor-ui-tree-picker-filter.element.ts
leekelleher: 👍 I've updated the property.

src/packages/core/property-editor/uis/tree-picker/config/filter/property-editor-ui-tree-picker-filter.element.ts
leekelleher: I've added this in now. 👍

src/packages/core/property-editor/uis/tree-picker/config/filter/property-editor-ui-tree-picker-filter.element.ts
leekelleher: I've renamed `umb-property-editor-ui-tree-picker-start-node` to `umb-property-editor-ui-tree-picker-source-picker`;
and renamed `umb-property-editor-ui-tree-picker-filter` to `umb-property-editor-ui-tree-picker-source-type-picker`.

I've also aligned the relevant class/type names.

iOvergaard (approved):

PR #1091, opened 2024-01-03
Selection manager: fixes initial value

leekelleher:
Whilst working on the Tree Picker editor, I discovered an issue where the selected values always had an `undefined` as the first item in the array. I've pinpointed it down to how the Selection Manager is setting the initial selection/value.

When the modal initially opens, the `#multiple` is `false`, (regardless of how the `multiple` property/attribute is configured), so the initial value, e.g. `value[0]` gets set to `[undefined]`. Then when the `#multiple` is observed as `true`, _(there must be an underlying bug with the modal context code somewhere),_ then the `#selection` array would already have an initial value of `[undefined]` so any selected items would be appended that array.

This patch will check if the initial `value` (in `setSelection()`) has any items, if so, it'll use the first value, otherwise it'll be an empty array.

### DIFF
--- a/src/shared/utils/selection-manager/selection.manager.ts
+++ b/src/shared/utils/selection-manager/selection.manager.ts
@@ -57,7 +57,7 @@ export class UmbSelectionManager extends UmbBaseController {
 	public setSelection(value: Array<string | null>) {
 		if (this.getSelectable() === false) return;
 		if (value === undefined) throw new Error('Value cannot be undefined');
-		const newSelection = this.getMultiple() ? value : value.length > 0 ? [value[0]] : value;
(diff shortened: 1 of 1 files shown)


leekelleher:
Set the PR as draft, as I've encountered an underlying issue with the `#multiple` property. I'll investigate further and update this PR accordingly.

leekelleher:
I figured out the underlying issue with the `#multiple` value; I've submitted a patch in PR #1098.

I've updated this PR as the Selection Manager could still encounter an issue with attempting to take the first item from an empty array. My refactor also removes the double ternary operator, uses `.slice(0, 1)` instead.

iOvergaard (approved):

PR #1098, opened 2024-01-04
Tree Element: Call Selection Manager methods in `connectedCallback()`

leekelleher:
Whilst investigating #1091, I found that the `#multiple` property was always `false` on initial load, regardless whether it was `true`.

Following the logic through, the call to the SelectionManager's `setSelection()` is being made from the Tree Element's `selection`  property setter method. The fault in the logic is that this method relies on the `#multiple` property being set, which at this point it isn't - so would always be `false`.

By moving the SelectionManager method calls to the `connectedCallback`, it ensures that all of the attribute/property values have been set.

If there's a better approach to doing this (in terms of Lit/TypeScript), I'm happy to learn and refactor this PR.

### DIFF
--- a/src/packages/core/tree/tree.element.ts
+++ b/src/packages/core/tree/tree.element.ts
@@ -17,29 +17,32 @@ export class UmbTreeElement extends UmbLitElement {
 		this.#treeContext.setTreeAlias(newVal);
 	}
 
+	private _selectable: boolean = false;
 	@property({ type: Boolean, reflect: true })
 	get selectable() {
 		return this.#treeContext.selection.getSelectable();
 	}
 	set selectable(newVal) {
-		this.#treeContext.selection.setSelectable(newVal);
+		this._selectable = newVal;
 	}
 
+	private _selection: Array<string | null> = [];
 	@property({ type: Array })
 	get selection() {
 		return this.#treeContext.selection.getSelection();
 	}
 	set selection(newVal) {
 		if (!Array.isArray(newVal)) return;
-		this.#treeContext?.selection.setSelection(newVal);
+		this._selection = newVal;
 	}
 
+	private _multiple: boolean = false;
 	@property({ type: Boolean, reflect: true })
 	get multiple() {
 		return this.#treeContext.selection.getMultiple();
 	}
 	set multiple(newVal) {
-		this.#treeContext.selection.setMultiple(newVal);
+		this._multiple = newVal;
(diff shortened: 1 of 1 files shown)


iOvergaard (commented):
This seems to work because you are now waiting until the connectedCallback lifecycle to set the values. However, if one of the props should change later on, it no longer will get cascaded to the `#treeContext` since you are now only invoking the setters in connectedCallback. Perhaps we need to retain the current logic as well as invoke the setters in connectedCallback OR figure out why it works when you wait until connectedCallback to set the values?

leekelleher:
@iOvergaard

> if one of the props should change later on, it no longer will get cascaded to the `#treeContext` since you are now only invoking the setters in connectedCallback.

Good point. I hadn't considered the reactivity of those specific properties. Will review.

> OR figure out why it works when you wait until `connectedCallback` to set the values?

It appears to be the order that the attributes/properties are being set, and their setter code is assuming the other attributes/properties have been set and available.

1. In the `tree-picker-modal.element`, the `selection` attribute is set: https://github.com/umbraco/Umbraco.CMS.Backoffice/blob/main/src/packages/core/modal/common/tree-picker/tree-picker-modal.element.ts#L47
2. This makes a call to the `tree.element`'s `selection` setter method: https://github.com/umbraco/Umbraco.CMS.Backoffice/blob/main/src/packages/core/tree/tree.element.ts#L34
3. In turn makes a call to the `selection.manager`'s `setSelection` method, but that method assumes that `getMultiple()` has already been configured, _(which it hasn't)_: https://github.com/umbraco/Umbraco.CMS.Backoffice/blob/main/src/shared/utils/selection-manager/selection.manager.ts#L60

I don't understand the lifecycle stages of when a web component's attribute values are read, (I assume sequentially), but if we were to move the `multiple` attribute higher up, then it would work as expected. https://github.com/umbraco/Umbraco.CMS.Backoffice/blob/main/src/packages/core/modal/common/tree-picker/tree-picker-modal.element.ts#L51

Which felt like a hacky workaround - so wanted to wait until all the attributes/properties had been set, then configure the Selection Manager.

iOvergaard:
@leekelleher Oh my goodness. So many dependencies. As we just discussed, you should try and make a configuration property instead where all of the dependants can be set simultaneously.

leekelleher:
Closing this pull-request, as it has been superseded by #1107.

PR #1104, opened 2024-01-09
block catalogue modal

loivsen:
To be continued...


src/packages/block/block-list/property-editors/block-list-editor/property-editor-ui-block-list.element.ts
```diff
@@ -93,47 +106,82 @@ export class UmbPropertyEditorUIBlockListElement extends UmbLitElement implement
...
-				(x) => x.contentUdi,
-				(layoutEntry) =>
-					html` <uui-button-inline-create></uui-button-inline-create>
-						<umb-property-editor-ui-block-list-block .layout=${layoutEntry}>
-						</umb-property-editor-ui-block-list-block>`,
+				(x, index) => x.contentUdi + index,
+				(layoutEntry, index) => html`
+					<umb-property-editor-ui-block-list-block .layout=${layoutEntry}> </umb-property-editor-ui-block-list-block>
+					${index !== this._layouts.length - 1 ? html`<uui-button-inline-create></uui-button-inline-create>` : nothing}
+				`,
```
nielslyngsoe, line 147: I'm not sure I understand this change. As we would like to have a Inline-create button above all Blocks. So you can create something at the top.

src/packages/block/block-list/property-editors/block-list-editor/property-editor-ui-block-list.element.ts
```diff
@@ -93,47 +106,82 @@ export class UmbPropertyEditorUIBlockListElement extends UmbLitElement implement
...
 			)}
-			<uui-button id="add-button" look="placeholder" @click=${this.#openBlockCatalogue} label="open">Add</uui-button>`;
+			<uui-button-group>
+				<uui-button
+					id="add-button"
+					look="placeholder"
+					label=${this.localize.term('content_createEmpty')}
+					@click=${() => this.#openBlockCatalogue('createEmpty')}>
```
nielslyngsoe, line 154: Parsing on these magic-strings is a bit fragile, from a code maintenance perspective. Instead we here could send just a single optional boolean argument like 'openClipboard'. like: `this.#openBlockCatalogue()` when opening just for Block Types and `this.#openBlockCatalogue(true)` when going to the Clipboard. It can be this simple, and easily changed for more complex situation if ever needed, as its an internal method. No one else uses this.

Inside #openBlockCatalogue, we then should set the data property openClipboard to true, and then the modal handles what should happen when its set to true. In that way no one else than the modal knows about views etc. Leaving that to the modal, making it simpler to figure out how that modal works once someone gets to in the future. :-)

nielslyngsoe: Ah, I now see that you did use a TypeScript type for it. That of course makes it no so fragile as I though, I sometimes still live in a non TS world. But again since someone might implement the Block Catalogue in their own thing, we still want to make each prop going into it as explicit as possible. So being specific about the achievement and not how it takes place. — The current one I've already seen implemented by multiple other property editors, so thats why I'm extra aware here.

src/packages/block/block-list/property-editors/block-list-editor/property-editor-ui-block-list.element.ts
loivsen: It was just to make it match the current backoffice - I can revert it 😊

src/packages/block/block-list/property-editors/block-list-editor/property-editor-ui-block-list.element.ts
loivsen: I had in mind what if a third or fourth view would come into play, how would we handle it then?

nielslyngsoe (approved):
Looks good

PR #1105, opened 2024-01-09
TinyMCE Server variables replacement + temporary files for drag & drop

iOvergaard:
Move over stuff from the server / old backoffice and find alternatives to certain server variables.

Bonus: Implemented drag & drop support on the media picker plugin


leekelleher (approved):
I've got it running fine with the mock API, all appear to be working fine. I can drag & drop media, it'll do a base64 embed of the image, but there's an API error with uploading the temp file (makes sense, since it's the mock API). Picking from the Media tree, it tries to embed the "nothing.jpg" (but this is because the mock API result doesn't have the `.url` property).

With the Management API, the drag & drop does the base64 embed, and it does upload the temp file (as in it's physically on disk), but then it 404s on reload... but I think this is because of the issue you'd mentioned about absolute URLs; with the Management API being on a different hostname/port.

In terms of this PR, I'm happy.  Once all the other missing pieces are in place, it'll be great! 👍

PR #1107, opened 2024-01-09
Tree Element: Adds selection configuration property

leekelleher:
> **This supersedes pull-request #1098.**

Introduces a `selectionConfiguration` property (and new type `UmbTreeSelectionConfiguration`) to the `<umb-tree>` element.

As discussed on https://github.com/umbraco/Umbraco.CMS.Backoffice/pull/1098#issuecomment-1883001464, this property is a container for the required configuration of a tree-picker selection.

I have attempted to update the other uses of `<umb-tree>` that I could find within the codebase, namely the Multi-URL Picker editor, Import Dictionary modal and the Partial View Picker.

### DIFF
--- a/src/packages/core/tree/tree.element.ts
+++ b/src/packages/core/tree/tree.element.ts
@@ -7,6 +7,12 @@ import { UmbObserverController } from '@umbraco-cms/backoffice/observable-api';
 import './tree-item-default/tree-item.element.js';
 import './tree-item-base/tree-item-base.element.js';
 
+export type UmbTreeSelectionConfiguration = {
+	multiple?: boolean;
+	selectable?: boolean;
+	selection?: Array<string | null>;
+};
+
 @customElement('umb-tree')
 export class UmbTreeElement extends UmbLitElement {
 	@property({ type: String, reflect: true })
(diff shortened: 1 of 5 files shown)


src/packages/core/tree/tree.element.ts
```diff
@@ -17,29 +23,21 @@ export class UmbTreeElement extends UmbLitElement {
...
+	private _selectionConfiguration: UmbTreeSelectionConfiguration = {
+		multiple: false,
+		selectable: true,
+		selection: [],
+	};
 
-	@property({ type: Boolean, reflect: true })
-	get multiple() {
-		return this.#treeContext.selection.getMultiple();
+	@property({ type: Object })
```
iOvergaard, line 32: We have not been too adamant about this, but to eventually support Lit 3, we need to get used to putting the `@property` declarations on the **setter** rather than the getter.

leekelleher: @iOvergaard Thanks, good to know.
Should I move the `@property` down to the setter, or move the setter to be before the getter?
_(My C# mindset thinks getter first.)_

leekelleher: > Does Lit have a recommendation?

The Lit 3 upgrade guide has an example of the setter being first, but it implies that this has always been the case (for previous Lit version), so I guess it's up to our coding style preferences? _No strong preference from me._
https://lit.dev/docs/releases/upgrade/#decorated-getter

leekelleher: Reordered the `@property` setter methods in commit d670111800467c8408b0d6dfc840edac9b66bace.

iOvergaard (approved):

PR #1111, opened 2024-01-11
TinyMCE initial value (patch for PR #1105)

leekelleher:
When creating a new document with an RTE, (against the Management API), the initial value is `undefined` and threw an error. So I have set the `value` to be nullable.

I've also updated the `value` data structure in the story.


iOvergaard (approved):
Great catch!

PR #1112, opened 2024-01-11
Color Picker Property Editor: Sets value

leekelleher:
Whilst working on MNTP, trying to save values on a document, I'd noticed that the Color Picker value would save, but it wouldn't populate the UI on reload.

I've wired up setting the `value` of the `<umb-input-color>`, which in turn passes it to the `<uui-color-swatches>` component.


loivsen (approved):

loivsen:
Looks good

PR #1125, opened 2024-01-15
Tree Picker: Pass in "filter" to picker

leekelleher:
For Multinode Treepicker, the "filter" option is passed in to the corresponding Document/Media/Member picker.

This is then used by the `pickableFilter` option on the Tree modal to filter out the selectable items.

For the Member picker, I've put placeholder code and commented out, (until the Member repository code has been implemented).

### DIFF
--- a/src/packages/documents/documents/components/input-document/input-document.element.ts
+++ b/src/packages/documents/documents/components/input-document/input-document.element.ts
@@ -2,7 +2,7 @@ import { UmbDocumentPickerContext } from './input-document.context.js';
 import { css, html, customElement, property, state, ifDefined, repeat } from '@umbraco-cms/backoffice/external/lit';
 import { FormControlMixin } from '@umbraco-cms/backoffice/external/uui';
 import { UmbLitElement } from '@umbraco-cms/internal/lit-element';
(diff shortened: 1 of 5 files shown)


src/packages/documents/documents/components/input-document/input-document.element.ts
```diff
@@ -63,8 +63,8 @@ export class UmbInputDocumentElement extends FormControlMixin(UmbLitElement) {
 	@property({ type: String })
 	startNodeId?: string;
 
-	@property({ type: String })
-	filter?: string;
+	@property({ type: Array })
+	filter?: string[] | undefined;
```
nielslyngsoe, line 67: I know this is not new in this PR. But there is something about the name here that I'm not fond of.

Could it be named `acceptableContentTypeIds`, to make it more explicit what it does?

Same goes for the others of this PR.

leekelleher: @nielslyngsoe I've replaced `filter` with `allowedContentTypeIds`, in commit 0897acee74fd61a137f29ba602f2257919a572ff.

nielslyngsoe (approved):
cool

PR #1127, opened 2024-01-17
Tree Picker: Item sorting

leekelleher:
Adds sorting feature to the Document/Media/Member Picker selected items.


src/packages/core/repository/repository-items.manager.ts
```diff
@@ -76,10 +87,25 @@ export class UmbRepositoryItemsManager<ItemType> extends UmbBaseController {
...
+					this.#items.next(this.#sortByUniques(data));
+				},
+				'_observeRequestedItems',
+			);
 		}
 	}
 
+	#sortByUniques(data: Array<ItemType>) {
+		const uniques = this.getUniques();
+		return [...data].sort((a, b) => {
```
bjarnef, line 102: Can't we use this here?

```
var uniqueAndSorted = [...new Set(myData)].sort() 
```

https://stackoverflow.com/a/42495707

leekelleher: Thanks @bjarnef.  The intent of the `#sortByUniques` method is to sort the `data` array (of objects) in the order of the unique values. More like this: https://stackoverflow.com/q/13304543/12787

bjarnef: Ahh okay, I haven't checked if `Set()` eventually can sort unique data objects in array:
https://stackoverflow.com/a/43667135

nielslyngsoe (approved):
Cool

PR #1129, opened 2024-01-18
block input groups

loivsen:


src/packages/block/block-grid/property-editors/block-grid-group-configuration/property-editor-ui-block-grid-group-configuration.element.ts
```diff
@@ -0,0 +1,61 @@
...
+	}
+	public set value(value: Array<BlockGridGroupConfigration>) {
+		this._value = value || [];
+	}
+
+	@property({ attribute: false })
+	public set config(config: UmbPropertyEditorConfigCollection | undefined) {}
+
+	#addGroup() {
+		if (Array.isArray(this.value)) {
```
nielslyngsoe, line 28: If you used _value here, then you can trust that its always an array.. Since you initially set it to an array And you as well ensure its an array as part of the value setter.

src/packages/block/block-grid/property-editors/block-grid-type-configuration/property-editor-ui-block-grid-type-configuration.element.ts
```diff
@@ -14,22 +16,60 @@ export class UmbPropertyEditorUIBlockGridTypeConfigurationElement
 	extends UmbLitElement
 	implements UmbPropertyEditorUiElement
 {
+	#context?: UmbPropertyDatasetContext;
+
 	@property({ attribute: false })
-	value: UmbBlockTypeBase[] = [];
+	value: UmbBlockTypeWithGroupKey[] = [];
```
nielslyngsoe, line 22: ideally in the block-grid folder, we make a types.js file that holds a UmbBlockGridType type. This interface would then extends UmbBlockTypeWithGroupKey.

src/packages/block/block-type/components/input-block-type/input-block-type.element.ts
```diff
@@ -88,19 +125,62 @@ export class UmbInputBlockTypeElement<BlockType extends UmbBlockTypeBase = UmbBl
...
+	}
+
 	render() {
-		return html`
-			${this._items ? repeat(this._items, (item) => item.contentElementTypeKey, this.#renderItem) : ''}
-			${this.#renderButton()}
-		`;
+		return html`${repeat(
+			this._mappedGroups,
+			(group) => group.key + group.blocks,
```
nielslyngsoe, line 157: I think + group.blocks would stringify it into "[array Array]". So in this cause its properly better to use the `group.blocks.length`?  but I think it would be better to just base this of the `group.key`. Lets try to make just that work.

src/packages/block/block-type/components/input-block-type/input-block-type.element.ts
```diff
@@ -9,16 +9,22 @@ import '../block-type-card/index.js';
 import { css, html, customElement, property, state, repeat } from '@umbraco-cms/backoffice/external/lit';
 import { UmbLitElement } from '@umbraco-cms/internal/lit-element';
 import { UmbChangeEvent } from '@umbraco-cms/backoffice/event';
+import { BlockGridGroupConfigration } from '@umbraco-cms/backoffice/block';
+import { UUIInputEvent } from '@umbraco-cms/backoffice/external/uui';
+import { UMB_PROPERTY_DATASET_CONTEXT, UmbPropertyDatasetContext } from '@umbraco-cms/backoffice/property';
 
 @customElement('umb-input-block-type')
-export class UmbInputBlockTypeElement<BlockType extends UmbBlockTypeBase = UmbBlockTypeBase> extends UmbLitElement {
+export class UmbInputBlockTypeElement<
```
nielslyngsoe, line 17: I think I would make a new Element called `UmbInputGroupedBlockTypeElement` which then implements UmbInputBlockTypeElement within each group.

This would require the create method to just fire an event and then move the create modal implementation to the group and or other implementation. Lets talk about this

nielslyngsoe (approved):
Very good :-)

PR #1132, opened 2024-01-19
Sign out on server

iOvergaard:
Call server endpoints to sign out the user (clear cookies) on the server after having revoked the tokens.

The process is:

1. The user clicks the "Sign out" button in the user modal
2. Revoke any known locally cached tokens (access_token and refresh_token)
3. Clear the local cache of tokens (StorageBackend)
4. Redirect the user to the signOut endpoint, which will clear the cookies
  1. This request contains a redirect uri back to the client
5. The client now does not have a token and will try to get one
6. result: The user will most likely end up on the login page again unless some external login mechanism took over on the server and gave the user a token based on external parameters


kjac (approved):
💪

Issue #1136, opened 2024-01-21
[BUG]: No way to update value of umb-input-tiny-mce


enkelmedia:
**Describe the bug**
Once the TinyMCE is loaded using `umb-input-tiny-mce` there is no way to update the value of the editor without re-rendering it. 

I've made a simple "repro" here:

```typescript
@state()
html = '<p>init html</p>';

setValue(){
	this.html = '<p>updated value</p>';
}

onChange(e : InputEvent){
	this.html = (e.target as UmbInputTinyMceElement).value?.toString() ?? '';
}

render() {
	return html`
	<button @click=${this.setValue}>Set</button>
	<pre>${this.html}</pre>
	<umb-input-tiny-mce
		@change=${this.onChange2}
		.configuration=${this.#configuration}
		.value=${this.html}></umb-input-tiny-mce>
		
	`;
}
```

In the example above both the editor and "pre" will be loaded with the text "init html", when changing the content of the RTE the html-state variable will be updated and hence also the pre-element. However, clicking on the button till not trigger TinyMCE to change but the pre-element will change to "updated value".

enkelmedia:
@nielslyngsoe Updated to use super and tested, it seems to work =D

enkelmedia:
Fixed by this: https://github.com/umbraco/Umbraco.CMS.Backoffice/pull/1137

PR #1137, opened 2024-01-21
Update RTE content when value changes

enkelmedia:
This addresses the issues described here: https://github.com/umbraco/Umbraco.CMS.Backoffice/issues/1136

## Description
I've added implementations to get/set methods for value that will only trigger an update to `editor` after the initial load and when the content is different from the current content of the editor (to avoid calling setContent() with the same content).

## Types of changes

## How to test?
There is a simple example-element that can be used for testing here: https://github.com/umbraco/Umbraco.CMS.Backoffice/issues/1136 when I worked on the PR I made these changes to the `umb-property-editor-ui-tiny-mce` locally for testing.

## Checklist


src/packages/core/components/input-tiny-mce/input-tiny-mce.element.ts
```diff
@@ -42,6 +42,19 @@ export class UmbInputTinyMceElement extends FormControlMixin(UmbLitElement) {
 		return this._editorElement?.querySelector('iframe') ?? undefined;
 	}
 
+	set value(newValue: FormDataEntryValue | FormData) {
+		this._value = newValue;
```
nielslyngsoe, line 46: Sadly we dont have any way to enforce calling the super method. But the FormControlMixin has a getter and setter for the value that we need to call.

So this should be changed into: `super.value = newValue`

nielslyngsoe: Otherwise really good! 👍

enkelmedia: @nielslyngsoe So you are saying that 

```
this._value = newValue;
```

should be

```
super.value = newValue;
```

I'll try that and update the PR =D

enkelmedia:
@nielslyngsoe Updated to use super and tested, it seems to work =D

src/packages/core/components/input-tiny-mce/input-tiny-mce.element.ts
enkelmedia: Updated with new commit

nielslyngsoe (approved):
Looks good, thanks!

Issue #844, opened 2023-08-25
Ensure clear labels for dialog actions


enkelmedia:
I actually started an issue about this for Belle years ago but it never really got any traction. While re-working the backoffice I think it's time to raise the issue again.

Many of the labels for primary actions inside dialogs do not really make sense. Here is an example:

![dialog primary action label](https://github.com/umbraco/Umbraco.CMS.Backoffice/assets/1782524/124971f0-4575-4c88-929c-21bd8b5bea02)

The user chooses a icon and then primary action is "Submit"? I think that these labels should be properly named, in this case mabe "Choose" or "Pick" would make more sense.

I'm not sure if "Submit" is some kind of default text somewhere but I think that there should not be any defaults here to force developers to user labels that make sense in the context.

nielslyngsoe:
@enkelmedia thanks for your input, I agree. The primary button of a dialog should state the action that it carries out.

I did a style guide for Dialogs, which in some way can be transferred to the needs of these Modals. This guide suggests similar naming as you suggested. In general ensure the button that opened the Modal, the headline of the Modal, and the primary action of the modal use the same terms to underline the alignment between these messages.

If your interested the style guide is here on UUI:
https://uui.umbraco.com/?path=/story/design-style-guide--dialog

I marked this as up for grabs

enkelmedia:
@nielslyngsoe Before I "scale out" the approach here, do you thing I'm doing this right or to you have any feedback? 

https://github.com/umbraco/Umbraco.CMS.Backoffice/pull/1139

enkelmedia:
Summary of feedback from the first PR: https://github.com/umbraco/Umbraco.CMS.Backoffice/pull/1139

> Goal is to align headlines and primary action to emphasise the Primary action and its relation to the modal.
> Guidelines: https://uui.umbraco.com/?path=/story/design-style-guide--dialog
>When approaching these kind of changes going forward we should:
> * Strive to ensure the headline and CTA contains the same word (Add, Update, Choose etc)

enkelmedia:
@nielslyngsoe I've updated the description of this card to outline concrete examples of changes I would like to make.

Based on the feedback on the last PR https://github.com/umbraco/Umbraco.CMS.Backoffice/pull/1139 I understood you as you would like to be 100% consistent between modal headers and the button in the modal.

I think this make is a great ide in 95% of the cases but I do have one scenario where I'm still not sure I agree. 

[...]

PR #1139, opened 2024-01-22
#844 fixes headline and button label for content type edit property

enkelmedia:
Partial PR for this issues: https://github.com/umbraco/Umbraco.CMS.Backoffice/issues/844

## Screenshots
After this change the headline and buttons is "contextual" based on if we're adding or updating:

image

## Checklist

### DIFF
--- a/src/assets/lang/en-us.ts
+++ b/src/assets/lang/en-us.ts
@@ -1522,6 +1522,7 @@ export default {
 		addGroup: 'Add group',
 		inheritedFrom: 'Inherited from',
 		addProperty: 'Add property',
+		editProperty : 'Edit property',
 		requiredLabel: 'Required label',
 		enableListViewHeading: 'Enable list view',
 		enableListViewDescription:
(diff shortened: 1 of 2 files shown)


nielslyngsoe (approved):
Very cool.

I do have a secret goal of aligning headlines and primary action. To emphasise the Primary action and its relation to the modal.

I know this is not directly a Dialog, but they are closely related. So to help guide us all in the same direction I did write this guide on Dialogs:
https://uui.umbraco.com/?path=/story/design-style-guide--dialog

So my point being, I feel it would be nice to use the same word in the headline and the primary action when user is editing a property.
With this change you changed it to "Edit Property" "Update" — which is good, but maybe it would be more clear if it was "Edit Property" "Edit"? How do you feel about that?

I'm approving anyway and getting it merged, but would love to get your input on it and eventually a new PR on that specifically :-)

enkelmedia:
@nielslyngsoe **Sidenote**: I also wanted your confirmation/comment on the `# isNew'-approach. Since even a new property will have a generated id I could not find any "official" way to know if the property is new - but my conclusion was that if the alias is empty it has to be new. Since you've merged I guess that you think that this approach is ok?

PR #1146, opened 2024-01-23
Bugfix: Prevent extension double registration

madsrasmussen:
This PR fixes an issue with the `registerMany` method on the extensionRegistry where it was possible to have multiple manifests in the same array with the same alias.

Now we ensure that each manifest is validated before registering the next.


nielslyngsoe (approved):
BEAUTIFUL

nielslyngsoe (approved):
BEAUTIFUL

Issue #1152, opened 2024-01-24
[BUG]: Monaco-editor auto-resize issue


enkelmedia:
**Describe the bug**
Seems like the code editor has problems following when the window is resized.

![umb-monaco-editor-resize](https://github.com/umbraco/Umbraco.CMS.Backoffice/assets/1782524/f75d493e-96da-4c26-8eb0-dc5380f96069)

**To Reproduce**
See the screenshare

**Expected behavior**
Editor should never be wider than the content area.

enkelmedia:
Fixed by this: https://github.com/umbraco/Umbraco.CMS.Backoffice/pull/1153

PR #1150, opened 2024-01-24
Split-view: fix layout

JesmoDev:
Fixes a layout bug introduced in the last split view layout fix.

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Chore (minor updates related to the tooling or maintenance of the repository, does not impact compiled assets)

## Checklist

- [x] If my change requires a change to the documentation, I have updated the documentation in this pull request.
- [x] I have read the **[CONTRIBUTING](<(https://github.com/umbraco/Umbraco.CMS.Backoffice/blob/main/.github/CONTRIBUTING.md)>)** document.
- [x] I have added tests to cover my changes.

PR #1145, opened 2024-01-23
Bugfix/split view overflow

JesmoDev:
Fixes split view overflow caused by fixed min-width on certain elements. Replaced flex with grid layout to address the issue. Note: Introduces horizontal scroll in cases of large min-width, prioritising improved split view visibility.

before:
Screenshot 2024-01-23 at 13 05 29

after:
Screenshot 2024-01-23 at 13 05 17

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Chore (minor updates related to the tooling or maintenance of the repository, does not impact compiled assets)

## Checklist

- [x] If my change requires a change to the documentation, I have updated the documentation in this pull request.
- [x] I have read the **[CONTRIBUTING](<(https://github.com/umbraco/Umbraco.CMS.Backoffice/blob/main/.github/CONTRIBUTING.md)>)** document.
- [x] I have added tests to cover my changes.

PR #1153, opened 2024-01-24
Adjust property-layout to avoid auto

enkelmedia:
## Description
Fixes this issue https://github.com/umbraco/Umbraco.CMS.Backoffice/issues/1152, turns out that having this:

```css
grid-template-columns: 200px auto;
```

On the `umb-property-layout` made the 2nd column in the grid allowed to grow as big as it wants to. The Monaco editor revealed this by growing and growing. This is what we need:

```css
grid-template-columns: 200px minmax(0,1fr);
```

Here is a screencast before and after:
umb-property-layout-resize

This also make the `umb-document-workspace-split-view` work with `display:flex` as before the PR for this issue https://github.com/umbraco/Umbraco.CMS.Backoffice/pull/1145 causing the UI to only show at 50% when there is no variant being edited.

So I would suggest merging this PR and reverting `umb-document-workspace-split-view`  back to:

```css
#splitViews {
   display: flex;
   width: 100%;
   height: calc(100% - var(--umb-footer-layout-height));
}
```

UPDATE: Turns out there is already a PR to revert 1145: https://github.com/umbraco/Umbraco.CMS.Backoffice/pull/1150

## Types of changes

## Checklist


JesmoDev:
Looks good! Thanks!

JesmoDev (approved):

PR #1160, opened 2024-01-25
Feature: Reload Tree Item Children Entity Action

madsrasmussen:
This PR implements the "Reload children" entity action for all trees. It uses the ActionEventContext to dispatch a request to reload children for a tree item. A tree item can then handle that request and fetch the tree items that are needed. Because of pagination have to make the tree item responsible for requesting the children as it is the only one that knows how many children to reload.

Known issue: We currently rely on a unique to refresh the tree root. A few trees haven't been updated to use unique yet, so nothing will happen when the action is clicked. Ex: Document, Media, Dictionary.


nielslyngsoe (approved):
Cool, interesting new concept.

PR #1162, opened 2024-01-25
Bugfix/too-many-token-requests

iOvergaard:
Make sure we wait for the authorization notifier before making any client redirects to allow the notifier to fetch new tokens.

This change effectively eliminates the race condition that was in place - a race to fetch the tokens before the client tried to log in again.

I have also removed an interface that was not really useful anymore making the maintenance on our part easier and reflecting how we create all other contexts.


src/apps/app/app.element.ts
```diff
@@ -156,17 +164,15 @@ export class UmbAppElement extends UmbLitElement {
...
 
-				// If we are on the installer or upgrade page, redirect to the root
-				// but if not, keep the current path but replace state anyway to initialize the router
-				let currentRoute = location.href;
-				const savedRoute = sessionStorage.getItem('umb:auth:redirect');
-				if (savedRoute) {
-					sessionStorage.removeItem('umb:auth:redirect');
-					currentRoute = savedRoute;
+				// If we are on installer or upgrade page, redirect to the root since we are in the RUN state
+				if (pathname === '/install' || pathname === '/upgrade') {
```
kjac, line 168: Is there any way we can have constants for the install and upgrade path names?

kjac (approved):
Looks like this does the trick 😄 also looks like you've fixed the "redirect back to where you were trying to get to after login" 👍 

I left a small comment for you to consider.

src/apps/app/app.element.ts
iOvergaard: Good idea, although they are only used in this file (where the main router is registered). I would like to find some way to better solidify the fact that all "known routes" should be ignored specifically here, so would be good to have a more programmatic approach to that in the future.

PR #1163, opened 2024-01-25
Feature: Render Template Query Item Samples

loivsen:
Small PR that renders the sample items returned by query builder (works on the server)

Example:
image


iOvergaard (approved):
Nice! ☺️

PR #1164, opened 2024-01-25
Feature: Dynamic Root

leekelleher:
Adds the Dynamic Root feature to the Multinode Treepicker property-editor and data-type configuration.

Includes...

- Dynamic Root repository _(to access Management API)_
  - ref: https://github.com/umbraco/Umbraco-CMS/pull/15505
- New extension types, e.g. `dynamicRootOrigin` and `dynamicRootQueryStep`, for client-side UI
  - Added manifest entries, (copied from v13, which were hardcoded in the template)
- Localizations, copied from v13
- Modals for the Origin and Query Step pickers
- With the MNTP data-type configuration, consolidated the "Pick root node" and "Specify Dynamic Root" buttons
  - Included code to support v13 data-type configuration backwards-compatibility


src/packages/core/property-editor/uis/tree-picker/config/source-type-picker/property-editor-ui-tree-picker-source-type-picker.element.ts
```diff
@@ -9,6 +9,8 @@ import { UmbLitElement } from '@umbraco-cms/internal/lit-element';
 import { UMB_PROPERTY_DATASET_CONTEXT } from '@umbraco-cms/backoffice/property';
 import { UmbPropertyValueChangeEvent } from '@umbraco-cms/backoffice/property-editor';
 
+import '../../../../../../documents/document-types/components/input-document-type/input-document-type.element.js';
```
iOvergaard, line 12: Consider if this could be imported from `@umbraco-cms/backoffice/components`. Perhaps it's missing an export.

src/packages/core/property-editor/uis/tree-picker/property-editor-ui-tree-picker.element.ts
```diff
@@ -36,12 +39,19 @@ export class UmbPropertyEditorUITreePickerElement extends UmbLitElement implemen
 	@state()
 	ignoreUserStartNodes?: boolean;
 
+	#dynamicRoot?: UmbTreePickerSource['dynamicRoot'] | undefined;
```
iOvergaard, line 42: You don't need to specify undefined explicitly as the `?` already implies this.

```suggestion
	#dynamicRoot?: UmbTreePickerSource['dynamicRoot'];
```

src/packages/core/property-editor/uis/tree-picker/property-editor-ui-tree-picker.element.ts
```diff
@@ -36,12 +39,19 @@ export class UmbPropertyEditorUITreePickerElement extends UmbLitElement implemen
...
+
+	#dynamicRootRepository: UmbDynamicRootRepository;
```
iOvergaard, line 44: You don't need to specify the type here if it's assigned in the constructor. I believe you could even new it up here directly.

```suggestion
	#dynamicRootRepository;
```

src/packages/core/property-editor/uis/tree-picker/property-editor-ui-tree-picker.element.ts
```diff
@@ -52,6 +62,37 @@ export class UmbPropertyEditorUITreePickerElement extends UmbLitElement implemen
 		this.ignoreUserStartNodes = config?.getValueByAlias('ignoreUserStartNodes');
 	}
 
+	constructor() {
+		super();
+
+		this.#dynamicRootRepository = new UmbDynamicRootRepository(this);
+
+		this.consumeContext(UMB_WORKSPACE_CONTEXT, (workspaceContext: typeof UMB_WORKSPACE_CONTEXT.TYPE) => {
```
iOvergaard, line 70: This type cast of course guarantees that your build will fail if you don't get the correct type, but let's say we change the output type, you would have to also replace this line. I think it's safe enough to accept the implied type:

```suggestion
		this.consumeContext(UMB_WORKSPACE_CONTEXT, (workspaceContext) => {
```

src/packages/core/property-editor/uis/tree-picker/property-editor-ui-tree-picker.element.ts
```diff
@@ -60,7 +101,7 @@ export class UmbPropertyEditorUITreePickerElement extends UmbLitElement implemen
 	render() {
 		return html`<umb-input-tree
 			.value=${this.value}
-			.type=${this.type}
+			.type=${this.type ?? 'content'}
```
iOvergaard, line 104: Looks like the tree picker already defaults to 'content'. Consider if you really need to specify it here as well.

src/packages/documents/documents/components/input-document/input-document.element.ts
```diff
@@ -148,7 +152,7 @@ export class UmbInputDocumentElement extends FormControlMixin(UmbLitElement) {
 
 	#openPicker() {
 		// TODO: Configure the content picker, with `startNodeId` and `ignoreUserStartNodes` [LK]
-		console.log('_openPicker', [this.startNodeId, this.ignoreUserStartNodes]);
+		console.log('#openPicker', [this.startNodeId, this.ignoreUserStartNodes]);
```
iOvergaard, line 155: Do we need to keep the console log?

```suggestion
```

iOvergaard (changes requested):
Looks very nice. Left a few comments concerning the quality of the code for you! 😎

src/packages/core/property-editor/uis/tree-picker/config/source-type-picker/property-editor-ui-tree-picker-source-type-picker.element.ts
leekelleher: I believe there was another recent change that caused me to do this, but just re-tested it and it works without the `import`, so I've removed it now.

src/packages/documents/documents/components/input-document/input-document.element.ts
leekelleher: I've left the comment in, as it's related to other tasks for wiring up the root (start node) ID and ignore start node features. It's already been removed in another work-in-progress branch.

leekelleher:
@iOvergaard FYI, I've updated this PR with the latest updates (e.g. sorter controller changes) from `main`.

src/packages/core/property-editor/uis/tree-picker/property-editor-ui-tree-picker.element.ts
```diff
@@ -52,6 +62,35 @@ export class UmbPropertyEditorUITreePickerElement extends UmbLitElement implemen
...
+		if (this.startNodeId) return;
+
+		const entityId = this.#workspaceContext?.getEntityId();
+		// TODO: Awaiting the workspace context to have a parent entity ID value. [LK]
+		// e.g. const parentEntityId = this.#workspaceContext?.getParentEntityId();
+		if (entityId && this.#dynamicRoot) {
+			this.#dynamicRootRepository.postDynamicRootQuery(this.#dynamicRoot, entityId).then((result) => {
+				if (result) {
+					this.startNodeId = result[0];
+				}
```
iOvergaard, line 89: I would really like if we could double wrap this check to ensure `result` has a length before accessing the 0-index.

src/packages/documents/documents/components/input-document-picker-root/input-document-picker-root.element.ts
```diff
@@ -1,101 +1,268 @@
...
-	#documentPickerContext = new UmbDocumentPickerContext(this);
+	#dynamicRootOrigin?: { label: string; icon: string; description?: string };
 
-	// TODO: DynamicRoot - once feature implemented, wire up context and picker UI. [LK]
-	#dynamicRootPickerContext = {
-		openPicker: () => {
-			throw new Error('DynamicRoot picker has not been implemented yet.');
-		},
-	};
+	#modalContext?: UmbModalManagerContext;
```
iOvergaard, line 38: Consider inferring the type from the token:

```suggestion
	#modalContext?: typeof UMB_MODAL_MANAGER_CONTEXT.TYPE;
```

src/packages/documents/documents/components/input-document-picker-root/input-document-picker-root.element.ts
```diff
@@ -1,101 +1,268 @@
...
+			getUniqueOfModel: (modelEntry) => {
+				return modelEntry.unique;
+			},
+			identifier: 'Umb.SorterIdentifier.InputDocumentPickerRoot',
+			itemSelector: 'uui-ref-node',
+			containerSelector: '#query-steps',
+			onChange: ({ model }) => {
+				if (this.data && this.data.querySteps) {
+					//const steps = [...this.data.querySteps];
+					const querySteps = model; //.map((index) => steps[parseInt(index)]);
```
iOvergaard, line 87: What should happen with these outcommented lines?

src/packages/documents/documents/components/input-document-picker-root/input-document-picker-root.element.ts
```diff
@@ -1,101 +1,268 @@
...
+		};
+	}
+
+	#updateDynamicRootQuerySteps(querySteps?: Array<UmbTreePickerDynamicRootQueryStep>) {
+		if (!this.data) return;
+
+		if (!this.#sorter) {
+			// NOTE: The sorter controller can only be initialized when the `#query-steps` element is available. [LK]
+			this.#initSorter();
+		}
```
iOvergaard, line 133: Is it possible at all to render an empty #query-steps uui-ref-list before any items are created?

src/packages/dynamic-root/modals/dynamic-root-origin-picker-modal.element.ts
```diff
@@ -0,0 +1,110 @@
...
+			})
+			.then(() => {
+				const selectedItems = this.#documentPickerContext.getSelection();
+				if (selectedItems.length !== 1) return;
+				this.#submit({
+					originAlias,
+					originKey: selectedItems[0],
+				});
+			});
+	}
```
iOvergaard, line 55: Consider implementing an async/await pattern here, which should ease the nested complexity a bit

src/packages/dynamic-root/modals/dynamic-root-query-step-picker-modal.element.ts
```diff
@@ -0,0 +1,106 @@
...
+			})
+			.then(() => {
+				const selectedItems = this.#documentTypePickerContext.getSelection();
+				this.#submit({
+					unique: UmbId.new(),
+					alias: alias,
+					anyOfDocTypeKeys: selectedItems,
+				});
+			});
+	}
```
iOvergaard, line 50: Consider the async/await pattern here as well

src/packages/dynamic-root/repository/dynamic-root.repository.ts
```diff
@@ -0,0 +1,46 @@
...
+				steps: query.querySteps!.map((step) => {
+					return {
+						alias: step.alias!,
+						documentTypeIds: step.anyOfDocTypeKeys!,
+					};
+				}),
+			},
+		};
+
+		const result = (await this.#dataSource.postDynamicRootQuery(model)) as DynamicRootResponseModel;
```
iOvergaard, line 38: Would you clarify the need for a type-cast here, please?

src/packages/dynamic-root/repository/dynamic-root.server.data.ts
```diff
@@ -0,0 +1,36 @@
...
+		const { data, error } = await tryExecuteAndNotify(
+			this.#host,
+			DynamicRootResource.postDynamicRootQuery({ requestBody }),
+		);
+
+		if (!error) {
+			return data;
+		}
+
+		return { error };
```
iOvergaard, line 30: There is something off about this return pattern, you are either returning the data as an object or a nested object with an 'error' property.

Would it be possible to simply return the data if it's there or nothing at all, since tryExecuteAndNotify already handles the error?

src/packages/dynamic-root/repository/dynamic-root.repository.ts
```diff
@@ -0,0 +1,46 @@
...
+
+		return result.roots;
```
iOvergaard, line 40: The output of `postDynamicRootQuery` could be an object containing only an `error` property at the moment and this scenario is seemingly not being handled. Consider what should happen in that case, or consider the comment I made in the data source itself.

iOvergaard (changes requested):
Happy path looks good. Left a few comments regarding quality of code and error handling in one specific place =)

src/packages/documents/documents/components/input-document-picker-root/input-document-picker-root.element.ts
leekelleher: I'll tidy this up. They were left over code from tweaking yesterday's sorter controller updates.

src/packages/dynamic-root/modals/dynamic-root-origin-picker-modal.element.ts
leekelleher: @iOvergaard I wasn't sure how to tackle this. Do we have any similar examples in the codebase?

iOvergaard: Haven't tested it but does something like this work?

```suggestion
	async #openDocumentPicker(originAlias: string): Promise<void> {
		await this.#documentPickerContext
			.openPicker({
				hideTreeRoot: true,
			});
		const selectedItems = this.#documentPickerContext.getSelection();
		if (selectedItems.length !== 1) return;
		this.#submit({
			originAlias,
			originKey: selectedItems[0],
		});
	}
```

leekelleher: @iOvergaard Would it be something like this...?

```
async #openDocumentPicker(originAlias: string) {
	await this.#documentPickerContext.openPicker({ hideTreeRoot: true });

	const selectedItems = this.#documentPickerContext.getSelection();
	if (selectedItems.length !== 1) return;

	this.#submit({
		originAlias,
		originKey: selectedItems[0],
	});
}
```

leekelleher: Super, thanks!

src/packages/core/property-editor/uis/tree-picker/property-editor-ui-tree-picker.element.ts
leekelleher: Good spot, I've added the extra check now.

src/packages/documents/documents/components/input-document-picker-root/input-document-picker-root.element.ts
leekelleher: Done. 👍

src/packages/documents/documents/components/input-document-picker-root/input-document-picker-root.element.ts
leekelleher: I've refactored the markup, so that the initial `<uui-ref-list>` elements are available the DOM.

src/packages/dynamic-root/modals/dynamic-root-origin-picker-modal.element.ts
leekelleher: Implemented! 👍

src/packages/dynamic-root/modals/dynamic-root-query-step-picker-modal.element.ts
leekelleher: Implemented! 👍

src/packages/dynamic-root/repository/dynamic-root.server.data.ts
leekelleher: I had copied the pattern from another repository/data-source (I can't recall which one it was now), it was my first repository, so I didn't scrutinize it too much.  I wasn't aware that `tryExecuteAndNotify` handled the error, (I've looked at that code now, I have a better understanding of it).

I've refactored the repository/data-source code, so it aligns better with the other repositories.

src/packages/dynamic-root/repository/dynamic-root.repository.ts
leekelleher: Resolved with refactoring.

src/packages/dynamic-root/repository/dynamic-root.repository.ts
leekelleher: Resolved with refactoring.

iOvergaard (approved):

PR #1168, opened 2024-01-28
Improvement: Document model mapping + side effects

madsrasmussen:
Update server models + remap code for document, document types, media, media types + any side effects of the new models.


nielslyngsoe (approved):
Yolo

PR #1172, opened 2024-01-30
Cleanup after macros

iOvergaard:
cleanup after macros and add styling to announce they are no longer supported for any migrated rte content


leekelleher (approved):
I've tried it with the mocked data and works as expected. 👍

PR #1179, opened 2024-01-31
OpenAPI should be accessible for consumers

iOvergaard:
Based on a discussion on Discord: https://discord-chats.umbraco.com/t/16356022/solved-v14-using-custom-authenticated-controllers-with-the-n

We would like a way to access the configured OpenAPI object in an official capacity so people can use it on their own end, when they, too, have used the library openapi-typescript-codegen as we are using in the core.


src/shared/auth/auth.context.ts
```diff
@@ -94,6 +97,42 @@ export class UmbAuthContext extends UmbBaseController {
...
+	 *
+	 * @example <caption>Using the default OpenAPI configuration</caption>
+	 * ```js
+	 *  	const defaultOpenApi = authContext.getOpenApiConfiguration();
+	 *  	OpenAPI = { ...OpenAPI, ...openApi };
+	 * ```
+	 * @returns The default OpenAPI configuration
+	 */
+	getOpenApiConfiguration() {
+		return this.#openApi;
```
madsrasmussen, line 133: Do we know what internals from the open API configuration we are exposing here?

madsrasmussen (approved):
Looks good 👍

PR #1180, opened 2024-01-31
Feature/block type create labels

loivsen:
Block type changes

- Few changes to the block catalogue so that it's data is being rendered
- Custom create button label now possible
- Create button label default is now named after the item if there only exists one.
- Setting max-width was already there 🚀


src/packages/block/block/modals/block-catalogue/block-catalogue-modal.element.ts
```diff
@@ -57,14 +82,9 @@ export class UmbBlockCatalogueModalElement extends UmbModalBaseElement<
...
 
-	/*
-	#onClickBlock(contentElementTypeKey: string) {
-		this.modalContext?.updateValue({ key: contentElementTypeKey });
-		this.modalContext?.submit();
+		this.#itemManager.setUniques(this._blocks.map((x) => x.contentElementTypeKey));
 	}
-	*/
 
 	render() {
```
nielslyngsoe, line 89: Create that you tested for broken group keys, could we append a bit of space above such, otherwise really great to make sure that they are shown.

image

src/packages/block/block/modals/block-catalogue/block-catalogue-modal.element.ts
```diff
@@ -57,14 +82,9 @@ export class UmbBlockCatalogueModalElement extends UmbModalBaseElement<
...
 		this._blockGroups = this.data.blockGroups ?? [];
-	}
 
-	/*
-	#onClickBlock(contentElementTypeKey: string) {
-		this.modalContext?.updateValue({ key: contentElementTypeKey });
-		this.modalContext?.submit();
+		this.#itemManager.setUniques(this._blocks.map((x) => x.contentElementTypeKey));
 	}
-	*/
```
nielslyngsoe, line 88: It would be nice to render the headlines to they look like Current BO.

Also notice in current backoffice, if you only see one group, we do not render the group headline. :-)

nielslyngsoe (approved):

PR #1217, opened 2024-02-12
Devops: Validate Package Exports

madsrasmussen:
This PR adds a small script to check if a package.json export matches a file in the build output. 

I have added the script after the build is done, and before the files are copied to the CMS folder.

The script will log if any errors are found during the validation.

Based on this script I found two wrong exports.

Be aware I have removed the `examples` exports. They are not included in the build, so I could not see the reason for them to be part of the pckage.json.


iOvergaard (approved):
Good catch on the missing log-viewer files. I guess we have to remember to run "generate:tsconfig" after changing package.json as well, so I did that and it looks good :)

PR #1218, opened 2024-02-13
Generate umbraco package

### LINKED ISSUE #1178: [IDEA]: Auto-generate import maps
We currently have to maintain all exports from the backoffice in several places:

- package.json
- tsconfig.json
- web-test-runner.config.mjs
- Default.cshtml
- Installer/Index.cshtml

The source of truth is that of `package.json`. We would like if we could take those imports and at least be able to auto-generate the import map on the Razor templates on the backend. That would ensure that only the backoffice repository needs to maintain the list.

If possible, we could also generate the maps inside `tsconfig.json` and `web-test-runner.config.mjs` based on package.json, but that is secondary to getting the importmaps generated automatically.

iOvergaard:
Generate an umbraco-package.json file for the backoffice containing data useful to serve the backoffice such as the importmap.

There is now also a new definition added to the UmbracoPackage type to allow to define an importmap.

Fixes #1178

Note: When this is merged, we can also merge https://github.com/umbraco/Umbraco-CMS/pull/15710 to generate the importmap dynamically on the server.

### DIFF
--- a/devops/build/create-umbraco-package.js
+++ b/devops/build/create-umbraco-package.js
@@ -1,0 +1,23 @@
+import { createImportMap } from "../importmap/index.js";
+import { writeFileSync, rmSync } from "fs";
+import { packageJsonName, packageJsonVersion } from "../package/index.js";
+
+const srcDir = './dist-cms';
+const outputModuleList = `${srcDir}/umbraco-package.json`;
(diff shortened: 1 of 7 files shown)


iOvergaard:
I have to merge this before it's approved unfortunately, as the corresponding PR on the server was merged too quickly :)

PR #1230, opened 2024-02-15
Media section without collections

JesmoDev:
Adds most of the media section but leaves out collections as that is being worked on in another PR.


src/packages/core/components/input-upload-field/input-upload-field.element.ts
```diff
@@ -146,38 +208,76 @@ export class UmbInputUploadFieldElement extends FormControlMixin(UmbLitElement)
...
+		if (['mp3', 'weba', 'oga', 'opus'].includes(extension)) return 'audio';
+		if (['mp4', 'mov', 'webm', 'ogv'].includes(extension)) return 'video';
+		if (['jpg', 'jpeg', 'png', 'gif'].includes(extension)) return 'image';
+
+		return 'file';
 	}
 
 	#renderButtonRemove() {
-		if (!this._currentFiles.length && !this._filePaths.length) return;
+		if (!this._files.length && !this._files.length) return;
```
iOvergaard, line 272: It looks like you are checking for the same thing twice - consider changing it to something like this:

```suggestion
		if (!this._files.length) return;
```

src/packages/core/components/input-upload-field/input-upload-field.element.ts
```diff
@@ -146,38 +208,76 @@ export class UmbInputUploadFieldElement extends FormControlMixin(UmbLitElement)
...
 		return html`<uui-button compact @click=${this.#handleRemove} label="Remove files">
-			<uui-icon name="icon-trash"></uui-icon> Remove file(s)
+			<uui-icon name="icon-trash"></uui-icon> Remove file ${this._files.length > 1 ? 's' : ''}
```
iOvergaard, line 274: We need to use the proper localization variable here. This is a good case to test out the new localization value factory, where the localization key is a function that returns a text based on the input.

src/packages/media/media-types/index.ts
```diff
@@ -1,10 +1,12 @@
 import './components/index.js';
 
+export type { UmbAllowedMediaTypeModel } from './repository/index.js';
 export {
```
iOvergaard, line 4: Consider if we should just export everything from the repository/index.js file instead.

src/packages/media/media/components/input-image-cropper/input-image-cropper.element.ts
```diff
@@ -1,167 +1,112 @@
...
+		return this.#renderDropzone();
 	}
 
-	render() {
+	#renderDropzone() {
 		return html`
-			<div id="main">${this.#renderMain()}</div>
-			<div id="side">${this.#renderSide()}</div>
+			<uui-file-dropzone id="dropzone" label="dropzone" @change="${this.#onUpload}">
+				<uui-button label="upload" @click="${this.#onBrowse}">Upload file here</uui-button>
```
iOvergaard, line 90: Localization missing here

src/packages/media/media/components/input-image-cropper/input-image-cropper.element.ts
```diff
@@ -1,167 +1,112 @@
...
-				html` <umb-image-cropper-preview
-					@click=${() => this.#onCropClick(crop)}
-					.crop=${crop}
-					.focalPoint=${this.focalPoint}
-					.src=${this.src}></umb-image-cropper-preview>`,
-		);
+	#renderImageCropper() {
+		return html`<umb-image-cropper-field .value=${this.value} .file=${this.file as File} @change=${this.#onChange}>
+			<uui-button slot="actions" @click=${this.#onRemove} label="Remove files">
+				<uui-icon name="icon-trash"></uui-icon> Remove image
```
iOvergaard, line 104: Localization missing here

src/packages/media/media/entity-actions/create/media-create-options-modal.element.ts
```diff
@@ -0,0 +1,92 @@
...
+			this._allowedMediaTypes = data.items;
+		}
+	}
+
+	async #retrieveHeadline(unique: string) {
+		if (!unique) return;
+		const { data } = await this.#mediaItemRepository.requestItems([unique]);
+		if (data) {
+			// TODO: we need to get the correct variant context here
+			this._headline = `Create at ${data[0].variants?.[0].name}`;
```
iOvergaard, line 50: Localization missing here

src/packages/media/media/entity-actions/create/media-create-options-modal.element.ts
```diff
@@ -0,0 +1,92 @@
...
+	// close the modal when navigating to data type
+	#onNavigate() {
+		this._submitModal();
+	}
+
+	render() {
+		return html`
+			<umb-body-layout headline=${this._headline ?? ''}>
+				<uui-box>
+					${this._allowedMediaTypes.length === 0 ? html`<p>No allowed types</p>` : nothing}
```
iOvergaard, line 63: Localization missing here

iOvergaard (changes requested):
Left a few comments and in general some localization is missing, and I didn't comment on all of those, so please look through the files and add localization keys :)

src/packages/core/components/input-upload-field/input-upload-field.element.ts
```diff
@@ -203,7 +203,9 @@ export class UmbInputUploadFieldElement extends FormControlMixin(UmbLitElement)
 				@change="${this.#onUpload}"
 				accept="${ifDefined(this.extensions?.join(', '))}"
 				?multiple="${this.multiple}">
-				<uui-button label="upload" @click="${this.#handleBrowse}">Upload file here</uui-button>
+				<uui-button label="upload" @click="${this.#handleBrowse}"
+					>${this.localize.term('media_clickToUpload')}</uui-button
+				>
```
iOvergaard, line 208: Please set this as the label rather than the default slot

```suggestion
				<uui-button 
				    label=${this.localize.term('media_clickToUpload')}
				    @click="${this.#handleBrowse}"></uui-button>
```

iOvergaard (approved):

PR #1236, opened 2024-02-19
remove background from uui-table-head in umb-table

nathanwoulfe:
Background props on the uui-table-head element in umb-table borked the external border radius - safe to remove the background styling as the parent umb-table has a background color, which respects the border radius


iOvergaard (approved):
LGTM

PR #1237, opened 2024-02-19
Bugfix: Remove parent field from detail model

madsrasmussen:
> [!IMPORTANT]  
> Merges together with: https://github.com/umbraco/Umbraco-CMS/pull/15819

In this PR we handle the removed `parent` field from the management API detail models. 

Instead of including the parent in the scaffold, it is now passed directly to the create method in a `detail` workspace. 

I have added a `parent` object on a workspace that holds the parent `unique` and parent `entityType`. We use this to reload the correct tree item when a new item is created. It will also be used in the future as the base for a breadcrumb.

For us to be able to both store the unique and entityType on a workspace I had to include the entityType as part of the create URL for a workspace.


nielslyngsoe (approved):
Looks good :-)

PR #1247, opened 2024-02-20
Save & Save & Publish modals, part 1

iOvergaard:
This adds modals for the Save & Save and Publish document actions that at the moment support sending all selected variants to the publishing endpoint. It will also show the status of each variant inside the modal, and although we do not yet have it wired up with the actual languages, the modal handles mandatory variants ([check Storybook for an example](https://ambitious-stone-0033b3603-1247.westeurope.1.azurestaticapps.net/?path=/docs/umb-document-variant-picker-modal--docs)).

> [!CAUTION]
> Please merge at the same time with this PR on the server, since the endpoint has been revamped: https://github.com/umbraco/Umbraco-CMS/pull/15680

https://github.com/umbraco/Umbraco.CMS.Backoffice/assets/752371/9a2f02f7-cc72-40a6-a269-b92c4788694d

To do yet: Publish/Unpublish entity actions. Unpublish button. Scheduled publishing.

To do also: I will create a common context to handle the whole "select variants" and subsequently the conversion of them into `UmbVariantId` process to be used for all the other entity and bulk actions that also need to be able to go through this process, so we don't sit around with duplicated business logic on filtering variants.


leekelleher (approved):
All looks good to me. 👍  Tested out against the mock API.

PR #1253, opened 2024-02-20
Feature: Media Section View

leekelleher:
- Updates the Media Section manifest
- Adds a Media Section View
- Removes the Media "dashboardCollection"

I'd appreciate a sanity check on these changes from @madsrasmussen and @nielslyngsoe, to make sure it is on the right track. Personally, I'm not 100% happy with the "media-management" section path.


leekelleher:
Followed up with @madsrasmussen, the use of "media-management" pathname for the section is acceptable (for the time being). We can review at a later stage.

iOvergaard (approved):

PR #1257, opened 2024-02-21
Save & Save & Publish modals, part 2 (entity actions)

iOvergaard:
- Adds a variant manager context to handle the logic of deciding if we are picking a variant either by opening a modal or selecting the primary variant
- Wires up the entity actions with the new context
- Re-wires the document workspace to use the same context


leekelleher (approved):
Tested the updates (against the mock API), all looking good! 👍

PR #1263, opened 2024-02-21
Feature: Media Collection: Create Action

leekelleher:
Adds the "Create" action/button for the Media Collection UI, with

### DIFF
--- a/src/packages/media/media/collection/action/manifests.ts
+++ b/src/packages/media/media/collection/action/manifests.ts
@@ -1,0 +1,22 @@
+import { UMB_COLLECTION_ALIAS_CONDITION } from '@umbraco-cms/backoffice/collection';
+import type { ManifestCollectionAction } from '@umbraco-cms/backoffice/extension-registry';
+
+export const createManifest: ManifestCollectionAction = {
+	type: 'collectionAction',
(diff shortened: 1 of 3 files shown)


src/packages/media/media/collection/action/manifests.ts
```diff
@@ -0,0 +1,22 @@
+import { UMB_COLLECTION_ALIAS_CONDITION } from '@umbraco-cms/backoffice/collection';
+import type { ManifestCollectionAction } from '@umbraco-cms/backoffice/extension-registry';
+
+export const createManifest: ManifestCollectionAction = {
+	type: 'collectionAction',
+	kind: 'button',
```
madsrasmussen, line 6: Is there any difference between how the media create button and the document create button works? It may be possible to make a new kind we can reuse for both buttons so we have less code to maintain.

leekelleher: @madsrasmussen Only differences are in which repository is called to get the content-type structure, and how the create route/paths are used. I'm sure they could be combined, I'd need some guidance on how to abstract them, but yes, good suggestion. 👍

iOvergaard (approved):

PR #1265, opened 2024-02-21
Extensions Workspace: Added filter toolbar

leekelleher:
@madsrasmussen, I had a need to check if specific manifests were being registered, so found myself doing a lot of `CTRL+F` searches through the pagination, so with my recent work on the Document/Media Collections, I've added a filter toolbar on the Extensions workspace.  Feel free to decline this you think it's not needed.

Notes: I should really get the `umbExtensionsRegistry.extensions` from the repository, but I wasn't sure how to best extend it, (and thought a PR would be a better point to have a conversation about it).


madsrasmussen:
Great! I think it is okay in this case. At least its a great improvement when using the workspace.

I think we should look into adding a "collection toolbar app" new extension point. Then we can add all these filters without a custom element + context.

madsrasmussen (approved):

PR #1274, opened 2024-02-22
Show new variants in the variant selector

iOvergaard:
## Description

The server will never give us the uncreated variants, so we have to augment the variant data received from the server with the language repository. 

There are some fine tunings to be made still:
- We currently "guess" if the document is allowed to be varied (we need to pull this from the document type, I guess)
- Invariant document types are not handled well, they are still showing a variant selector

I feel a bit like a fish out of the waters here, but this at least gives us the possibility to create those variants on the server, which then makes it way easier to test the Save and Save&Publish modals :)

## Types of changes

## How to test?

1. Create more than one language
2. Create a document type and allow it to be varied
3. Test that you can create new variants

image


src/packages/documents/documents/repository/detail/document-detail.server.data-source.ts
```diff
@@ -26,6 +28,7 @@ export class UmbDocumentServerDataSource implements UmbDetailDataSource<UmbDocum
 	 */
 	constructor(host: UmbControllerHost) {
 		this.#host = host;
+		this.#languageRepository = new UmbLanguageCollectionRepository(host);
```
nielslyngsoe, line 31: Also, i'm not sure if we do this in other places, but it feels wrong that a Data-Source initialises another Repository. But this is really just a stomach feeling i got when reading this. :-)

src/packages/core/workspace/components/variant-selector/variant-selector.element.ts
```diff
@@ -232,7 +243,9 @@ export class UmbVariantSelectorElement extends UmbLitElement {
 																? html`<uui-icon class="add-icon" name="icon-add"></uui-icon>`
 																: nothing}
 															<div>
-																${variant.name} <i>(${variant.culture})</i> ${variant.segment}
+																${variant.name ||
+																this._languages.getValue().find((x) => x.unique === variant.culture)?.name}
+																<i>(${variant.culture})</i> ${variant.segment}
 																<div class="variant-selector-state">${variant.state}</div>
 															</div>
 														</button>
```
nielslyngsoe, line 251: I would have wanted to put this comment on the static styles prop, but couldnt find out how.

I know this is not part of this PR. but I noticed it, so I'm going to mention it.

The styling of this component sets the font-family it self. It should not do that instead it should import and use the UmbTextStyle css :-)

src/packages/documents/documents/entity-actions/publish.action.ts
```diff
@@ -1,18 +1,19 @@
...
 
-export class UmbPublishDocumentEntityAction extends UmbEntityActionBase<UmbDocumentPublishingRepository> {
+export class UmbPublishDocumentEntityAction extends UmbEntityActionBase<UmbDocumentDetailRepository> {
 	async execute() {
+		if (!this.repository) throw new Error('Document repository not set');
+
 		const languageRepository = new UmbLanguageCollectionRepository(this._host);
 		const { data: languageData } = await languageRepository.requestCollection({});
 
 		// TODO: Not sure we need to use the Detail Repository for this, we might do just fine with the tree item model it self.
```
nielslyngsoe, line 15: Then you can remove this comment

src/packages/documents/documents/entity-actions/unpublish.action.ts
```diff
@@ -1,18 +1,19 @@
...
 
-export class UmbUnpublishDocumentEntityAction extends UmbEntityActionBase<UmbDocumentPublishingRepository> {
+export class UmbUnpublishDocumentEntityAction extends UmbEntityActionBase<UmbDocumentDetailRepository> {
 	async execute() {
+		if (!this.repository) throw new Error('Document repository not set');
+
 		const languageRepository = new UmbLanguageCollectionRepository(this._host);
 		const { data: languageData } = await languageRepository.requestCollection({});
 
 		// TODO: Not sure we need to use the Detail Repository for this, we might do just fine with the tree item model it self.
```
nielslyngsoe, line 15: And this one :-)

nielslyngsoe (approved):
looks good

PR #1275, opened 2024-02-22
Media section view collection amends

leekelleher:
Wired up the Media section view with the Collection data-type configuration.

Also added support for the `sortOrder` value in the Media Collection Views.

### DIFF
--- a/src/packages/media/media/section-view/media-section-view.element.ts
+++ b/src/packages/media/media/section-view/media-section-view.element.ts
@@ -1,22 +1,21 @@
 import { UMB_MEDIA_COLLECTION_ALIAS } from '../collection/index.js';
-import type { UmbCollectionBulkActionPermissions, UmbCollectionConfiguration } from '../../../core/collection/types.js';
+import { UmbMediaCollectionRepository } from '../collection/repository/index.js';
 import { css, html, customElement, state } from '@umbraco-cms/backoffice/external/lit';
 import { UmbLitElement } from '@umbraco-cms/backoffice/lit-element';
 import { UmbCollectionElement } from '@umbraco-cms/backoffice/collection';
 import { UmbDataTypeDetailRepository } from '@umbraco-cms/backoffice/data-type';
+import type {
+	UmbCollectionBulkActionPermissions,
+	UmbCollectionConfiguration,
+} from '@umbraco-cms/backoffice/collection';
 import type { UmbDataTypeDetailModel } from '@umbraco-cms/backoffice/data-type';
 import type { UmbRoute } from '@umbraco-cms/backoffice/router';
 import { UmbPropertyEditorConfigCollection } from '@umbraco-cms/backoffice/property-editor';
 
 @customElement('umb-media-section-view')
 export class UmbMediaSectionViewElement extends UmbLitElement {
-	// TODO: [LK] Check if we can get the default data-type ID (for the Media ListView) from the server.
-	//private readonly defaultDataTypeId: string = '3a0156c4-3b8c-4803-bdc1-6871faa83fff';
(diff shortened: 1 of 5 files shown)


src/packages/media/media/section-view/media-section-view.element.ts
```diff
@@ -1,48 +1,79 @@
 import { UMB_MEDIA_COLLECTION_ALIAS } from '../collection/index.js';
-import { css, html, customElement } from '@umbraco-cms/backoffice/external/lit';
+import type { UmbCollectionBulkActionPermissions, UmbCollectionConfiguration } from '../../../core/collection/types.js';
```
iOvergaard, line 2: Looks like you are importing from another package here. Consider if something is missing in the index.ts file of that module, so you can import from the alias instead.

src/packages/media/media/section-view/media-section-view.element.ts
```diff
@@ -1,48 +1,79 @@
...
-					dataTypeId: '', //'3a0156c4-3b8c-4803-bdc1-6871faa83fff', //'dt-collectionView',
-					allowedEntityBulkActions: {
-						allowBulkCopy: true,
-						allowBulkDelete: true,
-						allowBulkMove: true,
-						allowBulkPublish: false,
-						allowBulkUnpublish: false,
+	// TODO: [LK] Check if we can get the default data-type ID (for the Media ListView) from the server.
+	//private readonly defaultDataTypeId: string = '3a0156c4-3b8c-4803-bdc1-6871faa83fff';
+	private readonly defaultDataTypeUnique: string = 'dt-collectionView';
```
iOvergaard, line 15: Is this not the alias of our mock collection view? Would this work on the server?

iOvergaard (changes requested):

src/packages/media/media/section-view/media-section-view.element.ts
iOvergaard: Cool 😎

leekelleher: Yes, the old backoffice has the media section (dashboard) hardcoded for data-type ID `-96`.
https://github.com/umbraco/Umbraco-CMS/blob/release-13.1.1/src/Umbraco.Web.UI.Client/src/views/dashboard/dashboard.tabs.controller.js#L145

Which is the ID for the Media ListView data-type.
https://github.com/umbraco/Umbraco-CMS/blob/release-13.1.1/src/Umbraco.Core/Constants-DataTypes.cs#L31

So the question is, do we do the same and hardcode the data-type ID? 🤢 or do we ask the server for it? (but that means more work before the beta/rc) 🤔 

If we decide hardcode the ID for now, is it worth aligning the IDs from the server and mock data?
As on the server it is already hardcoded to `3A0156C4-3B8C-4803-BDC1-6871FAA83FFF`.
https://github.com/umbraco/Umbraco-CMS/blob/release-13.1.1/src/Umbraco.Core/Constants-DataTypes.cs#L115

madsrasmussen: @kjac is aware of this problem. I think he knows the solution.

iOvergaard: Kenn says, as we also discussed on Slack, that you should just request the media collection with `null` as the `dataTypeId` and the server should be able to figure out which view is meant to be shown.

src/packages/media/media/section-view/media-section-view.element.ts
leekelleher: Good spot! I've exported them in PR #1290.

PR #1277, opened 2024-02-22
Feature: Media Collection UI (workspace and property-editor)

leekelleher:
- Adds mock data
- Adds Media workspace condition
- Adds Media workspace view for collection
- Updates existing Collection property-editor (to work with both Documents and Media)

> [!NOTE]
> We'll need to review how `<umb-property-editor-ui-collection-view>` is wired up, as it is currently hardcoded to be either Document or Media. Ideally, it should be abstracted to work with any Entity type that supports ContentType structures.
> To be discussed further...

### DIFF
--- a/src/packages/media/media/workspace/views/collection/media-workspace-view-collection.element.ts
+++ b/src/packages/media/media/workspace/views/collection/media-workspace-view-collection.element.ts
@@ -1,0 +1,83 @@
+import type {
+	UmbCollectionBulkActionPermissions,
+	UmbCollectionConfiguration,
+} from '../../../../../core/collection/types.js';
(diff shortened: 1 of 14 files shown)


src/packages/media/media/workspace/views/collection/media-workspace-view-collection.element.ts
```diff
@@ -0,0 +1,83 @@
+import type {
+	UmbCollectionBulkActionPermissions,
+	UmbCollectionConfiguration,
+} from '../../../../../core/collection/types.js';
```
iOvergaard, line 4: Looks like you import from another module here. Please consider if an export is missing.

iOvergaard (changes requested):

PR #1281, opened 2024-02-23
Chore: Remove empty state element

madsrasmussen:
The Empty State element was only used in 1 place, so I don't think we should ship it as a global component. I have added local styling and removed the empty state element.

We can reintroduce the element when we can implement it in more places and make sure it works as a global component.


nielslyngsoe (approved):
looks good

PR #1283, opened 2024-02-23
Chore: Remove Plop Extension Generator

madsrasmussen:
The extension generator was made as a POC to test if having a generator would make it easier to create new extension files. It hasn't been maintained or improved, so let's get rid of it.


nielslyngsoe (approved):
cool

PR #1285, opened 2024-02-23
Chore: Remove lodash

madsrasmussen:
We only used a few functions from lodash. These can either be swapped with native functions or with custom implementations.

* I have added a toCamelCase and a debounce function to the utils module.
* groupBy is now part of the browser.
* We already have a clamp implementation that is now used everywhere.


nielslyngsoe (approved):
looks good

PR #1290, opened 2024-02-26
Exports types for `UmbCollectionBulkActionPermissions` and `UmbCollectionConfiguration`

leekelleher:
I wasn't sure whether to be specific with the exports or do a generic `export type * from './types.js';`?

---

Once this is merged in, I can mark PRs #1275 and #1277 as ready for re-review.


iOvergaard (approved):
Either it's fine in my opinion. I see you are following the conventions of the other exports of this file, which is great. LGTM.

PR #1295, opened 2024-02-26
Feature: Block Editor Blocks Sorter

loivsen:
## Description
Sorting for blocks in the Block Editor.
Allows to sort block order and change groups

## Types of changes


src/packages/block/block-type/components/input-block-type/input-block-type.element.ts
```diff
@@ -10,17 +11,33 @@ import { UmbLitElement } from '@umbraco-cms/backoffice/lit-element';
...
 @customElement('umb-input-block-type')
 export class UmbInputBlockTypeElement<
-	BlockType extends UmbBlockTypeBaseModel = UmbBlockTypeBaseModel,
+	BlockType extends UmbBlockTypeWithGroupKey = UmbBlockTypeWithGroupKey,
 > extends UmbLitElement {
+	#sorter = new UmbSorterController<BlockType, UmbBlockTypeCardElement>(this, {
+		getUniqueOfElement: (element) => element.contentElementTypeKey!,
+		getUniqueOfModel: (modelEntry) => modelEntry.contentElementTypeKey!,
+		identifier: 'block-editor-blocks-sorter',
+		itemSelector: '.cake',
```
nielslyngsoe, line 24: Ah, is that really the right name for this class.  You could just use `umb-block-type-card` no need for a class just to target it. :-)

nielslyngsoe:
I have turn it into a Draft PR. As the PR is not working, let me know if you need some assistance :-)

loivsen:
As we talked about, ran into some issues getting this to work properly due to how Block Editor syncs the values. Added a TODO. It is not 100% smooth but at least it is now able to sort one item at a time and it does save them in the correct order.

PR #1307, opened 2024-02-28
script for different dev modes

nielslyngsoe:
What it says, super simple, just an easy way to switch without manipulating files.


iOvergaard (approved):
Looks good but it probably only works on a Unix style system :)

PR #1311, opened 2024-02-28
Feature/workspace action menu

iOvergaard:
Adds a workspace action menu that shows extensions of the type `workspace-action-menu-item`.

image

This menu shows up related to workspace-actions, so you have to specify a workspace-action alias to attach it to.

image

Extensions are registered like this:

```js
{
		type: 'workspaceActionMenuItem',
		alias: 'Umb.Document.WorkspaceActionMenuItem.Unpublish',
		name: 'Unpublish',
		weight: 10,
		api: UmbUnpublishDocumentEntityAction,
		meta: {
			workspaceActions: ['Umb.WorkspaceAction.Document.SaveAndPublish'],
			label: 'Unpublish',
			repositoryAlias: 'Umb.Repository.Document.Detail',
			entityTypes: [UMB_DOCUMENT_ENTITY_TYPE],
		},
	}
```

They follow the same syntax as "entity actions" whereas entity actions are registered to be used in the tree, the workspace action menu items have to be bound to a workspace-action to show up. Additionally, they can also be filtered by entity types, so if you have the same button showing up in documents and media, you can specify this to only show up for "document" entities.

The menu items follow much of the same syntax as property actions, however, property actions currently have a TODO to align with entity actions, so I have gone ahead and made these new menu items aligned with entity actions, which we can then bring back to property actions as well.

If you don't specify an element in your registration, we will show a `<umb-entity-action>` element in the menu, which takes care of such nice things as allowing you to supply an API that supports `getHref()` and `execute()` (workspace actions). This all boils down to you being able to use the same API for entity actions, workspace actions, workspace action menu items, and in the future property actions as well aligning all of these "action" types.


src/packages/core/extension-registry/models/workspace-action-menu-item.model.ts
```diff
@@ -0,0 +1,22 @@
...
+
+export interface MetaWorkspaceActionMenuItem extends MetaEntityAction {
+	/**
+	 * Define which workspace actions this menu item should be shown for.
+	 * @examples [
+	 * 	['Umb.WorkspaceAction.Document.Save', 'Umb.WorkspaceAction.Document.SaveAndPublish']
+	 * ]
+	 * @required
+	 */
+	workspaceActions: string[];
```
nielslyngsoe, line 21: Do you think it makes sense to mention multiple? My initial though was just one :-) but maybe you have a use case to share?

src/packages/core/property-action/shared/property-action-menu/property-action-menu.element.ts
```diff
@@ -10,24 +10,25 @@ import { UmbLitElement } from '@umbraco-cms/backoffice/lit-element';
...
 	}
 
-	private _value?: unknown;
-
-	#propertyEditorUiAlias = '';
-
 	@property()
 	set propertyEditorUiAlias(alias: string) {
+		// If there is an existing initializer, we need to dispose it.
+		this.#actionsInitializer?.destroy();
```
nielslyngsoe, line 30: Actually this should already be happening, since the initialiser always gets the same alias. See line 43

src/packages/core/workspace/components/workspace-action/shared/workspace-action-menu/workspace-action-menu.element.ts
```diff
@@ -0,0 +1,148 @@
...
+			umbExtensionsRegistry,
+			'workspaceActionMenuItem', // TODO: Stop using string for 'workspaceActionMenuItem', we need to start using Const.
+			(action) => {
+				const containsAlias = action.meta.workspaceActions.some((alias) => this.workspaceActions.includes(alias));
+				const isValidEntityType = !action.meta.entityTypes.length || action.meta.entityTypes.includes(entityType);
+				return containsAlias && isValidEntityType;
+			},
+			(ctrls) => {
+				ctrls.forEach((ctrl) => {
+					ctrl.properties = { unique, entityType };
```
nielslyngsoe, line 70: I see what you are doing here.

You can 'just' set the properties prop on the this.#actionsInitializer.

Like this:
`this.#actionsInitializer.properties = { unique, entityType };`

then it will be parsed to all extension controllers, same effect, but there is a feature for it already. :-)

src/packages/core/workspace/components/workspace-action/shared/workspace-action/workspace-action.element.ts
```diff
@@ -1,34 +1,53 @@
...
-		if (oldValue !== this._manifest) {
+		const oldValue = this.#manifest;
+		this.#manifest = value;
+		if (oldValue !== this.#manifest) {
 			this.#createApi();
 			this.requestUpdate('manifest', oldValue);
 		}
 	}
 
+	@state()
```
nielslyngsoe, line 32: This can work for now, but for this to be as flexible as the overwrite feature, it should support overwrites of overwrites of overwrites.  And thats a bit complex, so for that to work, we should gather the aliases as part of the overwrite feature of the Extension Initialisers.

Another thing is, that this should not be calculated as a getter method on a state. Cause that will be triggered everytime there is a render for this element.

Instead do this as part of the setter of the manifest. As the Manifest is what determins the outcome of the calculation. Then thats can set the `_aliases` state to an array of strings. in this way it will just read te array when re-rendered and only do the calculation when the manifest is set. :-)

nielslyngsoe (changes requested):
Cool, just as I expected. Works and is well implemented, I have made a few smaller comments to be corrected and then its good to go :-)

src/packages/core/extension-registry/models/workspace-action-menu-item.model.ts
iOvergaard: I modeled it after property actions, which take an array of property editor UI aliases. I think it could make sense for something like the "Deploy" button if they want to add the button on both the documents and media workspace. What do you think?

src/packages/core/workspace/components/workspace-action/shared/workspace-action-menu/workspace-action-menu.element.ts
iOvergaard: I figured they would be set before initialization this way, but not sure it makes a difference, so I'm going to change it :)

src/packages/core/workspace/components/workspace-action/shared/workspace-action/workspace-action.element.ts
iOvergaard: Cool, thanks. Will do that.

nielslyngsoe (approved):
Very nice :-)

PR #1312, opened 2024-02-28
Bugfix: Create user + get initial password

madsrasmussen:
Server PR: https://github.com/umbraco/Umbraco-CMS/pull/15773

This PR updates our code to handle the new changes in the management API.

Updates:
* Use the location header id to retrieve the user after the creation
* Use the new endpoint to request a server-generated password for the new user

What to test:
* It is possible to create a new user
* You receive an initial password in the success dialog after the user is created.


iOvergaard (approved):
Tested against the server and it works!

PR #1313, opened 2024-02-28
Feature: Tree paging

madsrasmussen:
This PR aligns the tree extension with how we also build collection extensions. It now registers a wrapping initializing element and renders a kind inside that. We can now pass an object of props to the main element and each kind can have its additional props.

The PR also includes pagination for the tree. I have set the default page size on a tree item to be 50. It is the number of tree items that fit inside a large screen. The page size can be adjusted if we think we need more or less.

I have included a video of the functionality as you either have to update the code or create more than 50 items to test the feature:

Styling of the load more button is up-for-grabs 😁 

https://github.com/umbraco/Umbraco.CMS.Backoffice/assets/6078361/fe8b0386-2601-4308-bc16-703f11931c71

> [!IMPORTANT]  
> This PR will break all tree and tree item registrations as they now require a kind as part of the manifest.

Update the manifest to:

```js
const tree: ManifestTree = {
  type: 'tree',
  kind: 'default', // <---- This one
  alias: 'ALIAS',
  name: 'My Tree',
  meta: {
    repositoryAlias: 'REPO_ALIAS',
  },
};
```


nielslyngsoe (approved):
Cool, Mads & I talked about a few optimisations that we will append later.

PR #1315, opened 2024-02-28
Bugfix/document type editor correct tabs sorter

nielslyngsoe:
Adding JSDocs
Removing unnesecary #model
Using disable/enable feature of Sorter
Rename resolveVerticalDirection to resolvePlacement — as that makes more sense.
Correcting the resolvePlacement(resolveVerticalDirection) function.

## Types of changes

## Checklist


src/packages/documents/document-types/workspace/views/design/document-type-workspace-view-edit.element.ts
```diff
@@ -130,9 +125,9 @@ export class UmbDocumentTypeWorkspaceViewEditElement extends UmbLitElement imple
 				(isSorting) => {
 					this._sortModeActive = isSorting;
 					if (isSorting) {
-						this.#sorter.setModel(this._tabs!);
+						this.#sorter.enable();
```
loivsen, line 128: It looks like this new disable/enable feature is not really fully working. I am able to still move things around (groups) with and without going into sort mode.

Additionally, it seems like the UI sometimes breaks as well between sort mode and normal - I think this happens when i start sorting without going into sort mode.

loivsen (changes requested):
Looks good, found a bug 😊

iOvergaard (approved):
We'll fix the aforementioned issue later on. Merging this to allow for a better experience in the beta.

PR #1318, opened 2024-02-29
Bugfix: Health Check Dashboard

loivsen:
Fixes some loading issue on the health check dashboard


src/packages/health-check/dashboard-health-check.element.ts
```diff
@@ -58,7 +59,7 @@ export class UmbDashboardHealthCheckElement extends UmbLitElement {
 				weight: 500,
 				meta: {
 					label: group.name || '',
-					api: () => import('./health-check.context.js'),
+					api: UmbHealthCheckContext,
```
iOvergaard, line 62: Now that you are using the `loadManifestApi()` function, I think it should work if you leave this as a dynamic import, right?

src/packages/health-check/health-check.context.ts
```diff
@@ -43,6 +44,10 @@ export class UmbHealthCheckContext {
 			this._results.next(undefined);
 		}
 	}
+
+	public destroy(): void {
+		this.destroy();
+	}
```
iOvergaard, line 50: Just leave this function empty - currently it invokes itself :)

```suggestion
	public destroy(): void {
	}
```

src/packages/health-check/dashboard-health-check.element.ts
iOvergaard, line 62: In order to stay aligned with other manifests, could we then move the `api` property out of the "meta" object and onto the manifest itself?

iOvergaard (changes requested):
Looks good, left a few comments :)

src/packages/health-check/dashboard-health-check.element.ts
loivsen: Yeah last thing that I forgot to add 🚀

iOvergaard (approved):

PR #1321, opened 2024-02-29
fix: Make document save adjust model to expected from server

nielslyngsoe:
remodel the data thats send of to the server so it only includes the updates that the user accepted via Variant Picker Modal.

Invariant document-types should make the document store its variant information in a invariant variant.
When varying we should not get a invariant variant, though we do have invariant values.

## Types of changes

## Checklist


iOvergaard (approved):
The immediate problem is fixed and we can save again. However, the publish action is not always playing ball: If you click a bit around in the variants, you can quickly get it to send a payload such as `{culture: 'en-US'}, {culture: null}` along to the server, which is invalid and the server complains. I even managed to get it to send `en-US` twice in the same request, which the server also didn't like.

I'm approving this PR since it concerns saving specifically, which seems to work, and it only now allows us to uncover the errors with publication.

PS! The variant selector has its own set of problems, the most prominent being: if you add a second language, then click on that language in the variant selector, then try and go back to your primary language, you'll then see a blank page.

PR #1344, opened 2024-03-02
Breaking: Refactor modals to always retrieve a host

nielslyngsoe:
This will enable context api to, ALWAYS, be able to retrieve contexts from the origin of the modal, aka. where the modal was initiated.
This was a feature before, but you had to specifically parse on a prop called originTarget. This showed to be unnecessary complex, as developers assume it would just work. Plus we do not want to eventually forget such.

Currently i decided to go with a ControllerHost, though the immediate need was just an element. But this will secure clean up of the modal if the Host is destroyed. This will eventually mean if the host gets closed, the modal should also be closed, this might not be what some scenarios like, but I think it will show to be best architectural choice.
Especially in the perspective of the modal potentially utilising a context provided of the modal that then gets closed. So this should make it more clear who should own a modal.

Now that I was around the block al of it also got a big clean up, so hopefully a bit of TSC and runtime performance gained as well.

## Types of changes

## Checklist


madsrasmussen (approved):
Looks good to me

PR #1348, opened 2024-03-02
Feature: Entity Action kinds

madsrasmussen:


nielslyngsoe (approved):
Self approving for the win

PR #1365, opened 2024-03-05
Delete unused create dictionary modal

madsrasmussen:
The create dictionary modal is not in use anymore. We have a create route instead. This PR deletes everything related to that modal.


leekelleher (approved):
All good with the removal.

PR #1370, opened 2024-03-05
Bugfix/create-media

JesmoDev:
You can now create media items. But nested items(folders) are not supported yet


loivsen (approved):
Looks good to me

PR #1375, opened 2024-03-05
Bugfix: Remove option to reload children of file system file

madsrasmussen:
File system files can't be nested. This PR removes the option to reload children of such.


loivsen (approved):
🚀

PR #1380, opened 2024-03-05
Fix members

JesmoDev:
Fix variants in members.
Add missing group picker
Also added split view to align with media and documents


iOvergaard:
If I add a member without a password, it doesn't validate and I get shown this empty page:

image

iOvergaard:
Consider if the member workspace views should be switched around so that I see the "Member" first:

image

iOvergaard (changes requested):
Code looks fine but I left a few comments for consideration.

iOvergaard (approved):
Merging now to have something ready for the next build.

PR #1383, opened 2024-03-05
Add missing media type picker

iOvergaard:
What it says + now using it to configure the MediaPicker which should be able to choose media types and not use the Tree Picker.

> [!WARNING]
> I ended up messing about with the input values coming from the server. @kjac looked into it and said that all types are migrated to comma-separated strings, so I aligned the 4 different property editor UI's here, so we don't need to go back and forth with different value types.


loivsen (changes requested):
Looks like there picked items from input-media-type doesn't render in the UI. Saves correctly though.
Maybe can add a `catch => undefined` to the modal as well while we at it 🥰

iOvergaard:
> Looks like there picked items from input-media-type doesn't render in the UI. Saves correctly though. Maybe can add a `catch => undefined` to the modal as well while we at it 🥰

You are right. I don't think it's related to this, though. Those input types should "just work".

leekelleher:
re: CSV values. I thought it was only the property-editor UIs that needed to align with the server migrated values, and that the internal `input-` components could control their own value structures, e.g. native `string[]`.

If we want to align the value structure, that's fine, just feels that the CSV structure is there to support legacy data.

leekelleher (approved):
Tested it out, works as expected.

Spotted a small missing const import, (for MNTP's Member Type picker UI), which I've updated on here. All good.

loivsen (approved):

PR #1387, opened 2024-03-05
lit directive for setting focus

nielslyngsoe:
IMORTANT; this does not work, we need some work done in UUI before we can make this work properly.

## Types of changes

## Checklist


leekelleher (approved):
Tested out all the places where `umbFocus` was added, all work as expected. 👍 

Side note: there is a slight issue with workspaces that have an intermediary step, e.g. Create Stylesheet, opens modal, select New Stylesheet, and the name input isn't focused. Same with Create Script. But I don't want that to hold up the feature, so can be picked up on later.

PR #1391, opened 2024-03-06
Bugfix: Rename file system file

madsrasmussen:
This PR fixes renaming a file system file: Script, Partial View, and Stylesheet:

How to test:
* Rename a file through the entity action
* Make sure the tree is updated correctly.

> [!CAUTION]
> Breaking changes

**Rename of kind**
Entity Action `rename` kind is made specific for server files and renamed to `renameServerFile`

**Move event class to another module**
`UmbRequestReloadStructureForEntityEvent` moved from `@umbraco-cms/backoffice/event` module to` @umbraco-cms/backoffice/entity-action` module because of circular dependency.


iOvergaard:
When I rename a script, it doesn't autopopulate the "new name" field like it does with stylesheets and I have the same problem here as stylesheets where the URL doesn't change:

image

madsrasmussen:
This is ready for a second review.

* I have fixed the issue where the name input wasn't correctly populated.
* The problem with reloading is a bit trickier to fix. I have made a temporary solution that we need to improve on moving forward. This is the same case as If you delete an item you already have open. I have added an observer on the detail data and if it's removed from the store the workspace button is disabled. It won't prevent you from reloading, but you can't try to save it anymore. Ideally, we will have a workspace state that tells the entity is gone, but I think we should look into a solution for all workspaces at once.

iOvergaard (approved):

PR #1397, opened 2024-03-06
Bugfix: Remove rename of const

madsrasmussen:


nielslyngsoe (approved):
Looks good, I just double checked any references and they are all covered now :-)

PR #1399, opened 2024-03-06
Bugfix/media picker

JesmoDev:
Fixes the image picker property editor


src/packages/media/media/property-editors/media-picker/property-editor-ui-media-picker.element.ts
```diff
@@ -1,20 +1,21 @@
...
 
 /**
  * @element umb-property-editor-ui-media-picker
  */
 @customElement('umb-property-editor-ui-media-picker')
 export class UmbPropertyEditorUIMediaPickerElement extends UmbLitElement implements UmbPropertyEditorUiElement {
-	@property()
-	public value?: string;
+	@property({ attribute: false })
+	public value: Array<{ key: string; mediaKey: string; mediaTypeAlias: string }> = [];
```
iOvergaard, line 16: Please consider how the migrator converts values from V13 to V14 here. I could fear that it converts these values to a CSV format `key1,key2,key3`. Please verify this before changing the type here.

src/packages/media/media/property-editors/media-picker/property-editor-ui-media-picker.element.ts
```diff
@@ -30,21 +31,41 @@ export class UmbPropertyEditorUIMediaPickerElement extends UmbLitElement impleme
...
+
 	@state()
 	private _limitMin: number = 0;
 	@state()
 	private _limitMax: number = Infinity;
 
+	protected firstUpdated(_changedProperties: PropertyValueMap<any> | Map<PropertyKey, unknown>): void {
+		super.firstUpdated(_changedProperties);
+
+		this._items = this.value ? this.value.map((x) => x.mediaKey) : [];
```
iOvergaard, line 45: Consider moving this to a _setter_ instead (`public set value()`) which will ensure it happens every time the property is updated and not just when the component itself is updated.

src/packages/media/media/property-editors/image-cropper/property-editor-ui-image-cropper.element.ts
```diff
@@ -29,21 +30,20 @@ export class UmbPropertyEditorUIImageCropperElement extends UmbLitElement implem
...
-	// 	}
-	// }
+		if (!this.value) {
+			this.value = {
+				src: '',
+				crops: crops,
+				focalPoint: { left: 0.5, top: 0.5 },
+			};
+		} else {
+			this.value.crops = crops;
```
iOvergaard, line 44: Would this not mean that you are overriding any stored crops with the config by doing this? Do crops not also hold references to media items, or are they just simple config objects?

iOvergaard (changes requested):

src/packages/media/media/property-editors/media-picker/property-editor-ui-media-picker.element.ts
```diff
@@ -30,21 +31,41 @@ export class UmbPropertyEditorUIMediaPickerElement extends UmbLitElement impleme
...
+	}
+
 	private _onChange(event: CustomEvent) {
-		this.value = (event.target as UmbInputMediaElement).selectedIds.join(',');
-		this.dispatchEvent(new UmbPropertyValueChangeEvent());
+		const selectedIds = (event.target as UmbInputMediaElement).selectedIds;
+
+		const result = selectedIds.map((mediaKey) => {
+			return {
+				key: UmbId.new(),
```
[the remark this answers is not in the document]
JesmoDev: I was told that the key is just a unique for that value. And the mediaKey is the unique for the actual media file

iOvergaard (approved):

PR #1405, opened 2024-03-11
Route parameters such as `uc:store` are being falsely replaced

iOvergaard:
Add support for having paramater-like segments of a url that shouldn't be replaced + tests


mattbrailsford (commented):
The only issue I can see is if someone creates a route that starts with a parameter ie, `:something/edit` whereby it doesn't start with a '/' character

mattbrailsford (approved):
LGTM 👍

PR #1409, opened 2024-03-12
Feature: LogViewer pagination

loivsen:
## Description

Pagination added for common log messages (10 per page) and saved searches (15 per page). Can be changed

## Types of changes

### DIFF
--- a/src/packages/log-viewer/workspace/views/overview/components/log-viewer-message-templates-overview.element.ts
+++ b/src/packages/log-viewer/workspace/views/overview/components/log-viewer-message-templates-overview.element.ts
@@ -39,7 +39,6 @@ export class UmbLogViewerMessageTemplatesOverviewElement extends UmbLitElement {
 	}
 
 	#getMessageTemplates() {
-		//const take = this._messageTemplates?.length ?? 0;
 		const skip = this.#currentPage * this.#itemsPerPage - this.#itemsPerPage;
 		this.#logViewerContext?.getMessageTemplates(skip, this.#itemsPerPage);
 	}
--- a/src/packages/log-viewer/workspace/views/overview/components/log-viewer-message-templates-overview.element.ts
+++ b/src/packages/log-viewer/workspace/views/overview/components/log-viewer-message-templates-overview.element.ts
@@ -1,25 +1,31 @@
 import type { UmbLogViewerWorkspaceContext } from '../../../logviewer.context.js';
 import { UMB_APP_LOG_VIEWER_CONTEXT } from '../../../logviewer.context.js';
 import { UmbTextStyles } from '@umbraco-cms/backoffice/style';
-import { css, html, customElement, state } from '@umbraco-cms/backoffice/external/lit';
+import { css, html, customElement, state, nothing } from '@umbraco-cms/backoffice/external/lit';
 import { UmbLitElement } from '@umbraco-cms/backoffice/lit-element';
 import type {
-	PagedLogTemplateResponseModel,
+	LogTemplateResponseModel,
 	SavedLogSearchResponseModel,
 } from '@umbraco-cms/backoffice/external/backend-api';
+import type { UUIPaginationEvent } from '@umbraco-cms/backoffice/external/uui';
 
-//TODO: fix pagination bug when API is fixed
 @customElement('umb-log-viewer-message-templates-overview')
 export class UmbLogViewerMessageTemplatesOverviewElement extends UmbLitElement {
+	#itemsPerPage = 10;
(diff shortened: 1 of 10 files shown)


src/packages/log-viewer/workspace/views/overview/components/log-viewer-message-templates-overview.element.ts
```diff
@@ -1,39 +1,52 @@
...
-			this._messageTemplates = templates ?? null;
+			this._messageTemplates = templates?.items ?? [];
+			this._total = templates?.total ?? 0;
 		});
 	}
 
 	#getMessageTemplates() {
-		const take = this._messageTemplates?.items?.length ?? 0;
-		this.#logViewerContext?.getMessageTemplates(0, take + 10);
+		//const take = this._messageTemplates?.length ?? 0;
```
iOvergaard, line 42: I think it's safe to remove this line :)

```suggestion
```

src/packages/log-viewer/workspace/views/overview/components/log-viewer-message-templates-overview.element.ts
```diff
@@ -54,11 +67,11 @@ export class UmbLogViewerMessageTemplatesOverviewElement extends UmbLitElement {
 	render() {
 		return html`
 			<uui-box headline="Common Log Messages" id="saved-searches">
-				<p style="font-style: italic;">Total Unique Message types: ${this._messageTemplates?.total}</p>
+				<p style="font-style: italic;">Total Unique Message types: ${this._total}</p>
```
iOvergaard, line 70: Do we have a localization key for this or is it new?

iOvergaard (changes requested):
Looking good - just found a few things. Having tested the saved searches, I don't believe we can make a pagination for it at the moment since it uses a shared context across the whole logviewer workspace. We are showing the saved searches in a dropdown in the Search view, which messes with the pagination in the Overview view:

image

Honestly, I think you should set the maximum of saved searches to something like 999 and then remove the uui-pagination. It's one of those edge cases where we'd just like to show everything if possible.

iOvergaard:
Also -- are you planning to build pagination for the "Search" view as well? I think that would be the biggest/best addition to the log viewer workspace.

loivsen:
> Also -- are you planning to build pagination for the "Search" view as well? I think that would be the biggest/best addition to the log viewer workspace.

Doesn't it already have pagination? or am i looking at the wrong place?
image

src/packages/log-viewer/workspace/views/overview/components/log-viewer-message-templates-overview.element.ts
loivsen: This message exists in old Backoffice, but it doesn't seem like there is any localization keys for it. There seems to be more things missing for log viewer like "common log messages" and "number of errors", etc.

iOvergaard:
> > Also -- are you planning to build pagination for the "Search" view as well? I think that would be the biggest/best addition to the log viewer workspace.
> 
> Doesn't it already have pagination? or am i looking at the wrong place?

Cool, I must have tested with not enough data. Carry on 🥇

loivsen:
> Looking good - just found a few things. Having tested the saved searches, I don't believe we can make a pagination for it at the moment since it uses a shared context across the whole logviewer workspace. We are showing the saved searches in a dropdown in the Search view, which messes with the pagination in the Overview view:
> 
> image
> Honestly, I think you should set the maximum of saved searches to something like 999 and then remove the uui-pagination. It's one of those edge cases where we'd just like to show everything if possible.

Alternatively we could also just add pagination to the other places like the dropdown in the search.
I changed `take` to 999 for now and letting the pagination logic stay in case we change our mind later down the line. It doesn't render pagination anyway unless the limit is reached (999 now in this case).

iOvergaard:
> > Looking good - just found a few things. Having tested the saved searches, I don't believe we can make a pagination for it at the moment since it uses a shared context across the whole logviewer workspace. We are showing the saved searches in a dropdown in the Search view, which messes with the pagination in the Overview view:
> Alternatively we could also just add pagination to the other places like the dropdown in the search. I changed `take` to 999 for now and letting the pagination logic stay in case we change our mind later down the line. It doesn't render pagination anyway unless the limit is reached (999 now in this case).

Yeah, but I guess we need to split the context a bit, since currently, we have one Observable for `savedSearches` in the context, so that means if you change to page 2 in one place, it will show the page 2 content in the other place, but the pagination won't update automatically.

Good idea to fix it for now with what you just committed 👍

PR #1423, opened 2024-03-14
Feature/individual publication modals

iOvergaard:
This splits the "one and only" save/publish/unpublish modal up into 3 new modals which use a common element for the variant selection part. This also adds a schedule publish modal, which is quite different from the rest. All modals have corresponding stories to view.

Some of the "should we open a modal or not" logic is duplicated with small nuances between the different states. There are presumably some optimizations to be done here to lessen the amount of duplication. There is already some common logic in the modals to handle filtering of variants, however, I had to realize that there are just too many differences in which variants to show, how to show them, what to do when there is only one language, how to map languages to variants (for bulk publish/unpublish), etc., so that it wasn't possible to get a good overview before we created all of the modals.

Storybook: https://ambitious-stone-0033b3603-1423.westeurope.1.azurestaticapps.net/?path=/docs/umb-document-publish-modal--docs

### What to test

**Documents with no variants**
- [ ] Workspace action: Save directly without a modal
- [ ] Workspace action: Save and publish directly (no modal)
- [ ] Workspace menu action: Unpublish... (user should confirm)
- [ ] Workspace menu action: Schedule... (you should be able to select publish and unpublish dates)
- [ ] Entity action: Publish (no modal)
- [ ] Entity action: Unpublish (user should confirm)

**Documents with variants**
- [ ] Workspace Action: Save and choose variants in a modal
- [ ] Workspace Action: Save and publish (choose variants in a modal)
- [ ] Workspace menu action: Unpublish... (choose variants in a modal)
- [ ] Workspace menu action: Schedule... (choose variants and publish/unpublish dates in a modal)
- [ ] Entity action: Publish (select which language to publish in a modal)
- [ ] Entity action: Unpublish (select which language to unpublish in a modal)

**Bulk actions**
- [ ] Select 1 item and publish: Should fall back to the entity action (if 1 language = publish, if more than 1 = show modal)
- [ ] Select 1 item and unpublish: Should fall back to the entity action (if 1 language = unpublish, if more than 1 = show confirm modal)
- [ ] Select 2 items and publish: Should let the user select languages to publish with no variant states shown (we don't know the states of the variant)
- [ ] Select 2 items and unpublish: Should let the user select languages to unpublish with no variant states shown


leekelleher (approved):
Tried out all the test scenarios, all working as expected. 👍

PR #1432, opened 2024-03-15
Feature/unpublish with references

iOvergaard:
### Description

Adds tracked references to the unpublish modal by way of a new table element that can fetch the references.

Parallel to that, the unpublish modal now also verifies tracked references to see if it should show a warning. Additionally, it pulls down the server configuration to see if you are allowed to unpublish based on the `disableUnpublishWhenReferenced` server setting. It does so by asking the new `UmbDocumentConfigurationContext`, which is registered as a global context, so that we may use it for the delete modal as well as other things around the document workspace.

### What to test
- [x] Unpublish a page without references = no table shown, no warning shown, **unpublish succeeds**
- [x] Unpublish a page with references and `disableUnpublishWhenReferenced=false` = table shown, warning shown, **unpublish succeeds**
- [x] Unpublish a page with references and `disableUnpublishWhenReferenced=true` = table shown, warning shown, **unpublish button disabled** and no other way to submit the form

NB! It is still possible to unpublish through bulk actions, as we cannot check every document for this - this was also the case in the old backoffice.

Story for unpublish with tracked references: https://ambitious-stone-0033b3603-1432.westeurope.1.azurestaticapps.net/?path=/story/umb-document-unpublish-modal--with-tracked-references

Story for the tracked references table isolated: https://ambitious-stone-0033b3603-1432.westeurope.1.azurestaticapps.net/?path=/docs/umb-document-tracked-reference-table--docs

### Screenshots

As displayed in the unpublish modal:
image


leekelleher (changes requested):
Tested the UI, it works as expected - with both `disableUnpublishWhenReferenced` states, looking good. I spotted an issue with the tracked references being displayed when there were no references, which I've fixed and committed to this PR, 9b4c7588903a4a1656bee91916a2bcdd70955a29.

Unfortunately, there's an issue with the unpublish mechanism for non-variant documents. The `#selectionManager` appears to not have a selection, so doesn't trigger the unpublish event. I'll try to debug, but you may have a quicker idea on a fix?

leekelleher (approved):
Re-tested, the non-variants now work as expected. 👍

PR #1433, opened 2024-03-15
Feature/publish with descendants

iOvergaard:
### Description
Adds the "Publish with descendants" workspace menu item and modal.

### How to test
- [x] Go to a page with **no variants** and check that the modal opens with a text including the name of the page
- [x] Go to a page with **more than one variant** and check that they can be selected
- [x] Verify that "Include unpublished content items" work as expected, i.e. it publishes content item children that were not previously published

There are two stories to test both views of the modal: https://ambitious-stone-0033b3603-1433.westeurope.1.azurestaticapps.net/?path=/docs/umb-document-publish-with-descendants-modal--docs

### Screenshots

Menu item:
image

Modal:
image

### DIFF
--- a/src/packages/documents/documents/modals/index.ts
+++ b/src/packages/documents/documents/modals/index.ts
@@ -1,4 +1,5 @@
 export * from './publish-modal/index.js';
+export * from './publish-with-descendants-modal/index.js';
 export * from './save-modal/index.js';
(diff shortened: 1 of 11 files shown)


leekelleher (approved):
Tested, works as expected! 👍

PR #1434, opened 2024-03-15
add weight + icons + missing stubs for entity actions

iOvergaard:
### Description

This adds a weight + a few missing actions to signal where the Core actions will be located and under which weight. This enables extension developers to be aware of which weight the built-in actions have so that they may place their own actions.

image

### DIFF
--- a/src/packages/core/entity-action/common/sort-children-of/manifests.ts
+++ b/src/packages/core/entity-action/common/sort-children-of/manifests.ts
@@ -1,0 +1,3 @@
+import { manifest as sortKindManifest } from './sort-children-of.action.kind.js';
+
+export const manifests = [sortKindManifest];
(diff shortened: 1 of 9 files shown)


bjarnef:
@iOvergaard not sure if I think the RSS feed icon for notifications 😅

could it use the megaphone as in the current backoffice https://lucide.dev/icons/megaphone or perhaps bell https://lucide.dev/icons/bell which often seems to be use in admin interfaces although typically in header. Off topic - perhaps as a notification header app at some point. E.g. with the send to translation action.

leekelleher (approved):
Tested out. Updated weightings are looking good. 👍 

I have questions about the use of ellipsis, I'm sure the old backoffice had an option to `add-ellipsis` on a button if it required further interaction, e.g. opened a modal.

Here the ellipsis are part of the manifest's label. Just wondering whether they should be?

The use-case I'm thinking of is "Publish", it doesn't have an ellipsis suffix, which is fine for non-variant documents, but for variant documents it would open a modal for further interaction, so should have an ellipsis.

Anyway, I think that's for a discussion outside of the remit of this PR.

PR #1436, opened 2024-03-18
Bugfix/add property route fix

nielslyngsoe:
Fix re-render of groups, so add property routes works.

Further investigation will follow, but this is the fix for now.

I did some testing with compositions, merging and unmerging, and it seems to still work fine :-)

## Types of changes

## Checklist


leekelleher (approved):
Tested it out. I can add a new property-group (entering the label) and add new properties to it.

PR #1439, opened 2024-03-18
rename selectedIds + Corrections for Lit warnings

nielslyngsoe:
Get rid of a few more Lit warnings on getter/setter methods

And this also corrects `selectedIds` property to become `selection` to become consistent across the app. (This is a breaking change, not sure how widely it would be implemented but notice this!)

And this removed `@property` decorator for many config setter on property editors, as they where never used in the rendering and therefor not needed as a `@property`. (As well configs should never be set as an attribute, removing the other potential reason for this decorator)

And corrects if-sentence that hides add button, we should only hide add button if max is set to 1. Otherwise we want to keep the ability to add, and then have the validation kicking it — Validation is still work in progress.

## Types of changes

## Checklist


leekelleher (approved):
Tested out, everything is still working as expected. 👍

PR #1452, opened 2024-03-19
Feature/brand icons

iOvergaard:
Added a few icons that were added late in V13 which we didn't migrate to V14.

I couldn't find all of them, e.g. the "Google" icon doesn't exist on Lucide.


bjarnef:
It doesn't look like the "x" icon is the X (Twitter) icon - just a cross/close icon.
https://lucide.dev/icons/x

iOvergaard:
@bjarnef Yeah, Lucide isn't great for that, is it? :(

Simple Icons look quite different than Lucide icons. It seems there is an active request to remove all brand icons from Lucide, so this pull request is a no-go. We need to do something else entirely.

I will close this PR while we figure something out.

bjarnef:
@iOvergaard yes, perhaps there's a better icon set somewhere for brand icons, but there are potentially many and from time to time they tends to redesign brand/social icon anyway 😅
I guess they are mainly used for external login providers in core. Maybe it could just pick some SVG icons with a similar style (outlined as Lucide icons) or an extension if external login providers are an addon/package?

bjarnef:
That seems to be great to use and better than picking some random Social icons from different sources.
I guess developers can include their own social icons and possible replace the ones used for external login providers if they wanted to use a different icon set like [Material Icons ](https://mui.com/material-ui/material-icons/) or new icon sets later.

iOvergaard:
They should be able to replace them, yes, but we haven't figured out how yet.

leekelleher:
FWIW, Simple Icons has a [CC0 1.0 Universal license](https://creativecommons.org/publicdomain/zero/1.0/), so would be the most simplest use (in terms of licensing).

PR #1454, opened 2024-03-19
Optimise `<umb-localize>` to maintain a fixed text node

iOvergaard:
optimise the localize `render` method to stick to `get text()` rather than re-rendering its own root over and over


iOvergaard:
Closing this for now since we don't have time to look into it.

PR #1458, opened 2024-03-20
Feature: Active menu/tree item

madsrasmussen:
This PR adds an active state to a menu/tree item. The active state is based on whether the path of the navigation item is part of the current location. This also means that there is no need for a custom implementation of the active state.

Uploading Kapture 2024-03-20 at 13.53.19.mp4…

**Known missings:**
* This PR only implements the active state. The tree won't unfold to the active item. That will be fixed in another PR.

### DIFF
--- a/src/packages/core/menu/menu-item-layout/menu-item-layout.element.ts
+++ b/src/packages/core/menu/menu-item-layout/menu-item-layout.element.ts
@@ -1,5 +1,6 @@
-import { html, customElement, property, ifDefined } from '@umbraco-cms/backoffice/external/lit';
+import { html, customElement, property, ifDefined, state } from '@umbraco-cms/backoffice/external/lit';
 import { UmbLitElement } from '@umbraco-cms/backoffice/lit-element';
+import { debounce } from '@umbraco-cms/backoffice/utils';
 
 /**
  * @element umb-menu-item-layout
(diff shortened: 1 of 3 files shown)


mattbrailsford:
Hmm, I don't think I can use this yet. My menu currently renders out the `uui-menu-item` directly, rather than using the `umb-menu-item-layout` I think this is mainly because there is no way to pass context to entity actions so I have to pass the parent it in the `unique` attribute, but if I used `umb.menu-item-layout` that is always set to null.

We need a way of passing context to menu items.

madsrasmussen:
I think our implementation is ok, then we will have to figure out how to make it reusable of other not using the layout.

For a menu-item the implementation is pretty simple, so I think it would be ok to just copy it until we find a way to inherit / reuse it. 

https://github.com/umbraco/Umbraco.CMS.Backoffice/commit/1d274b3cd769a1bda4541508cf2da7b8ff2d8b20

mattbrailsford:
Was considering trying today whether my menu could expose a context that the entity actions could use. If that works, I should be able to use the menu item layout.

mattbrailsford:
Doesn't look like the approach of setting a context from a parent menu item works. I assume the entity actions render outside the dom of the parent menu item and so the context isn't returned. So yea, could do with some way of sending contextual args.

PR #1464, opened 2024-03-20
Feature: Item Picker modal filter

leekelleher:
Adds a filter input to the Item Picker modal.


iOvergaard (approved):

iOvergaard:
Looks good but you have a conflict :)

PR #1468, opened 2024-03-21
rename Templates to Views in UI

nielslyngsoe:
Ideally we would also rename everything in code, but for now lets change it in the UI and then we can rename things in code when we have more time.

## Types of changes

## Checklist


iOvergaard:
We spoke about this on Slack and agreed that either we rename everything or nothing at all at this stage. We'd like to do this, but since it will introduce breaking changes if we go all the way, we'd like to do it in another major version such as V15 later this year 👍

PR #1472, opened 2024-03-21
Feature: Datatypepickerflow - Property Icons

loivsen:
## Description

Added icons added in datatypeflowpicker based on propertyEditorUi
Added icons for data-type tree based on propertyEditorUi
Some datatypes still shows `icon-circle-dotted` due to the property `propertyEditorUiAlias` received from endpoint is `undefined`.

## Types of changes

### DIFF
--- a/src/packages/data-type/components/ref-data-type/ref-data-type.element.ts
+++ b/src/packages/data-type/components/ref-data-type/ref-data-type.element.ts
@@ -31,9 +31,9 @@ export class UmbRefDataTypeElement extends UmbElementMixin(UUIRefNodeElement) {
 				(dataType) => {
 					if (dataType) {
 						this.name = dataType.name ?? '';
-						this.propertyEditorUiAlias = dataType.editorUiAlias ?? '';
 						this.propertyEditorSchemaAlias = dataType.editorAlias ?? '';
-						if (this.propertyEditorUiAlias) {
+						if (dataType.editorUiAlias ?? '' !== this.propertyEditorUiAlias) {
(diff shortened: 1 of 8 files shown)


src/packages/data-type/components/ref-data-type/ref-data-type.element.ts
```diff
@@ -30,6 +33,9 @@ export class UmbRefDataTypeElement extends UmbElementMixin(UUIRefNodeElement) {
 						this.name = dataType.name ?? '';
 						this.propertyEditorUiAlias = dataType.editorUiAlias ?? '';
 						this.propertyEditorSchemaAlias = dataType.editorAlias ?? '';
+						if (this.propertyEditorUiAlias) {
```
nielslyngsoe, line 36: Want is was looking for here, was not only to just for this but also so do this: ```dataType.editorUiAlias ?? '' !== this.propertyEditorUiAlias```

and then move ```this.propertyEditorUiAlias = dataType.editorUiAlias ?? '';```inside the If-sentence.

nielslyngsoe: So getIconFromUiAlias is only called if that specific property did change. Otherwise we would load the icon again if another property of the manifest got changed. And we do only want to trigger an actual change of what effects the icon.. aka. the propertyEditorUiAlias.

loivsen: Ah I think i get it now , thank you 😄

nielslyngsoe (approved):
looks good

PR #1474, opened 2024-03-22
Bugfix: Relation types

madsrasmussen:
> [!IMPORTANT]  
> Merges together with PR: https://github.com/umbraco/Umbraco-CMS/pull/15910

## Description

This PR reimplements the relation types and relations UI. The UI has been implemented as Read Only because it doesn't make sense to create relation types in the UI, as they also need a server implementation. We have also removed the tree for relation types as they can't be nested or support folders. 

The UI is now simplified as a list of relation types. When you navigate to a relation type you can see all relations for that type including the details for the relation type.

Relations on a relation-type detail workspace are paginated and will show 50 relations per page.

**Demo:**

https://github.com/umbraco/Umbraco.CMS.Backoffice/assets/6078361/d44aa4ff-635a-4c19-b4d3-18ff63a38c31

With the UI updates, we have also aligned the repositories so they follow the same naming as the rest of our repositories.

- [ ] Chore (minor updates related to the tooling or maintenance of the repository, do not impact compiled assets)

> [!CAUTION]
> Breaking Changes:

### Changes to Relation Types
* Renamed `UmbRelationTypeRepository` to `UmbRelationTypeDetailRepository`
* Renamed `requestById` to `requestByUnique`
* Removed `requestRelationsById` method
* Removed `save` method

```ts
// Current version
const repo = new UmbRelationTypeRepository();
const { data } = await repo.requestById('1234-1234');

// New version
const repo = new UmbRelationTypeDetailRepository();
const { data } = await repo.requestByUnique('1234-1234');
```

### Changes to Relations

* Renamed `UmbRelationRepository` to `UmbRelationCollectionRepository`
* Renamed method `requestChildRelationById` to `requestCollection`

```ts
// Current version
const repo = new UmbRelationRepository();
const { data } = await repo.requestChildRelationById( '1234-123')

// New version
const repo = new UmbRelationCollectionRepository();
const { data } = await repo.requestCollection({ relationTypeUnique: '1234-123' })
```

#### Known issues
* Menu item loses active state when navigating to relation type details


leekelleher (approved):
Functionality this works well, all as expected. I did notice a few cosmetic issues, which could be picked up on later.

On Relations overview...

- Large gap above the table, (due to the empty `<umb-collection-toolbar>`)

On Relation Type detail view...

- Missing breadcrumbs
- Missing a back button, (or if we have a UI pattern for going back to a Collection view?)
- Whilst relations are loading, the UI says "No relations found" for a moment. We could have a `<uui-loading>`?
- For "Related Document" and "Related Media", in the info box, the "Parent Type" and "Child Type" are both empty. I believe this is a server/data concern, (as it's missing from the API response), but it's whether we do anything about this in the UI?
- Wondering if the Relation's Parent and Child should be able to open the corresponding item editor? (although, this may open a can of worms)

PR #1478, opened 2024-03-22
add script to check for a configurable length of paths

iOvergaard:


nielslyngsoe (approved):
I guess this is a fine solution until we fine better ways for overcome this problem.

PR #1479, opened 2024-03-22
Feature/public-extensions

iOvergaard:
## Description

This ensures we can load public extensions outside of the backoffice, e.g. on the login screen.


nielslyngsoe (approved):
Good :-)

PR #1502, opened 2024-03-27
Feature/tracked references server update

iOvergaard:
**Unpublish**
image

**Info tab (documents)**
image

**Info tab (media)**
image


src/packages/documents/documents/tracked-reference/repository/document-tracked-reference.repository.ts
```diff
@@ -10,29 +10,19 @@ export class UmbDocumentTrackedReferenceRepository extends UmbControllerBase {
...
-			parentUnique,
-			skip,
-			take,
-			filterMustBeIsDependency,
-		);
+		return this.#trackedReferenceSource.getTrackedReferenceDescendantsByParentId(parentUnique, skip, take);
 	}
 
-	async requestTrackedReferenceItems(uniques: string[], skip = 0, take = 20, filterMustBeIsDependency = true) {
+	async requestTrackedReferenceItems(uniques: string[], skip = 0, take = 20) {
```
madsrasmussen, line 23: It doesn't look like this method is used. I think we should remove it for simplicity. Then we can also implement the items logic like the rest of the repositories. The same goes for the media repo + data source.

src/packages/documents/documents/tracked-reference/repository/document-tracked-reference.repository.ts
```diff
@@ -10,29 +10,19 @@ export class UmbDocumentTrackedReferenceRepository extends UmbControllerBase {
...
+		return this.#trackedReferenceSource.getTrackedReferenceById(unique, skip, take);
 	}
 
-	async requestTrackedReferenceDescendantsFromParentUnique(
-		parentUnique: string,
-		skip = 0,
-		take = 20,
-		filterMustBeIsDependency = false,
-	) {
+	async requestTrackedReferenceDescendantsFromParentUnique(parentUnique: string, skip = 0, take = 20) {
```
madsrasmussen, line 18: It doesn't look like this method is used. I think we should remove it for simplicity. The same goes for the media repo + data source.

src/packages/documents/documents/tracked-reference/repository/document-tracked-reference.server.data.ts
```diff
@@ -61,14 +52,13 @@ export class UmbDocumentTrackedReferenceServerDataSource {
 	 * @return {*}
 	 * @memberof UmbDocumentTrackedReferenceServerDataSource
 	 */
-	async getTrackedReferenceItem(id: string[], skip = 0, take = 20, filterMustBeIsDependency = true) {
+	async getTrackedReferenceItem(id: string[], skip = 0, take = 20) {
```
madsrasmussen, line 55: It doesn't look like this method is used. I think we should remove it for simplicity. Then we can also implement the items logic like the rest of the repositories.

src/packages/documents/documents/tracked-reference/repository/document-tracked-reference.server.data.ts
```diff
@@ -38,19 +35,13 @@ export class UmbDocumentTrackedReferenceServerDataSource {
 	 * @return {*}
 	 * @memberof UmbDocumentTrackedReferenceServerDataSource
 	 */
-	async getTrackedReferenceDescendantsByParentId(
-		parentId: string,
-		skip = 0,
-		take = 20,
-		filterMustBeIsDependency = false,
-	) {
+	async getTrackedReferenceDescendantsByParentId(parentId: string, skip = 0, take = 20) {
```
madsrasmussen, line 38: It doesn't look like this method is used. I think we should remove it for simplicity.

src/packages/relations/relations/types.ts
```diff
@@ -17,3 +22,8 @@ export interface UmbRelationDetailModel {
 	createDate: string;
 	comment: string | null;
 }
+
+export type UmbReferenceModel =
```
madsrasmussen, line 26: I think we should get our types for these, but it might be something we do after this release.

madsrasmussen (approved):
In general it looks good. I would like if we could remove the "tracked" part of the name, so its just called a "reference", but it might be something for another PR.

PR #1503, opened 2024-03-27
Feature: Media Collection Dropzone +  TempFile Upload Changes 

loivsen:
## Description
Temporary File Manager
- Temp file manager now have a return value when done uploading (file, status and id).

Media Collection Dropzone:
- drag and drop file(s) into the media collection

## Types of changes


leekelleher (approved):
Small coding style suggestion, the `getMediaTypeByFileExtension` utility function could use a `switch` statement rather than several `if`s, but that's me nitpicking. 😊

Good work, it is all working well. 🚀 

I've added in a new event called `UmbProgressEvent` and wired up a `<uui-loader-bar>` to show the upload progress. I would have chatted with you about this, but due to public holiday in DK, I decided to add it in, (we can review and polish it later).

PR #1519, opened 2024-04-02
Feature: `blobDownload` function

leekelleher:
Adds a `blobDownload` utility function, used to prompt a browser download of content delivered from the server.

This was extracted from the `export.action.ts` code and reused in `workspace-package-builder.element.ts`.

### DIFF
--- a/src/packages/core/utils/download/blob-download.function.ts
+++ b/src/packages/core/utils/download/blob-download.function.ts
@@ -1,0 +1,25 @@
+/**
+ * Triggers a client-side download of a file, using a `Blob` object.
+ * To do this, a temporary anchor element is created, appended to the document body,
+ * immediate triggered to download, then removed from the document body.
+ *
+ * @param {any} data - The data to be downloaded.
+ * @param {string} filename - The name of the file to be downloaded.
+ * @param {string} mimeType - The MIME type of the file to be downloaded.
+ * @returns {void}
+ *
+ * @example
+ * blobDownload(data, 'package.xml', 'text/xml');
+ */
+export const blobDownload = (data: any, filename: string, mimeType: string) => {
+	const blob = new Blob([data], { type: mimeType });
+	const url = window.URL.createObjectURL(blob);
+	const a = document.createElement('a');
+	a.href = url;
+	a.download = filename;
+	a.style.display = 'none';
+	document.body.appendChild(a);
+	a.dispatchEvent(new MouseEvent('click'));
+	a.remove();
+	window.URL.revokeObjectURL(url);
+};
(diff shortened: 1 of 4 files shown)


src/packages/core/utils/download/blob-download.function.ts
```diff
@@ -0,0 +1,25 @@
...
+	const url = window.URL.createObjectURL(blob);
+	const a = document.createElement('a');
+	a.href = url;
+	a.download = filename;
+	a.style.display = 'none';
+	document.body.appendChild(a);
+	a.dispatchEvent(new MouseEvent('click'));
+	a.remove();
+	window.URL.revokeObjectURL(url);
+};
```
bjarnef, line 25: I think this in fine for older browsers / fallback to simulate click on anchor with `download` attribute, but I am pretty sure there a more modern ways to handle this using the HTML5 API:
https://stackoverflow.com/a/76716324

bjarnef: IIRC this part works in IE / Edge

```
if (navigator.msSaveBlob){ // For IE and Edge
   return navigator.msSaveBlob(file.content, file.name);
}
```

https://medium.com/@szpytfire/create-a-downloadable-file-using-javascript-typescript-c7d8bf798baa

leekelleher: @bjarnef I think the post may be outdated (circa 2020), since we're no longer supporting IE11, and Edge switch engines from EdgeHTML to Blink in 2021. 😬

bjarnef: Yes, it was just from an example I previous used in a few projects although it does fallback to the hacky anchor link 😅

```
const downloadToFile = (content, filename, contentType) => {

      const blob = new Blob([content], { type: contentType });

      if (
          window.navigator &&
          window.navigator.msSaveOrOpenBlob
      ) return window.navigator.msSaveOrOpenBlob(blob);

      var URL = window.URL || window.webkitURL;

      // Convert your blob into a Blob URL (a special url that points to an object in the browser's memory)
      const blobUrl = URL.createObjectURL(blob);

      const link = document.createElement('a');
      link.href = blobUrl;
      link.download = filename;

      // Append link to the body
      document.body.appendChild(link);

      // Dispatch click event on the link
      // This is necessary as link.click() does not work on the latest firefox
      link.dispatchEvent(
          new MouseEvent('click', {
              bubbles: true,
              cancelable: true,
              view: window
          })
      );

      // Remove link from body
      document.body.removeChild(link);
  };
```

But maybe there isn't a better way to handle this nowadays.. just seems old school 👴

Maybe something like this?
https://stackoverflow.com/a/29390393

iOvergaard (approved):

PR #1524, opened 2024-04-02
Remove exports from packages/core/index.ts

nielslyngsoe:
In general we should use the dedicated import maps, this was most likely some legacy code that never was cleaned up. So I have removed it.

Breaking if anyone use those imports via the core import map. We did not, so I dont think its very likely.

## Types of changes

## Checklist


src/packages/core/workspace/components/workspace-action/default/workspace-action.element.ts
```diff
@@ -1,8 +1,8 @@
-import type { UmbWorkspaceAction } from '../../../../index.js';
 import { UmbActionExecutedEvent } from '@umbraco-cms/backoffice/event';
 import { html, customElement, property, state } from '@umbraco-cms/backoffice/external/lit';
 import type { UUIButtonState } from '@umbraco-cms/backoffice/external/uui';
 import { UmbLitElement } from '@umbraco-cms/backoffice/lit-element';
+import type { UmbWorkspaceAction } from '@umbraco-cms/backoffice/workspace';
```
iOvergaard, line 5: Doesn't this import from itself creating a cyclic dependency?

nielslyngsoe: Sharp one, I couldn't make it complain about it, I guess its because it's an type import(only a compile time import) and therefor okay with circular. I have of-course changed it to prevent a potential issue here :-)

iOvergaard: You are right, the `import type` prevents it from being an actual runtime danger :)

iOvergaard (changes requested):
Storybook seems to be failing because it can't import `core/components` now, would you check, please? :)

PR #1534, opened 2024-04-03
Feature/admin mfa login providers

iOvergaard:
## Description

Allow admins to disable MFA for all users, and users to disable for other users. You can check the state of 2fa providers via the dialog as well.

If you select yourself in the Users section, it will open the same modal flow as if you had clicked on your "avatar" in the top right corner. The only discernible difference effect is that you have to enter a verification code from your verification app to disable the provider.

https://github.com/umbraco/Umbraco.CMS.Backoffice/assets/752371/0cf3f9d5-cd77-44c0-93ea-5f2dc3fe7de3


leekelleher (approved):
Tested out, works exactly as in the video demo, all good! 👍

PR #1539, opened 2024-04-04
`umb-stack` element

mattbrailsford:
## Description

The PR introduces a new `umb-stack` element. The purpose of this element is to provide consistent spacing between stacks of items.

image

There is a default and a compact mode, as well as a flag to include a dividing line. 

## Types of changes

## Motivation and context

I was finding myself adding the same styles in a lot of places to space items in a list. These were becoming inconsistent and a annoying to copy. I thought it would be good if we had a component that applied the spacing when needed. This should also be useful for things like `umb-property` which is currently struggling to align correctly in places. With this, we can remove any styling from individual elements, and let the stack do the layout.

## How to test?

Create a `umb-stack` element and create any number of items inside of it. These should all be spaced nicely.

```javascript
<umb-stack>
    <div>Item 1</div>
    <div>Item 2</div>
    <div>Item 3</div>
</umb-stack>
```

## Checklist

### DIFF
--- a/src/packages/core/components/index.ts
+++ b/src/packages/core/components/index.ts
@@ -33,4 +33,5 @@ export * from './multiple-color-picker-input/index.js';
 export * from './multiple-text-string-input/index.js';
 export * from './popover-layout/index.js';
 export * from './ref-item/index.js';
+export * from './stack/index.js';
 export * from './table/index.js';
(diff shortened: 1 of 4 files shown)


bjarnef:
Perhaps the divider itself could be an component something like this:
https://shoelace.style/components/divider/

But not sure it is needed specific for the stacked content (unless it should have an option/attribute og `umb-stack` to tell is dividers should be added).

PR #1553, opened 2024-04-06
Feature: Entity context

madsrasmussen:
## Description

This PR adds an Entity Context.  The context provides the current entity type and unique. It is used as the foundation for document user permissions as they currently will need to know the unique for granular permissions.

Currently, the context is provided in a few places: from the entity action list element. This element wraps all entity actions. It is also provided from the document workspace to support permissions for the workspace actions. We will need to provide it from all entity workspaces, but I will wait until we have a base class to support it.

## Types of changes


nielslyngsoe (approved):
:-)

PR #1554, opened 2024-04-06
Feature: Document User Permission

madsrasmussen:
## Description

Implementation of the Document user permissions. 

Using the condition in a manifest:

```ts
conditions: [
  {
    alias: 'Umb.Condition.UserPermission.Document',
    allOf: [UMB_USER_PERMISSION_DOCUMENT_CREATE],
  },
],
```

When using a condition on a manifest it will remove the element from the DOM if the conditions aren't met. For workspace actions, we want to disable the button instead. I have solved this by using the condition programmatically in the action class and setting the disable state based on the condition state. It looks something like this:

```ts
const condition = new UmbDocumentUserPermissionCondition(host, {
  host,
  config: {
    alias: 'Umb.Condition.UserPermission.Document',
    allOf: [UMB_USER_PERMISSION_DOCUMENT_UPDATE, UMB_USER_PERMISSION_DOCUMENT_PUBLISH],
  },
  onChange: () => {
    condition.permitted ? this.enable() : this.disable();
  },
});
```

See full example here: https://github.com/umbraco/Umbraco.CMS.Backoffice/pull/1554/files#diff-1ed58ab655d2a26796804f499d7eb7c1ee496c486763fdc16f67a660411852f0R11

## How to test?

* Set up different permission rules for user groups. Both the fallback (default) permissions and granular permissions
* Make sure that the user can only do what is allowed.

## Types of changes


leekelleher (approved):
Tested, works as expected. 👍

PR #1557, opened 2024-04-08
Rollback

JesmoDev:
Adds rollback with diffing

Also adds the package [diff](https://www.npmjs.com/package/diff). Same package as used in the old backoffice
Also corrects the naming in `document-property-dataset-context.token.ts`


src/packages/documents/documents/modals/rollback/repository/rollback.repository.ts
```diff
@@ -0,0 +1,36 @@
...
+
+	async setPreventCleanup(versionId: string, preventCleanup: boolean) {
+		const { error } = await this.#dataSource.setPreventCleanup(versionId, preventCleanup);
+		return { error };
+	}
+
+	async rollback(versionId: string, culture?: string) {
+		const { error } = await this.#dataSource.rollback(versionId, culture);
+		return { error };
+	}
```
iOvergaard, line 33: It's okay to return directly in a repository if you are not using the data for anything rather than extracting each variable here. You could for instance be showing a success notification for some of these if there is no error.

src/packages/documents/documents/workspace/views/info/document-workspace-view-info-history.element.ts
```diff
@@ -88,6 +95,16 @@ export class UmbDocumentWorkspaceViewInfoHistoryElement extends UmbLitElement {
 		this.#getLogs();
 	}
 
+	#onRollbackModalOpen = () => {
+		const modalContext = this.#modalContext?.open(this, UMB_ROLLBACK_MODAL, {});
```
iOvergaard, line 99: Consider calling on the modal manager with `this.getContext(UMB_MODAL_MANAGER_CONTEXT)` rather than having to listen on it constantly, since this is a one-off method that is only executed when you click on something else.

src/packages/documents/documents/modals/rollback/rollback-modal.token.ts
```diff
@@ -0,0 +1,15 @@
...
+
+export interface UmbRollbackModalData {}
+export interface UmbRollbackModalValue {}
+
+export const UMB_ROLLBACK_MODAL = new UmbModalToken<UmbRollbackModalData, UmbRollbackModalValue>(
+	UMB_ROLLBACK_MODAL_ALIAS,
+	{
+		modal: {
+			type: 'sidebar',
+			size: 'full',
```
iOvergaard, line 12: Does the `size` property actually do anything, and if so, is it supposed to be "full"?

iOvergaard (changes requested):
Looks good, left a few comments for optimization, and you need to add localization in general to the elements 😎

src/packages/documents/documents/modals/rollback/rollback-modal.token.ts
JesmoDev: Size controls the width of the sidebar. Full means that it takes up the entire screen (except for a small gap on the left). Same as the old backoffice.

src/packages/documents/documents/modals/rollback/rollback-modal.element.ts
```diff
@@ -0,0 +1,380 @@
...
+			<umb-body-layout headline="Rollback">
+				<div id="main">
+					<uui-box headline=${this.localize.term('rollback_versions')} id="box-left">
+						<div>${this.#renderVersions()}</div>
+					</uui-box>
+					<uui-box headline=${this.currentVersionHeader} id="box-right"> ${this.#renderCurrentVersion()} </uui-box>
+				</div>
+				<umb-footer-layout slot="footer">
+					<uui-button slot="actions" look="secondary" @click=${this.#onCancel}
+						>${this.localize.term('general_cancel')}</uui-button
```
iOvergaard, line 290: Consider moving this to the label property instead, otherwise the button won't have a label. It will show the text automatically in the default slot if you don't have anything else there.

src/packages/documents/documents/modals/rollback/rollback-modal.element.ts
```diff
@@ -0,0 +1,380 @@
...
+					>
+					<uui-button slot="actions" look="primary" @click=${this.#onRollback}
+						>${this.localize.term('general_cancel')}</uui-button
```
iOvergaard, line 293: Consider moving this to the label property instead, otherwise the button won't have a label. It will show the text automatically in the default slot if you don't have anything else there.

src/packages/documents/documents/workspace/views/info/document-workspace-view-info-history.element.ts
```diff
@@ -88,6 +90,17 @@ export class UmbDocumentWorkspaceViewInfoHistoryElement extends UmbLitElement {
 		this.#getLogs();
 	}
 
+	#onRollbackModalOpen = async () => {
+		const modalManagerContext = await this.getContext(UMB_MODAL_MANAGER_CONTEXT);
+		const modalContext = modalManagerContext.open(this, UMB_ROLLBACK_MODAL, {});
+
+		if (!modalContext) return;
```
iOvergaard, line 97: Is this ever a case that occurs?

src/packages/documents/documents/workspace/views/info/document-workspace-view-info-history.element.ts
```diff
@@ -88,6 +90,17 @@ export class UmbDocumentWorkspaceViewInfoHistoryElement extends UmbLitElement {
...
+
+		modalContext.onSubmit().then(() => {
```
iOvergaard, line 99: Consider using the async/await pattern here.

src/packages/documents/documents/workspace/views/info/document-workspace-view-info-history.element.ts
```diff
@@ -88,6 +90,17 @@ export class UmbDocumentWorkspaceViewInfoHistoryElement extends UmbLitElement {
...
+			console.log('Rollback modal submitted');
```
iOvergaard, line 100: I'm not sure it's a good idea to have console logs here. Should we show a notification or maybe it doesn't matter?

src/packages/documents/documents/workspace/views/info/document-workspace-view-info-history.element.ts
```diff
@@ -96,7 +109,7 @@ export class UmbDocumentWorkspaceViewInfoHistoryElement extends UmbLitElement {
 						label=${this.localize.term('actions_rollback')}
 						look="secondary"
 						slot="actions"
-						@click=${() => alert('TODO: Rollback Modal')}>
+						@click=${this.#onRollbackModalOpen}>
 						<uui-icon name="icon-undo"></uui-icon>
 						<umb-localize key="actions_rollback"></umb-localize>
```
iOvergaard, line 114: This is already set as a label so we don't need to set it in the default slot as well.

src/packages/documents/documents/modals/rollback/rollback-modal.element.ts
```diff
@@ -0,0 +1,380 @@
...
+		const tempItems: DocumentVersion[] = [];
+
+		const uniqueUserIds = [...new Set(data?.items.map((item) => item.user.id))];
+
+		const { data: userItems } = await this.#userItemRepository.requestItems(uniqueUserIds);
+
+		data?.items.forEach((item: any) => {
+			tempItems.push({
+				date: item.versionDate,
+				user: userItems?.find((user) => user.unique === item.user.id)?.name || 'Unknown user',
```
iOvergaard, line 101: Please add localization for 'Unknown user'

src/packages/documents/documents/modals/rollback/rollback-modal.element.ts
```diff
@@ -0,0 +1,380 @@
...
+		if (!this.currentVersion) return;
+
+		const id = this.currentVersion.id;
+		const culture = this.availableVariants.length > 1 ? this.currentCulture : undefined;
+		this.#rollbackRepository.rollback(id, culture);
+
+		const docUnique = this.#workspaceContext?.getUnique() ?? '';
+		// TODO Use the load method on the context instead of location.href, when it works.
+		// this.#workspaceContext?.load(docUnique);
+		location.href = 'section/content/workspace/document/edit/' + docUnique;
```
iOvergaard, line 149: Would it perhaps be better to use history.pushState here?

iOvergaard (changes requested):

iOvergaard (approved):

PR #1562, opened 2024-04-09
Feature: Adds custom request file for `openapi-typescript-codegen`

leekelleher:
> [!IMPORTANT]
> If PR #1565 is merged in, then we can disregard this PR.

Part of the code fixes in PR #1538 for the Packages section had code to detect whether a server response was a `Blob` and handle it accordingly. I'd unwittingly made the fix in the `request.ts`, which is auto-generated by the  `openapi-typescript-codegen` tool. So a subsequent update of the server models meant that my code fix had been reverted. 🤦 

Following the instructions from: https://github.com/ferdikoomen/openapi-typescript-codegen/wiki/Custom-request-file, it is possible to have a custom request file which is copied onto `request.ts` when the server models are regenerated.

---

**Notes**

I am happy to rename or move `request.custom.ts` to something more appropriate.

TypeScript threw a bunch of errors in the custom request file, so I've added it to the "exclude" in `tsconfig.json`. If there is a better way to handle this, I'm open to suggestions.


madsrasmussen:
Replaced by #1565

PR #1565, opened 2024-04-10
Feature: Replace `openapi-typescript-codegen` with `@hey-api/openapi-ts`

leekelleher:
Since the [`openapi-typescript-codegen` package has announced it will be deprecated from 1st of May 2024](https://github.com/ferdikoomen/openapi-typescript-codegen?tab=readme-ov-file#important-announcement), a fork has made at [`@hey-api/openapi-ts`](https://github.com/hey-api/openapi-ts).

This PR migrates the OpenAPI TypeScript code-generation script to use the `@hey-api/openapi-ts` library.

There are a few opinionated decisions made in the fork which will affect our existing code, but no current showstoppers.

- All the generated models and service code are no longer created as individual files, they are placed into single files, `models.ts` and `services.ts` respectively.
- The use of `any` type has been replaced with `unknown`, _(not a bad thing, only caused TypeScript compile errors, which were fixable)._
- The `postfixServices` parameter has been deprecated (for a future release), with no alternative feature. So we will need to replace the `Resource` suffix with `Service` at some point, _(I didn't want to pollute this PR with that particular change)_.

This PR uninstalls the `openapi-typescript-codegen` package, updating the existing `"generate:server-api"` script commands (and local/dev variations) with the `openapi-ts` tool.

To test, run...
```
`npm run generate:server-api`
```
...and check the backoffice runs as expected.


iOvergaard (commented):
Looks good to me. I think we should expose our interceptors in `UmbAuthContext.getOpenApiConfiguration()` so that people can hook on to those for their own APIs if they want.

nielslyngsoe (approved):
I read the code changes, they look good.
The conversation continued on Slack where we talked about how to test the changes...

mrlubos:
Hey all, I just want to add that if you ever need a feature from us, feel free to reach out. As noted in this pull request's description, we want to be opinionated on generated code because the opposite makes it harder to maintain the package and ship features, and it may lead users down the wrong path. For example, people using arguments style parameters (they were enabled by default) have a really hard time migrating to options as their codebase might have hundreds/thousands of errors to resolve.

iOvergaard:
Hi @mrlubos, thank you so much for your message and your sentiments! We are pleased to dive into this welcomed upgrade of openapi-typescript-codegen which already seems to solve a couple of issues we were having. We expect to be using even more of your new package soon regarding interceptors.

(and yes, luckily we were already using the "parameters as options" mode before -- the arguments style didn't seem maintainable)

PR #1570, opened 2024-04-10
Feature/user-profile-actions

iOvergaard:
## Description

Adds a model to add buttons to the current user modal. This will also allow extensions to unregister the built-in Edit and Change Password buttons, which could be needed for instance if you don't want to allow externally created users to edit their local user.

The buttons will show up in the red circle:

image

The next step is to convert our buttons to this new type, which will come in another PR.

## How to test

To test it you need to register a button with an API:

```ts
// manifest
	{
		type: 'currentUserAction',
		kind: 'default',
		alias: 'Umb.CurrentUser.Button.ChangePassword',
		name: 'Current User Change Password Button',
		weight: 900,
		api: UmbChangePasswordCurrentUserAction,
		meta: {
			label: '#general_changePassword',
			icon: 'lock',
		},
	},
```

```ts
// API
import { UmbActionBase } from '@umbraco-cms/backoffice/action';
import type { UmbCurrentUserAction, UmbCurrentUserActionArgs } from '@umbraco-cms/backoffice/extension-registry';

export class UmbChangePasswordCurrentUserAction<ArgsMetaType = never>
	extends UmbActionBase<UmbCurrentUserActionArgs<ArgsMetaType>>
	implements UmbCurrentUserAction<ArgsMetaType>
{
	async getHref() {
		return undefined;
	}

	async execute() {
		alert('the user wants to change their password');
	}
}
```

### DIFF
--- a/src/packages/core/extension-registry/models/current-user-action.model.ts
+++ b/src/packages/core/extension-registry/models/current-user-action.model.ts
@@ -1,0 +1,73 @@
+import type { ConditionTypes } from '../conditions/types.js';
+import type { UmbAction } from '../../action/action.interface.js';
+import type { UmbControllerHostElement } from '@umbraco-cms/backoffice/controller-api';
+import type { ManifestElementAndApi, ManifestWithDynamicConditions } from '@umbraco-cms/backoffice/extension-api';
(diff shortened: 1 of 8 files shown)


leekelleher (approved):
Great addition, works well! 🚀

PR #1571, opened 2024-04-10
Feature/user-profile-actions-default

iOvergaard:
> [!WARNING]
> This depends on https://github.com/umbraco/Umbraco.CMS.Backoffice/pull/1570 being merged first

## Description

- Convert all current user buttons into currentUserAction's.
- Removes 'External login' until it can be properly implemented
- Convert the 'Configure Two-Factor' button into an action instead, since it only ever opens a modal anyway

## Screenshots

Before:

image

After:

image


leekelleher (approved):
Superb, I like this feature. Tested, all works as expected! 🙌

PR #1572, opened 2024-04-10
update lucide-static to 0.367.0 and remap now renamed icons

iOvergaard:
## Description

- Remapped all renamed icons 
- Fixed some that missed the `.svg` file extension
- Added `icon-rectangle-ellipsis` that we use for "Configure Two-Factor" in https://github.com/umbraco/Umbraco.CMS.Backoffice/pull/1571


leekelleher (approved):
Tested out, icons are displaying in the backoffice as expected. 👍

PR #1574, opened 2024-04-10
Bugfix: Picker inputs, updated to use `UmbUniqueTreeItemModel`

leekelleher:
## Description

Straddling between chore and bugfix, I'd noticed that an issue in the MemberType picker where the `MemberTypeItemResponseModel` had bled through, whereas the `UmbUniqueTreeItemModel` could be used. Which also solved an issue I had with the Entity picker input component.

## Types of changes

### DIFF
--- a/src/packages/core/components/input-entity/input-entity.element.ts
+++ b/src/packages/core/components/input-entity/input-entity.element.ts
@@ -4,7 +4,7 @@ import { UUIFormControlMixin } from '@umbraco-cms/backoffice/external/uui';
 import { UmbLitElement } from '@umbraco-cms/backoffice/lit-element';
 import type { UmbControllerHost } from '@umbraco-cms/backoffice/controller-api';
 import type { UmbPickerInputContext } from '@umbraco-cms/backoffice/picker-input';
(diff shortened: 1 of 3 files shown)


madsrasmussen:
In this case, I think the correct model is an `UmbUniqueItemModel`. It is the item endpoint we use to look up what has been picked. I don't know if we have this model (which is a mistake), so it would be good to add one.

I have for a long time thought about removing the "unique" part of the name. We have transitioned everything to work with a unique, so it might be good to get it on the base model instead. Any thoughts?

leekelleher:
@madsrasmussen Thanks for the feedback, I've put this PR to draft and I'll investigate the `UmbUniqueItemModel` type (if it exists or not yet). 👍

leekelleher:
Okay, the `UmbUniqueItemModel` didn't exist, so I've added that for now.  I agree that we could (should) drop the "Unique" part from the name, but `UmbItemModel` feels too opinionated (that an item must have a unique, name and icon), so started to think along the lines of `UmbPickedItemModel`, but I'm unsure.  So we'll go with `UmbUniqueItemModel` and refactor when we have a better name.

madsrasmussen (approved):
Looks good to me

PR #1583, opened 2024-04-11
Feature: Blueprint Create Action URL

loivsen:
Blueprint Create Action: Navigate to the correct URL when creating a blueprint from a documenttype.


leekelleher (approved):
Tested, create action loads up a new blueprint workspace. 👍

PR #1591, opened 2024-04-12
Bugfix: Duplicate to Entity Action (part 1)

madsrasmussen:
## Description

This PR fixes the duplicate entity action.

The PR introduces 2 kinds for duplicating an entity: `duplicate` and `duplicateTo`. When using the `duplicate` kind the entity will be duplicated to the same location as the entity. When using the `duplicateTo` kind the user will have to select the destination for the copied entity.

The Duplicate functionality is implemented for all supported trees:

* Data Type: `duplicateTo` - select destination.
* Document Type: `duplicateTo` - select destination.
* Document: extended `duplicateTo` - select destination + options.
* Media Type: `duplicateTo` - select destination.
* Member Type: `duplicate` - same location.

Known missing:
* The tree destination is not reloading automatically. We will fix this together with move and restore.

## Types of changes


leekelleher (approved):
Tested out the duplications for the Data Types and Content Types, as well as Documents with and without "include descendants" option. All works as expected! 🚀

PR #1598, opened 2024-04-12
Feature: Hide entity actions when Tree used inside a Modal

leekelleher:
## Description

@madsrasmussen @nielslyngsoe As discussed recently, a proposed PR.

Adds a flag to the `umb-tree` to hide the tree-item's entity actions. This can be useful when using a tree-picker modal and don't want to enable the user to access the entity actions.

As for whether the the larger feature of "why not?" allow content-managers to run actions within pickers, we can discuss and explore further.

## Types of changes


madsrasmussen (approved):
Looks good to me

PR #1602, opened 2024-04-15
Add support for token expiration less than 10 minutes

iOvergaard:
## Description

The management api tokens are by default short-lived and potentially less that 10 minutes. This forced buffer prevents tokens from expiring in less than 10 minutes, so we set it to 0 (seconds) to allow tokens to expire faster.

Merging this PR resolves the prerequisite to merge the relevant PR on the backend: https://github.com/umbraco/Umbraco-CMS/pull/16028

## Types of changes

## How to test?

Set the Umbraco:CMS:Global:TimeOut to a value of 20 minutes and check that you are still allowed to log in:

```
  "Umbraco": {
    "CMS": {
      "Global": {
        "TimeOut": "00:20:00",
```

The 20 minutes here is the lifetime of the **refresh token**. The backend takes this value and divides it by 4 to calculate the lifetime of the **access token** which is the one we are checking with the `isValid()` function call.


leekelleher (approved):
Tested against CMS PR https://github.com/umbraco/Umbraco-CMS/pull/16028, works as expected. 👍

PR #1603, opened 2024-04-15
Temporary keyboard shortcut solution

nielslyngsoe:
Temporary ctrl + s solution for workspaces built with the submittable workspace context base class.

## Types of changes

## Checklist

### DIFF
--- a/src/packages/core/workspace/contexts/submittable-workspace-context-base.ts
+++ b/src/packages/core/workspace/contexts/submittable-workspace-context-base.ts
@@ -9,6 +9,10 @@ import { UMB_MODAL_CONTEXT } from '@umbraco-cms/backoffice/modal';
 import type { Observable } from '@umbraco-cms/backoffice/external/rxjs';
 import { UmbValidationContext } from '@umbraco-cms/backoffice/validation';
 
+// TODO: Move this feature into a keyboard shortcuts feature [NL]
+//const macOsPattern = /Mac|iPod|iPhone|iPad/;
+//const isMac = macOsPattern.test(window.navigator.platform);
+
 export abstract class UmbSubmittableWorkspaceContextBase<WorkspaceDataModelType>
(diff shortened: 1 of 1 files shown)


leekelleher (commented):
Tested on Chrome, Windows 11. Pressing <kbd>CTRL+S</kbd> prompts the browser's "Save as..." dialog. It will save the workspace, but it depends where the focus is set, e.g. inside the workspace header, but outside of that (e.g. inside a property-editor textbox), it doesn't currently work.

src/packages/core/workspace/contexts/submittable-workspace-context-base.ts
```diff
@@ -46,6 +50,16 @@ export abstract class UmbSubmittableWorkspaceContextBase<WorkspaceDataModelType>
 		this.consumeContext(UMB_MODAL_CONTEXT, (context) => {
 			(this.modalContext as UmbModalContext) = context;
 		});
+
+		// TODO: Temporary keyboard shortcut solution, should be changed to a better generic shortcut solution. [NL]
+		(this.getHostElement() as HTMLElement).addEventListener('keyup', (e: KeyboardEvent) => {
```
leekelleher, line 55: Experimenting, I had more success with using `'keydown'` to prevent the browser's "Save as..." dialog.

leekelleher: I tried binding the event listener to the `document` and had greater success, but that led to a few other issues, so I'm not suggesting that we do that, just noting that the browser's "Save as..." dialog is preventable at the top of the DOM.

nielslyngsoe:
@leekelleher I see this is way more complex than what we have time to solve right now. It seems to me we need a more extensive system. As we most likely should bind the keys to the Document, but then we need a system to activate and de-active keybindings when an 'element'(aka. workspace) is in front.. so we dont have key bindings from other modals reacting.
So I'm closing this one.

PR #1604, opened 2024-04-15
Bugfix: Date Picker story

leekelleher:
## Description

Follow up with PR #1601, this fixes up the stories for Date Picker.

On a side-note, I keep hitting a TypeScript compile error with setting the `.value` of a `UUIFormControlMixin` based component,
e.g. `Type 'string | File | FormData' is not assignable to 'ValueType | DefaultValueType'`

## Types of changes


iOvergaard (approved):
Your story works, but the built version of Storybook does not. I have to figure out what is wrong there.
image

PR #1605, opened 2024-04-15
Feature: Document Type Create Actions

leekelleher:
## Description

Adds the Create Actions for Document Types, e.g. "Document Type with Template", "Element Type", etc.

Create Actions for Document Type

> [!NOTE]
> After a good discussion with @nielslyngsoe, we agreed to leave the "Composition" option out, as this should be open for further discussion, on how to better surface the Compositions concept and usage.

A concept of a _preset alias_ is used in routable path, this can set the icon and enable `isElement` flag.

With "Document Type with Template", a `createTemplateMode` flag is set, to create the template on the submit event, then assign the ID to the allowed templates, then save the document-type.

## Types of changes


iOvergaard (approved):
It works as intended. I have suggested directly to Lee to build in some checks before trying to create the template to see, if we can do anything to prevent it from being created if it already exists or if there was an error saving the document type.

PR #1606, opened 2024-04-15
Bugfix: Color Picker prefix hex values with a hash

leekelleher:
## Description

This PR fixes a bug introduced in PR #1443, where the color hex value is prefixed with a hash inside the `umb-input-color` component. The reason behind this was that the data-type configuration for the Color Picker property-editor did not prefix the color hex values with a hash, _(this may or may not be a bug in the `umb-property-editor-ui-color-swatches-editor` component)_.

In order to keep the `umb-input-color` component "pure", the logic has been moved to the Color Picker property-editor itself, by ensuring the `"items"` configuration (color swatches) and the `value` are prefixed with a hash.

## Types of changes

### DIFF
--- a/src/packages/core/components/input-color/input-color.element.ts
+++ b/src/packages/core/components/input-color/input-color.element.ts
@@ -1,7 +1,7 @@
 import { html, customElement, property, map, nothing } from '@umbraco-cms/backoffice/external/lit';
-import { UUIFormControlMixin } from '@umbraco-cms/backoffice/external/uui';
 import { UmbChangeEvent } from '@umbraco-cms/backoffice/event';
 import { UmbLitElement } from '@umbraco-cms/backoffice/lit-element';
+import { UUIFormControlMixin } from '@umbraco-cms/backoffice/external/uui';
 import type { UmbSwatchDetails } from '@umbraco-cms/backoffice/models';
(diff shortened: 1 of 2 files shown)


bjarnef:
@leekelleher does it allow the color to be rba(a) too and color names, e.g. `red`?

iOvergaard (approved):

leekelleher:
@bjarnef Do you mean in the property-editor's configuration or when using the `umb-input-color` component directly?

In `umb-property-editor-ui-color-picker` (property-editor), **no**, it's hex color values only, (because the `umb-property-editor-ui-color-swatches-editor` is designed that way).

In `umb-input-color`, **yes**, you can use any valid CSS color value you want. 👍

bjarnef:
@leekelleher mainly regarding the predefined colors as in the current backoffice, but I just checked the current behaviour where `red` resolve the hex code and I can ensure `rgba(0,0,0,0.5)` but this is enforced to `#000000`.

image

I think it would be great if this supported `rgba` values as well.

leekelleher:
Thanks @bjarnef, I wasn't aware that the current backoffice (v13) would let you input the `rgba()` values. For v14, it sounds like a bug with the `umb-property-editor-ui-color-swatches-editor`. We can pick up on that during the RC phase, _(it's all hands on deck for this week's RC release!)_

PR #1615, opened 2024-04-15
Chore: Remove unused store connector for trees

madsrasmussen:
## Description

Trees are manually updated, so the store connector doesn't do anything for the tree store. This PR removes all the store connector code for trees.

## Types of changes


leekelleher (approved):
I was unsure if there was anything specific to test with this. I've tested out each of the trees (based in the commits) and they all appear to be working as previously. 👍

PR #1616, opened 2024-04-15
Bugfix: Check for tree root items

madsrasmussen:
## Description

Check for tree root items

## Types of changes


leekelleher (commented):
Tested out, it works as intended, but it's quite chatty with the server calls, e.g. in the Settings section there were 9 additional calls.

I know that we'll be adding in stores to cache the data locally, but wasn't sure if this would be in time for RC001?

Otherwise, I'm happy to press approve. 👍

madsrasmussen:
Yeah, it will add some additional calls. I think this is OK for the RC. I am not to worry about these calls as there is not much we can do about it at the moment. It is all the js loading that is the real problem and we should look into that first.

leekelleher (approved):

PR #1617, opened 2024-04-15
Feat: bind server feedback with validation system

nielslyngsoe:
Simple first version of server validation binding to the client validation system.

This means properties of a Document can get the validation, but anything more complex than such is not enriched, only the property will know if it contains a validation error.

## Types of changes

## Checklist


leekelleher (approved):
Tested out with Document, with mandatory properties. Validation works as expected.

I'd noticed that the server returns a localization key as `#validation.invalidEmpty` (with a dot/period delimiter), whereas we use an underscore for client-side localization keys. Not a showstopper for this PR, but whether the server should modify their values?

PR #1623, opened 2024-04-16
Bugfix/storybook-extensions-registry-error

iOvergaard:
This fixes the error where Storybook failed to initialize because we took on an instance before it existed.


leekelleher (approved):
Tested out locally, all builds & runs the preview as expected. 👍

PR #1628, opened 2024-04-16
Feature/public-entrypoints

iOvergaard:
## Description

It has been possible to divide packages into "public" and "private" for a while now, where the public ones don't require you to be authenticated, which is ideal to register external login providers and such, but if you try and register an entryPoint or a bundle in a "public" package, it still won't be loaded until the umb-backoffice element is visible on the page. This is because this element is the initializer of all entrypoints and bundles. And it won't be visible until you are logged in defeating the purpose of trying to register entrypoints publically, which you may want to do, if you for some reason would like to change the configuration depending on external circumstances.

This PR adds a new type of entry point module called `appEntryPoint`, which will be initialized in the UmbAppElement. This ensures that appEntryPoints may have logic that runs both before and after authentication for example.

UmbAppElement = `appEntryPoint` and `bundle`
UmbBackofficeElement = `entryPoint`

The initializers are now also based on an abstract base initializer class to share the observer logic and make sure the instantiate and unload methods are called appropriately. Additionally, there is now a new exportable function on entry point modules called `onUnload` which is registered, if the entry point module is ever being unregistered by something else. This way, an entry point can listen for `destroy()` on its host element as well as implement its logic to run in case it's ever being unregistered.

The instance of an entrypoint is being saved on the initializer to make sure we don't accidentally run its side-effects twice.

This showcases what we need to be able to do to for example register dynamic external login providers. It's up for debate if it's the right name of the function, or if we should do something else entirely.

## How to test

Register a public package with an entrypoint and try and let that entrypoint register an auth provider:

```json
// umbraco-package.json
{
  "$schema": "../../umbraco-package-schema.json",
  "name": "My Auth Package",
  "allowPublicAccess": true,
  "extensions": [
    {
      "type": "appEntryPoint",
      "alias": "My.AuthProvider.EntryPoint",
      "name": "My Auth Provider Entry Point",
      "js": "/App_Plugins/ExternalLogin/ext-entrypoint.js"
    }
  ]
}
```

```js
// ext-entrypoint.js
export const onInit = (host, registry) => {
  console.log('public onInit', host);

  registry.register({
    type: 'authProvider',
    name: 'Facebook',
    alias: 'My.AuthProvider.Facebook',
    forProviderName: 'Facebook',
    meta: {
      label: "Sign in with Facebook",
      defaultView: {
        icon: "icon-facebook"
      }
    }
  })
}
```

Now click the Sign out button and check that you see the fake "Facebook" provider.


src/libs/extension-api/types/manifest-entrypoint.interface.ts
```diff
@@ -7,4 +7,12 @@ import type { ManifestPlainJs } from './base.types.js';
  */
 export interface ManifestEntryPoint extends ManifestPlainJs<UmbEntryPointModule> {
 	type: 'entryPoint';
+
```
nielslyngsoe, line 10: This does not seem like its used anywhere. :-) Am I right about that?

src/libs/extension-api/types/manifest-entrypoint.interface.ts
```diff
@@ -7,4 +7,12 @@ import type { ManifestPlainJs } from './base.types.js';
...
+	/**
+	 * The scope of the entry point. If global, the entry point will be loaded on the root host and never be destroyed.
+	 * If local, the entry point will be loaded on the BackofficeElement and will be destroyed when the host is removed for example when signing out.
+	 * @default 'local'
+	 * @enum ['global', 'local']
+	 */
+	scope?: 'global' | 'local';
```
nielslyngsoe, line 17: This does not seem like its used anywhere. :-) Am I right about that?

nielslyngsoe: But I do think this could be an interesting concept, so lets have a chat about that :-)

nielslyngsoe (approved):
Cool, I'm very happy that you took the effort to refactor the base for all three extension types. Very nice!

PR #1632, opened 2024-04-16
Bugfix: Logviewer: removes the `decodeURIComponent`

leekelleher:
## Description

In the Log Viewer's Saved searches, any queries clicked on that contained a percentage sign `%` would throw an malformed URI error, this was because the value was being double-decoded, (since the router-slot `query()` function we're using already handles the URL decoding).

This PR removes the extra `decodeURIComponent` call, so clicking on the saved searches (that contain percentage signs) now work.

## Types of changes


madsrasmussen (approved):
Looks good to me

PR #1635, opened 2024-04-16
Bugfix: Persisting user group start nodes

madsrasmussen:
## Description

Fixes the User Group start node UI and model updates so the data gets persisted on the server. I have aligned the UI for start nodes with the languages so we now use a toggle to allow access to everything.

## Types of changes

## How to test?

* Make sure that you can set, update and remove document and media start nodes on a user group.

### DIFF
--- a/src/assets/lang/en-us.ts
+++ b/src/assets/lang/en-us.ts
@@ -1893,6 +1893,8 @@ export default {
 		searchAllChildren: 'Search all children',
 		languagesHelp: 'Limit the languages users have access to edit',
 		allowAccessToAllLanguages: 'Allow access to all languages',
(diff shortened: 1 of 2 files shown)


leekelleher (commented):
I've found a slight bug with this. If you have the "allow access to all" toggles disabled, then select a content start node, then enable the "allow access to all languages", it will clear out the content start node value.

I've took a look at the code, but couldn't spot anything obvious that would do this. 🤔

madsrasmussen:
Good catch. That took me some time to discover. The problem was a duplicated observer alias. I did find another bug along the way, so that is also fixed.

leekelleher (approved):
All good now! 🙌

PR #1655, opened 2024-04-18
Add localization files

liamlaverty:
Adds localization files from the existing Umb v13 repo, updated into the new TS/JSON format. 

There's a unique commit for each localization to make it easier to cherry-pick/reject. I've not migrated the two existing files `da-dk.ts` and `en-us.ts`.

## Description

I wrote a [console app](https://github.com/liamlaverty/umb-localization-migrator/) which takes the [existing XML localization files in the v13 repo](https://github.com/umbraco/Umbraco-CMS/tree/v13/contrib/src/Umbraco.Core/EmbeddedResources/Lang) and migrates them into the new typescript format outlined in the Storybook's [localization section](https://ambitious-stone-0033b3603.1.azurestaticapps.net/?path=/docs/api-localization-intro--docs). 

This technique will miss a lot of properties. Any new features added into v14, or any features where the localization alias has changed will unfortunately not be migrated.  To help with this, in the console app's repo, there's a record of the properties which haven't been migrated for each language in the [reports folder](https://github.com/liamlaverty/umb-localization-migrator/tree/main/IOFiles/DiffFinder/reports). For example, if a property existed in the V14 en-US file, but didn't exist in the V13 fr-fr file, that is detailed in [fr-fr-migration-report.md](https://github.com/liamlaverty/umb-localization-migrator/blob/main/IOFiles/DiffFinder/reports/fr-fr-migration-report.md).

## Types of changes

- [ ✅ ] New feature (non-breaking change which adds functionality)
- [  ] Chore (minor updates related to the tooling or maintenance of the repository, does not impact compiled assets)

## Motivation and context

V13 has a lot of localization files, which are currently not ported to V14

## How to test?

- Change the back office language to fr-fr
- Attempt to copy any entity in the backoffice
- The text should read "Copier" instead of "Copy"

## Checklist

- [ ✅ ] I have read the **[CONTRIBUTING](<(https://github.com/umbraco/Umbraco.CMS.Backoffice/blob/main/.github/CONTRIBUTING.md)>)** document.


iOvergaard:
Hi @liamlaverty 

This is so cool...!

I'm inclined to merge it and take it from there. I just wanted to check one thing and that is if the 'en-gb' file really should be called 'en.ts' instead. It's a bit finicky but I _think_ the default language in the old Backoffice was 'en-us' but if for some reason the variable didn't exist, it would fall back to 'en.xml', which is en-GB.

I know that our localization manager first will check with a region code (like en-GB) and then a secondary check without (like 'en'), so I'm wondering if we should register the language as simply 'en'.

Also wondering if we should change this line from 'en-US' to 'en':

https://github.com/umbraco/Umbraco.CMS.Backoffice/blob/feature/public-entrypoints/src/libs/localization-api/localization.manager.ts#L30

Lastly, we need to register these languages in the manifest file found here: https://github.com/umbraco/Umbraco.CMS.Backoffice/blob/feature/public-entrypoints/src/packages/core/localization/manifests.ts

What do you think?

liamlaverty:
Hey @iOvergaard,

I've updated the manifest so that it's got references to the new files in [3da712c](https://github.com/umbraco/Umbraco.CMS.Backoffice/pull/1655/commits/3da712cc972a188dfeb2f61d22390d4270879fe8). 

In  [6643b34](https://github.com/umbraco/Umbraco.CMS.Backoffice/pull/1655/commits/6643b34bbf7681f3869807aa7cee40c582e8a0bc) I've updated so that the header comments have both the international name and the local name for each language. 

I'm happy to defer to your/HQ's judgement on the `en`/`en-us`/`en-gb` considerations. I'm not certain which result most end-users would expect or prefer.

iOvergaard (approved):
I see from the generated reports that we have a lot of syncing to do on the language files still. I am going to approve this so we can start working on the files.

PR #1656, opened 2024-04-18
Feature: icons extension

nielslyngsoe:
Ability to add icons as a extension type.

Also makes sure we support all color-aliases from v.13 but using new colors, this does mean currently that some colors are really not the color they had in v.13, like orange is brown. But lets see the push back, the old not well converted colors are marked as legacy and cannot be re-picked.
In other words, not the legacy colors are horrible, but we can always improve later, now they at least are defined so its easier, also as a community member to correct this.

## Types of changes

## Checklist


iOvergaard (changes requested):
Looks all right, but there is a build error where it tries to load the file you deleted.

src/packages/core/icon-registry/icon.registry.ts
```diff
@@ -13,13 +12,21 @@ interface UmbIconDescriptor {
...
+		this.#icons = [...this.#icons, ...icons];
+	}
 	/**
 	 * @param {string} iconName
 	 * @return {*}  {boolean}
 	 * @memberof UmbIconStore
 	 */
 	acceptIcon(iconName: string): boolean {
-		const iconManifest = icons.find((i: UmbIconDescriptor) => i.name === iconName);
+		const iconManifest = this.#icons.find((i: UmbIconDescriptor) => i.name === iconName);
```
iOvergaard, line 29: I think because this isn't an async method, but the icons are loaded asynchronously, you will end up with cases where an icon is loaded after it has been requested and so will never show up. There seems to be no check in this registry to check whether the observer in UmbIconRegistryContext has been run or not, or is there?

iOvergaard (changes requested):
Icons load in the backoffice, it seems, but I can't get them to load before you are logged in (for instance in the login provider screen). You can force this view by registering an extension somewhere with `type: 'authProvider'` and you should see the screen when accessing the backoffice.

src/packages/core/icon-registry/icon.registry.ts
nielslyngsoe: Oh, yes this is a problem. I will see what our options are.

src/packages/core/icon-registry/icon-registry.context.ts
```diff
@@ -0,0 +1,49 @@
...
+	//#host: UmbControllerHost;
+	#registry: UmbIconRegistry;
+	#manifestMap = new Map();
+	#icons = new UmbArrayState<UmbIconDefinition>([], (x) => x.name);
+	readonly icons = this.#icons.asObservable();
+	readonly approvedIcons = this.#icons.asObservablePart((icons) => icons.filter((x) => x.legacy !== true));
+
+	constructor(host: UmbControllerHost) {
+		super(host, UMB_ICON_REGISTRY_CONTEXT);
+		//this.#host = host;
```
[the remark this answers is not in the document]
nielslyngsoe: Good one, done.

src/apps/app/app.element.ts
```diff
@@ -79,7 +78,6 @@ export class UmbAppElement extends UmbLitElement {
 
 		OpenAPI.BASE = window.location.origin;
 
-		new UmbIconRegistry().attach(this);
```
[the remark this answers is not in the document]
iOvergaard: I think we would have the same problem if we wanted to show icons in the installer for example 👀

src/apps/backoffice/backoffice.element.ts
```diff
@@ -10,33 +10,32 @@ import { UmbLitElement } from '@umbraco-cms/backoffice/lit-element';
...
 import './components/index.js';
 
-// TODO: temp solution to load core packages
 const CORE_PACKAGES = [
-	import('../../packages/audit-log/umbraco-package.js'),
-	import('../../packages/block/umbraco-package.js'),
-	import('../../packages/data-type/umbraco-package.js'),
-	import('../../packages/dictionary/umbraco-package.js'),
+	import('../../packages/umbraco-news/umbraco-package.js'),
+	import('../../packages/core/icon-registry/umbraco-package.js'),
```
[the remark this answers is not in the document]
nielslyngsoe: Thanks. That make good sense.

src/packages/core/icon-registry/icon.registry.ts
nielslyngsoe: There is a fix now.

src/apps/app/app.element.ts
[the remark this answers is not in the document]
nielslyngsoe: Good catch

src/packages/core/icon-registry/icon.registry.ts
```diff
@@ -13,26 +12,68 @@ interface UmbIconDescriptor {
  * @description - Icon Registry. Provides icons from the icon manifest. Icons are loaded on demand. All icons are prefixed with 'icon-'
  */
 export class UmbIconRegistry extends UUIIconRegistry {
+	#initResolve?: () => void;
+	#init: Promise<void> = new Promise((resolve) => {
+		this.#initResolve = resolve;
+	});
+
+	#icons: UmbIconDescriptor[] = [];
+	#unhandledRequests: { name: string; provider: UUIIconHost }[] = [];
```
iOvergaard, line 21: Could this be a `Set` or do we need to handle each request individually for each icon host?

src/packages/core/icon-registry/icon.registry.ts
```diff
@@ -13,26 +12,68 @@ interface UmbIconDescriptor {
...
+		}
+		// Go through the new icons and see if they are already requested.
+		const newIcons = this.#icons.filter((i) => !oldIcons.find((o) => o.name === i.name));
+		newIcons.forEach((icon) => {
+			// Do we already have a request for this one
+			const unhandled = this.#unhandledRequests.find((i) => i.name === icon.name);
+			if (unhandled) {
+				if (unhandled.provider) {
+					this.#loadIcon(icon.name, unhandled.provider).then(() => {
+						this.#unhandledRequests.filter((i) => i.name !== icon.name);
```
iOvergaard, line 38: I dont believe that Array.prototype.filter replaces in-place. According to MDN, it produces a shallow copy of the elements that pass the test. I believe we need to do something like this:


```suggestion
						this.#unhandledRequests = this.#unhandledRequests.filter((i) => i.name !== icon.name);
```

src/packages/core/icon-registry/icon.registry.ts
```diff
@@ -13,26 +12,68 @@ interface UmbIconDescriptor {
...
+			this.#initResolve();
+			this.#initResolve = undefined;
+		}
+		// Go through the new icons and see if they are already requested.
+		const newIcons = this.#icons.filter((i) => !oldIcons.find((o) => o.name === i.name));
+		newIcons.forEach((icon) => {
+			// Do we already have a request for this one
+			const unhandled = this.#unhandledRequests.find((i) => i.name === icon.name);
+			if (unhandled) {
+				if (unhandled.provider) {
```
iOvergaard, line 36: Consider combining the two if-statements into one: `if (unhandled?.provider)`.

Also a question: Can `provider` ever be undefined?

iOvergaard (changes requested):
It works with the umb-app element and everywhere else now, great! Left a few comments for optimization in the icon registry :)

src/packages/core/icon-registry/icon.registry.ts
nielslyngsoe: Oh yes, of course, thanks!

src/packages/core/icon-registry/icon.registry.ts
nielslyngsoe: Yes, good spotting. I really dont think so. I have removed the last check. I must admit it looks like I solved the case, but never got around to clean up the code.

src/packages/core/icon-registry/icon.registry.ts
nielslyngsoe: I would say it could be a Map. So I've updated that :-)

nielslyngsoe:
@iOvergaard thanks for second round review. Lets try a third round :-D

iOvergaard (approved):

PR #1669, opened 2024-04-19
Feature: Data-type created via Content-Type Property Flow gets a generated name

nielslyngsoe:
Get a generated name when creating data-types via the Content-Type Editor.
The name is based on DocumentType Name + Property Type Label + Property Editor Name.
image

This feature most likely should be revisited to make the most sense, but this solves the technicalities and is better than none.

## Types of changes

## Checklist


nielslyngsoe:
@leekelleher I investigated a bit, its because the server does not have any editorUiAlias stored out of the box, and we are filterings on that in this expeience. So there is a missing link, cause we own the `defaultPropertyEditorUiAlias`in our manifests, meaning the server could never respond correctly. Maybe it could be solved if we could filter on both the given editorUiAlias + non defined.. and then have the server also provide the editorAlias, so we could look up the `defaultPropertyEditorUiAlias` in our aliases to get the data to fulfil the filtering...
Anyhow, thats another task we need to look into.. you can test this now by finding your Editor by typing its name, or create a new datatype first. :-)

PR #1688, opened 2024-04-23
Feature: Add /error route

iOvergaard:
## Description

We already have the app error element, so this essentially maps out a new route called /error to show that element. The error element itself has then been updated to take care of a few route parameters that can occur for instance when external login fails in a number of scenarios such as login, linking, and un-linking. The error codes follow the RFC for OAuth2.0 + a few more added that Umbraco itself uses.

## How to test

The backend isn't entirely linked up yet, but you can navigate to something like http://localhost:5173/error?flow=external-login&error=access_denied&error_description=You%20dont%20have%20access&error_uri=https://umbraco.com to check that an error shows up and is localized.


leekelleher (approved):
Tested out, looks good. 👍 

I've added a CSS rule for the `<summary>` to use a pointer cursor, (to indicate an interaction).

PR #1689, opened 2024-04-23
Chore: Clean up UI around two-factor

iOvergaard:
## Description

Minor fixes to the UI of two-factor providers. Removed an unused, obsolete element. Updated the mock handlers for manifest endpoints to match the server.


leekelleher (approved):
Tested out, all good. 👍

PR #1693, opened 2024-04-23
Feature: Media picker modal

loivsen:
## Description

Added Media picker modal.

Other changes:
- RTE uses Media Picker rather than Media Tree Picker.
- Dropzone can now use the browse function from the uui file dropzone.
- Dropzone now takes in a parentUnique for uploads.

Missing:
- Media picker modal search function
- Thumbnail of images in media picker

## Types of changes


src/packages/media/media-types/utils/index.ts
```diff
@@ -1,13 +1,16 @@
+// They are currently hardcoded on the backend. Go by GUIDs in case they get renamed..?
```
nielslyngsoe, line 1: Hi @loivsen I know this is not obvious, but I want you to be aware that you can make your own Folder type and/or Image types. So we cannot identify them by either Alias or GUID.
Also implementors can change the file-type/ext of the files a Type accepts. meaning the ImageType could be reconfigured to hold PDF. Or the Video one could be allowed as a folder.
In other words we can not be clever about the types.
Also these preconfigured types was introduced some time in v.8, meaning projects upgraded from before that version has another set of MediaTypes pre-configured, and again there is no guarantee that any of the pre-configured stuff is there any more or is like we expect it to be. So I would encourage to make the code unaware about any of this, we should not, in code, know anything about the types. but make it work with the types of the server.
I hope that makes sense :-)

I will turn this PR into a draft, cause it requires a different approach to work as expected.

nielslyngsoe: Update: Lone and I had a voice-call going through the options and what the potential solutions could be. :-)
Lone is exploring the options

PR #1695, opened 2024-04-23
Chore: Move property editor uis to standalone package

madsrasmussen:
## Description

Move all "common" Property Editors from the core to a standalone package. We didn't export anything from the property editor UIs so this should be a safe move without any breaking changes. This first PR only includes the uis. Schemas will be moved as part of another PR.

## Types of changes


leekelleher (approved):
Tested against Kenn's comprehensive database, going through each of the property-editors, all works as expected. 🚀

PR #1698, opened 2024-04-24
Chore: move property editor schemas into each property editor folder

madsrasmussen:
## Description

This PR moves the property editor schema into the corresponding property editor folder. The schemas aren't exported anywhere so the move is safe with no breaking changes.

The PR exploded a bit in size because I found a need to add type safety to exported manifests. The change is very simple but it included a lot of files.

## Types of changes


leekelleher (approved):
Tested against Kenn's comprehensive database, going through each of the property-editors, all works as expected. 🚀

PR #1708, opened 2024-04-25
Make input-content-type-property element local

madsrasmussen:
## Description

The component has been moved out of the global core components. The element makes the core package depend on `document-type`, `media-type` and `member-type` modules which it shouldn't do. The core can only depend on libs and externals.

The `umb-input-content-type-property` is a local component of the configuration and should not be used by others.

## Types of changes

> [!CAUTION]
> Removes global load of the `umb-input-content-type-property` element. The element shouldn't be used anywhere else.


nielslyngsoe (approved):
Dangerous, but I agree, it will be hard to guarantee the usage of this element, So I would also advice not using this in packages. So good to get it away. Ideally we do pre-fix it namewise with Collection. to emphasise it, cause it sounds like a generic one you can use everywhere. I can make a PR for the renaming.

PR #1710, opened 2024-04-25
Bugfix: Member Picker property-editor, `max` = 1

leekelleher:
## Description

Corrected the Member Picker property-editor to have a maximum of 1 selectable item, for backwards-compatibility with previous Umbraco versions.

## Types of changes

### DIFF
--- a/src/packages/members/member/components/input-member/input-member.element.ts
+++ b/src/packages/members/member/components/input-member/input-member.element.ts
@@ -1,6 +1,6 @@
 import type { UmbMemberItemModel } from '../../repository/index.js';
 import { UmbMemberPickerContext } from './input-member.context.js';
-import { css, html, customElement, property, state, repeat } from '@umbraco-cms/backoffice/external/lit';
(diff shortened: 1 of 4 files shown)


bjarnef:
@leekelleher does it mean the member picker is single picker by default?
In the current backoffice I found it odd that some pickers a single picker and other multiple picker by default. In fact I think member picker (not MNTP) never supported multiple members.

bjarnef:
Okay, I thought it was a good chance to streamline the pickers although it may be a breaking change from previous versions. Couldn't a migration handle it setting `max = 1` for existing datatype instances?

Yes, MNTP supports multiple members although I wish when only a single type can be picked, the PVC returned `T` or `IEnumerable<T>` instead of `IPublishedContent` and `IEnumerable<IPublishedContent>` but that's another [discussion](https://github.com/umbraco/Umbraco-CMS/discussions/15266).

iOvergaard (approved):

PR #1716, opened 2024-04-26
Bugfix: Remove block circular dependency

madsrasmussen:
## Description

The core package is currently depending on the block package. It uses an import to `@umbraco-cms/backoffice/block` which it cannot. This PR is a temporary solution to solve the block circular dependency.

The PR moves the block condition types from the block package to the extension registry module. It is not the ideal solution but as long as we have the type unions this is the best we can do.

## The result

You will no longer find imports from the core package to the `@umbraco-cms/backoffice/block` module.

## Types of changes

> [!CAUTION]
> Breaking change
> import of `BlockEntryShowContentEditConditionConfig` and `BlockWorkspaceHasSettingsConditionConfig` is moved from `@umbraco-cms/backoffice/block` to `@umbraco-cms/backoffice/extension-registry`;


nielslyngsoe (approved):
looks good

PR #1719, opened 2024-04-26
Bugfix: Remove History Item circular dependency

madsrasmussen:
## Description

The History Item component depends on the `APP_CONTEXT` to render the avatar. I have changed the component to slot in the avatar element instead. This is closer to how other generic components work. This also moves the `APP_CONTEXT` responsibility to the implementing part.

On a side note. We shouldn't rely on the `serverUrl` to create the full avatar URL. This can change on the server. I have asked for a change in the Management API to include the absolute URL as we have in other places that expose the avatar URLs.

## Types of changes

> [!CAUTION]
> Breaking Change
> It is no longer possible to pass a `src` prop to the `umb-history-item` element to render an avatar. Use the avatar slot instead.

### DIFF
--- a/.github/workflows/build_test.yml
+++ b/.github/workflows/build_test.yml
@@ -34,7 +34,7 @@ jobs:
           node-version: ${{ matrix.node-version }}
           cache: 'npm'
       - run: npm ci --no-audit --no-fund --prefer-offline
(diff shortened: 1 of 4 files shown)


nielslyngsoe (approved):
looks good

PR #1720, opened 2024-04-26
Feature: Login without losing changes

iOvergaard:
## Description

> [!NOTE]
> When this PR is merged, we should merge https://github.com/umbraco/Umbraco-CMS/pull/16194 on the backend right away

This allows the Backoffice to use any OpenId provider (Umbraco backend included) without losing your changes if, say, a timeout should occur in the middle of you editing anything.

**How does it work?**
This PR enables all auth providers to be shown in a popup and it works in general by sending the user back to the `/oauth_complete` route, which initializes the PKCE code -> token exchange. Every time such an exchange happens, the local storage is updated. All open tabs listen for the `storage` event and react if there is a change to the token. If a change is detected, the tab will either load that into memory or it will close itself (if it has a `window.opener`). This allows us to redirect back to the backoffice inside the popup, which will close itself, and it allows any open tabs to reinitialize their token without throwing the user away from their active work.

**When does it activate?**
There are two cases when it works; if more than one `authProvider` is detected, we will show the auth provider selection screen right from the start, and any clicks on the auth provider will open in a popup.

https://github.com/umbraco/Umbraco.CMS.Backoffice/assets/752371/5b19e196-a0cf-4342-a363-22ed248e790e

The second case is when a user times out, which can happen if their token is revoked (and they make a request) or the token expires. In that case, the same modal will open and let the user log in inside a popup.

**Whatever else changed?**
Recently, we introduced a new kind of entrypoint called `appEntryPoint`, which loads before the backoffice itself. We had to move that a bit down the stack and only initialize after the call to the `/manifest/manifest/public` has run, so that we know when they load. The reason for this is that these entrypoints could and should be used to register auth providers, and we really need to make sure they have been registered before trying to initialize anything.

## How to test

1. Test that you can log in like normally with the backend
2. Set up another auth provider
3. Test that you can log in
4. Wipe the tokens in the database (or wait for timeout) and test that you can log in again without losing what you were doing

You might need these appsettings in `Umbraco:CMS:Security` if you are testing from a Vite client:

```json
      "Security": {
        "BackOfficeHost": "http://localhost:5173",
        "AuthorizeCallbackPathName": "/oauth_complete",
        "AuthorizeCallbackLogoutPathName": "/logout",
        "AuthorizeCallbackErrorPathName": "/error"
      }
```


leekelleher (commented):
Tested it out, (with the corresponding backend PR https://github.com/umbraco/Umbraco-CMS/pull/16194).

Good news is that the login/authentication parts work as expected. 👍 

Bad news is the timeout/losing changes part doesn't work as I expected.
e.g. if I edit a document, enter new text into a textarea, then leave the browser untouched (for longer than the timeout, in my case 4 minutes), nothing happens. I then tried navigating elsewhere, and then I'm redirected to the login screen. I log back in and I'm navigated to the Content Welcome dashboard, (not on the document where I was previously).

Next attempt, (edit a textarea in a document), let the timeout expire, then press Save... I'm redirected to the login screen, I log back in and navigated to the Content Welcome dashboard again. (I go to the document and my textarea amends are gone.)

I'm unless whether I'm misunderstanding how the feature should work?

leekelleher (approved):
Re-tested with @iOvergaard, the PR working as expected.

The issue I encountered is still a concern, but with how the Document Save actions are handled, which can be investigated in a separate PR.

PR #1746, opened 2024-05-01
Feature: Property Editor Config Value Converter

leekelleher:
## Description

A proposal to add a `converter` function parameter to the `UmbPropertyEditorConfigCollection.getValueByAlias<T>()` method.

This is for scenarios where a configuration value can be converted from `unknown` into a target object-type. e.g. to parse, validate, return or to potentially set a sensible fallback/default value.

Consider the general use-case for minimum/maximum configurations, typically these would be positive integers, so would need to ensure that negative values can't leak through.

e.g. to ensure a maximum value...
```
config.getValueByAlias('max', (value) => { const num = Number(input); return num > 0 ? num : Infinity; });
```

I'm open to alternative approaches on how to parse/converter the config values.

## Types of changes

### DIFF
--- a/src/packages/core/property-editor/config/property-editor-config-collection.class.ts
+++ b/src/packages/core/property-editor/config/property-editor-config-collection.class.ts
@@ -13,11 +13,15 @@ export class UmbPropertyEditorConfigCollection extends Array<UmbPropertyEditorCo
 		return Array;
 	}
 
-	getValueByAlias<T>(alias: string): T | undefined {
+	getValueByAlias<T>(alias: string, converter?: (value: unknown) => T): T | undefined {
 		const property = this.getByAlias(alias);
 
 		if (property?.value === undefined || property?.value === null) {
 			return;
+		}
+
+		if (converter) {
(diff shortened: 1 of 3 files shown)


iOvergaard (commented):
It looks all right to me. I'm not sure how widespread the usage would be if, say, you had a single value, you could convert it then and there, but it doesn't hurt to have.

Now that we have two parameters in the function, could I get you to add jsdoc for the parameters with some examples inlined in the docs?

leekelleher:
> It looks all right to me. I'm not sure how widespread the usage would be if, say, you had a single value, you could convert it then and there, but it doesn't hurt to have.

So far, it'll be for the minimum/maximum config values (which applies to all the picker editors). I'll give it some thought.

> Now that we have two parameters in the function, could I get you to add jsdoc for the parameters with some examples inlined in the docs?

Okay, will do.👍

leekelleher:
I'm going to close this off, as the usage would be limited, and config value conversion can be handled within the property-editors.

PR #1755, opened 2024-05-02
Umbraco-CMS/issues/16205 - multiple black color options in icon picker

liamlaverty:
Filters the list of `umbracoColors` in `src\packages\core\resources\extractUmbColorVariable.function.ts` so that users are only presented with unique colours.

https://github.com/umbraco/Umbraco-CMS/issues/16205

## Description

In the icon picker, users are currently offered two `black` options, as the file `src\packages\core\resources\extractUmbColorVariable.function.ts` contains two entries with identical `varName` properties: 

```typescript
export const umbracoColors = [
	{ alias: 'text', varName: '--uui-color-text' },
	{ alias: 'black', varName: '--uui-color-text' },
```

I've added a function to get the same umbracoColor set, but filtered to be unique by `varName`: 

```typescript
/** 
 * Gets `umbracoColors` unique by their `varName` property
 */
export function getUniqueUmbracoColors() {
	const uniqueBy = 'varName'
	const colorVariablesUniqueBy_varName = [...new Map(umbracoColors.map(item => 
		[item[uniqueBy], item]
	)).values()];
	return colorVariablesUniqueBy_varName;
}
```

## Types of changes

- [ ✅ ] Breaking change (fix or feature that would cause existing functionality to change)

This may cause users upgrading from RC2 to RC3 to encounter errors, if for example they've selected `umbracoColors.black`, and it's filtered out of the list. 

## Motivation and context
327355868-593e3217-cc38-47a1-934e-2f02db1dc063

## How to test?

Open a _Select Icon_ window, and check the first two colours are `text` then `yellow`, instead of `text` then `black`

## Checklist

- [ ✅ ] I have read the **[CONTRIBUTING](<(https://github.com/umbraco/Umbraco.CMS.Backoffice/blob/main/.github/CONTRIBUTING.md)>)** document.

### DIFF
--- a/src/packages/core/modal/common/icon-picker/icon-picker-modal.element.ts
+++ b/src/packages/core/modal/common/icon-picker/icon-picker-modal.element.ts
@@ -5,7 +5,7 @@ import { UmbTextStyles } from '@umbraco-cms/backoffice/style';
 
 import type { UmbIconPickerModalData, UmbIconPickerModalValue } from '@umbraco-cms/backoffice/modal';
 import { UmbModalBaseElement } from '@umbraco-cms/backoffice/modal';
-import { extractUmbColorVariable, umbracoColors } from '@umbraco-cms/backoffice/resources';
+import { extractUmbColorVariable, getUniqueUmbracoColors } from '@umbraco-cms/backoffice/resources';
 import { umbFocus } from '@umbraco-cms/backoffice/lit-element';
 import { UMB_ICON_REGISTRY_CONTEXT, type UmbIconDefinition } from '@umbraco-cms/backoffice/icon';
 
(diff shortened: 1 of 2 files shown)


iOvergaard:
We must be a bit careful about the color aliases since I believe some of them are saved in the database for migrated content. If I'm reading this correctly, it would mean that the #000000 color would be called "text" and not "black", and we need to investigate what the alias has to be. 

I'm also unsure if we can't just remove the "text" color altogether. Maybe @nielslyngsoe can give us an answer to that?

iOvergaard:
Thanks for this proposed fix. However, the 'text' color was not supposed to be there, and other colors were missing. I have taken the liberty to fix it in another PR that introduces the missing colors: https://github.com/umbraco/Umbraco.CMS.Backoffice/pull/1794

PR #1758, opened 2024-05-02
remove the unused 'allowDecimals' configuration and the "Integer" property editor

### LINKED ISSUE #1660: [BUG]: Decimal & Numeric property editor with "allow decimals" set to false still allows decimal values
**Describe the bug**
On the Decimal property editor, when **Allow Decimals** is set to **false**, a user can still enter decimals. 

The same bug exists in both the `Numeric` and `Decimal` property editors

**To Reproduce**
1. Fresh install the V14 RC
2. Create a document type with a new Decimal property editor
3. Configure its properties:
  - Minimum: 0
  - Maximum: 10
  - Step Size: 0.01
  - Allow Decimals: false
4. Save the property & docype
5. Create a new piece of content using the doctype
6. Set the decimal value to `1` (or any whole integer)
7. Press the `.` character, this should set the cursor to the left of the `1`
8. Press the `.` character a second time, this should add a `0.` in front of the `1`
  - The `input` is correctly highlighted red
10. Clicking save and publish incorrectly saves the document with `0.1`, reloading the page returns `0.1`

**Expected behavior**
`Allow Decimals` being set to `false` should throw an error on **Save and publish** if a decimal is on the property. 

In the `Decimal` property editor creator, a validation step should check that "step size" is non-decimal if "allow decimals" is set to `false`

**Screenshots**

**Config**
Screenshot 2024-04-18 at 14 05 52

**Decimal content in the disallow decimal property**
Screenshot 2024-04-18 at 14 05 46

**Desktop (please complete the following information):**
 - OS: macOS Sonoma 14.2.1 (23C71)
 - Browser MS Edge
 - Version 122.0

**Additional context**
Tested on RC-14

iOvergaard:
this aligns the "numeric" and "decimal" property editors with V13

fixes #1660

### DIFF
--- a/src/mocks/data/data-type/data-type.data.ts
+++ b/src/mocks/data/data-type/data-type.data.ts
@@ -985,7 +985,7 @@ export const data: Array<UmbMockDataTypeModel> = [
 		id: 'dt-integer',
 		parent: null,
 		editorAlias: 'Umbraco.Integer',
-		editorUiAlias: 'Umb.PropertyEditorUi.Integer',
+		editorUiAlias: 'Umb.PropertyEditorUi.Number',
 		hasChildren: false,
 		isFolder: false,
 		isDeletable: true,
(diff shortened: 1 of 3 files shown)


leekelleher (approved):
Tested out, all good. 👍

I'd picked up on a rogue reference to "Umb.PropertyEditorUi.Integer", which I've swapped over to use "Umb.PropertyEditorUi.Number" instead.

Should this be marked as a breaking-change, since "Umb.PropertyEditorUi.Integer" has been removed since the initial RC?

PR #1759, opened 2024-05-02
`property-value-change` should not be hardcoded

iOvergaard:
this updates (essentially search/replaces) all instances of hardcoded `property-value-change` with the `UmbPropertyValueChangeEvent` class


madsrasmussen (approved):
Looks good 👍

PR #1760, opened 2024-05-02
fix: ensure any property editor UIs save their initial value on the context

iOvergaard:
## Description
This ensures that if, say, a textarea is saved with "rows=10" without touching the input box at all

**Before**
image

**After**
image

## How to test
1. Create a new data type
2. Select textarea
3. Hit save
4. Verify that the default configuration is saved, i.e. rows=10

### DIFF
--- a/src/packages/core/property/property/property.element.ts
+++ b/src/packages/core/property/property/property.element.ts
@@ -234,7 +234,10 @@ export class UmbPropertyElement extends UmbLitElement {
 				this.#valueObserver = this.observe(
 					this.#propertyContext.value,
 					(value) => {
+						// Set the value on the element:
 						this._element!.value = value;
+						// Set the value on the context as well, to ensure that any default values are stored right away:
+						this.#propertyContext.setValue(value);
 						if (this.#validationMessageBinder) {
 							this.#validationMessageBinder.value = value;
 						}


iOvergaard:
I noticed that if the value is undefined, then the alias will be saved sort of without a value. We need to make sure that scenario works on the server.

src/packages/core/property/property/property.element.ts
```diff
@@ -234,7 +234,10 @@ export class UmbPropertyElement extends UmbLitElement {
 				this.#valueObserver = this.observe(
 					this.#propertyContext.value,
 					(value) => {
+						// Set the value on the element:
 						this._element!.value = value;
+						// Set the value on the context as well, to ensure that any default values are stored right away:
+						this.#propertyContext.setValue(value);
```
nielslyngsoe, line 240: When reading this code I get suspicious that something else is wrong.

Like what I read happening here is that the observation callback is called when the propertyContext value changes, and in that callback the value gets set again? Like why would it need to set the value again, if it was just set.

I'm not sure how that could make any difference, but if so I would say something else should be implemented differently.

In terms of the preset data, like in your example, it does to me sound as if the data of the preset isn't begin set to the workspace data? and this code might trigger something so that happens, but this is not good from a data observation point.

iOvergaard: That might very well be. I think this might just have been an oversight. As you see on the line immediately above, the value is sent to the element `this._element!.value` so the element knows about initial values. However, the element will never inform the property context of those values before the user touches them. That is why I thought if we just gave the context the initial value as well, then all values would count as "touched" and thereby be saved.

It's another story entirely if we don't want to persist initial values on purpose, then we need to let the elements know that they should fallback to hardcoded initial values if the value is undefined.

What do you think, do you want to look more into it?

leekelleher (approved):
From initially looking at the code, I'd have thought that `this.#propertyContext.setValue` would have updated `this.#propertyContext.value` and cause an infinite loop, but on inspection, it makes a call to `this.#datasetContext?.setPropertyValue(alias, value)`. So it's probably that `.setValue` is a misnomer?

I've this tested it out on Data Types (with default data) and Documents/Media, and it works as expected, no adverse effects.

src/packages/core/property/property/property.element.ts
nielslyngsoe: We are sort of missing a default value system. But I think that is a feature out of our scope.

So to get around this I made a PR to change the way the default are set in the Data-Type Workspace.
Just cause I'm affright this code could lead to other problems in the future.

PR #1780, opened 2024-05-08
Bugfix: examine dashboard status and modals

loivsen:
## Description

- Fixed a bug where the search bar wouldn't show if the index health status was unhealthy.
- Now shows the health message if there is any. Fallback message is the health status.
- Now shows a yellow clock icon while the index is rebuilding, rather than a green checkmark.
- When searching for an index item, the corresponding workspace now opens when clicking on the name.
- Fixed a bug where the modals didn't load properly
- Fixed a bug where the UI would be forever stuck when starting to rebuild.
- Adds localization to Examine Management

This fixes some of the issues mentioned here https://github.com/umbraco/Umbraco-CMS/issues/16213

## Types of changes


leekelleher (approved):
Tested out, all working as expected. 🚀

PR #1784, opened 2024-05-08
fix: token after refresh is not persisted

iOvergaard:
## Description

When your access_token expires, we exchange the refresh_token for a new access_token, however this new token is never persisted in local storage meaning, that next time you refresh you are "logged out" prematurely

## How to test

1. Let your access token expire (usually 5 minutes but configurable with "TimeOut" in appsettings) and check that the client exchanges its refresh token to new access token
2. Now refresh the browser (F5) and verify that you are still logged in


leekelleher (approved):
Tested out, works as expected. 🚀

PR #1801, opened 2024-05-13
Feature: Login without popup

iOvergaard:
## Description

Currently, when you have more than one provider, we are logging you in through a popup. This is a major difference from the normal flow where you are just redirected to the login screen. This PR aims to align the two, so regardless whether you have 1 or 3 login providers, you always get the same experience.

Log in (no previous session): You click a login provider, you get sent to the login screen of that provider

Login (no previous session, no providers): You get automatically redirected to the login screen (this experience has not been altered with this PR)

Timed out (all cases): You see the login modal on top of your backoffice, you click a login provider and log in through a popup

This PR also fixes a small-ish issue where the redirect back to the frontpage could happen even if there was no token saved yet (race condition).


leekelleher (approved):
Tested out (with the different scenarios), works as expected. 👍

PR #1809, opened 2024-05-14
Feature: Pass the entity type along the unique when requesting tree data.

madsrasmussen:
## Description

Pass the entity type along the unique when requesting tree data.

## Types of changes


leekelleher (approved):
Tested out, all is still working as expected! 👍 Debugged various tree sources to check the `args.parent` value, entity type was present, all great! 🎉

PR #1811, opened 2024-05-14
Mark themes as experimental

JesmoDev:
Clearly marks themes as an experimental feature.

image


leekelleher:
Noticed this PR just merged in, and wondered if the `uui-box`'s "header-actions" slot could be used with a `uui-tag`...
What do you think?

Screenshot 2024-05-15 140629

PR #1815, opened 2024-05-15
Feature: Link and Unlink External Login Providers

iOvergaard:
## Description

Allows backoffice users to link and unlink external providers that support it.

## How to test

1. Create a backoffice user and log in on that
2. Go to the user sidebar - external logins - link the user to a provider
3. Log out and log back in now using the provider, check that it works
4. Go back to the sidebar and verify the provider is marked as connected
5. Try and unlink, verify you are logged out
6. Log back in with the Umbraco credentials and verify the provider is now unlinked

## Screenshots

**Two providers, one is enabled and one doesn't exist on the server**
image


leekelleher (approved):

PR #1831, opened 2024-05-16
Bugfix: Await Property Value Changes + Block Grid Editor fixes

nielslyngsoe:
Important update of the way we set value of properties.

With this update a property value change will be stored without observations triggered, until all property value changes has been retrieved. This fixes a complex problem of the Block Grid Editor where it was triggering multiple auto corrections(updates of its value) which resulted in value updates begin stacked and coming back again later cause of the delay when setting a property value in variant contexts or other contexts which potentially had a minor delay when setting property values. — In other words, old values came back with a slight delay causing endless loops.

This problem was revealed via Block Grid Editor, but would be relevant for any Property Editor that would carry out more than one auto corrected update of its value.

As well this contextually allows for value to be set to be of type Promise, meaning a long-running operation for setting a property value can be used to lock the data of a workspace for the period. This could in another PR be utilised to prevent saving a workspace in locked mode. To ensure user first saves when its unlocked(Most likely it should just await the state before carrying out the save command)

Notice this brings a feature to our DeepStates called Mute/Unmute, this is what enables a workspace to carry out data changes without notifying the Observers. (A few tests has been implemented for this feature)

This code as well implements/correct layout corrections for the Block Grid Editor. columnSpan and rowSpan will now be adjusted to configuration and limited to the context of its existence — If a block is located within an Area that spans 6 columns the Block should as well be limited to 6 columns and auto adjusted for such.

## Types of changes

## Checklist


leekelleher (approved):
Tested out with various scenarios, all working as expected! 👍

PR #1837, opened 2024-05-16
Feature: Thumbnails and sorter dropzone collision

loivsen:
## Description

Better thumbnail quality for big screens
Fixed a bug where sorting an item triggered the dropzone component.


leekelleher (approved):
Super, works well!

PR #1840, opened 2024-05-17
Feature: MediaPickerModal Filters & InputMedia Configuration

loivsen:
## Description

Adds filter options to the Media Picker Modal

Added some of the missing configuration of Property Editor Media Picker

Media Collection now has the checkered background for transparent images 

🎉🎉

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Chore (minor updates related to the tooling or maintenance of the repository, does not impact compiled assets)

PR #1841, opened 2024-05-20
Feature: Adds `umb-input-rich-media` component

leekelleher:
## Description

@loivsen, I was looking over PR #1840 (branch `feature/media-picker-filter-and-media-input-config`), and noticed properties added to the `umb-input-media` component (for focal point and crops), which may cause us a potential issue with the different use-cases between the **Media entity picker** (that stores the Guid values) and the **richer Media Picker** (that stores the focal/crop data).

This PR adds in a `umb-input-rich-media` component (that inherits from `umb-input-media`, so not to duplicate too much code), and adds the newer features.

Let me know what you think about these changes.

## Types of changes


loivsen:
Seems like a good change, I'll go ahead and merge it into mine👌

PR #1843, opened 2024-05-20
Feature: Adds Mustache external library

leekelleher:
## Description

As part of discussion #1729 about Label Templates, it had been suggested to make use of the Mustache templating library.

- #1729 

Adding the dependency as a separate PR, so it can be used as a point of reference for further discussions.

## Types of changes


madsrasmussen:
@leekelleher Do you have an example of how to "register" a tag for the Mustache template? Ex. How did you solve to get the property headline in your example?

Could we live without the mustache syntax in the first version just to leave things as simple as possible for now?

leekelleher:
@madsrasmussen In my example _(from an internal demo)_, I didn't need to use any custom helpers to render the property's label name and alias with Mustache.

Using the template ``"This property has an alias of `{{alias}}`, and a label of **{{label}}**."``, and the instance of the `umb-property-layout` component as the modal, they can be rendered with `const rendered = mustache.render(template, model);`

If any helpers are needed, then they would be added to the `model` as [Mustache functions](https://github.com/janl/mustache.js?tab=readme-ov-file#functions). We'll need to explore this more.

leekelleher:
Closing this PR for now. We don't need the dependency on the Mustache library just yet.

PR #1844, opened 2024-05-20
Feature: Render Property description with Markdown

leekelleher:
## Description

Adds a **Formatting API** controller and function to process/transform property descriptions through...

1. Umbraco backoffice localization
2. Markdown, using [Marked.js](https://github.com/markedjs/marked) library
3. HTML sanitized with [DOMPurify](https://github.com/cure53/DOMPurify) library

Once we've decided on using the Mustache templating library, (PR #1843), we could incorporate it in the processing.

re: [Umbraco Dictionary Item (translations)](https://docs.umbraco.com/umbraco-cms/fundamentals/data/dictionary-items#using-dictionary-values-in-document-types) as labels/descriptions; I haven't implemented this yet, as I'm unsure of when we should load up all the dictionary items from the server.

**Configurations**

For reference the configurations used for the static references of both Marked and DOMPurify can be found on the documentation for each library.

- https://marked.js.org/using_advanced
- https://github.com/cure53/DOMPurify/tree/main/demos#what-is-this
- https://github.com/cure53/DOMPurify?tab=readme-ov-file#can-i-configure-dompurify

## Types of changes


iOvergaard (commented):
Looks good, but I don't think you need to create a global context as you don't seem to carry any state around with it. You could probably create a file that exports a few static functions. Try and search for any file ending on `.function.ts` to see how it's done elsewhere ☺️

leekelleher:
@iOvergaard I've refactored the code to be a controller class, with a new `formatting-api` library.

iOvergaard:
Looks great. The only thing blocking this right now is that DOMPurify does not play well with our test system. Will have to investigate.

iOvergaard (approved):
I fixed web-test-runner so it doesn't use the 'browser' field in npm packages. That way it seems to work by using the esm version of any packages including DOMPurify. Also made sure we are not importing it twice. Looks good to me.

PR #1851, opened 2024-05-21
Bugfix: request reload of entity after publish

madsrasmussen:
## Description

Dispatch request reload event after publish and publish bulk actions.

## How to test:
* Test that a collection publishing state is updated when using the bulk actions.
* Test that a collection publishing state is updated when using the tree item actions

## Types of changes


leekelleher (approved):
Tested out, all good! 👍

PR #1857, opened 2024-05-22
Feature: Login should support returnPath

iOvergaard:
## Description

When you log in with Basic Auth, you are expected to be returned to something else that the backoffice -- to a relative path on the website. This feature enables someone to send the user to the backoffice and force them to log in (`?status=false`) and to be redirected to something else (`?returnPath=xyz`). We need to make sure that an actual log in happens since a cookie needs to be created on the backend, so it doesn't help to have a token lying around, which is why the `?status=false` clears all tokens.

## How to test

1. Log in normally on the backoffice
2. Navigate to `/umbraco?status=false&returnPath=/some-public-url`
3. Verify that you see the login screen again, and that you can log in normally
4. Verify that after log in, you are sent to /some-public-url (change the URL in the parameter if you have an actual page)

The purpose is to be able to log in to the public website using the Backoffice (so that external login providers work), and to be sent right back to the public website after the fact, but to test it, you don't need to set up the whole backend environment (there's a pending fix on the backend as well) -- this feature can and should work independently.

Read more on the Docs here: https://docs.umbraco.com/umbraco-cms/reference/configuration/basicauthsettings#redirecttologinpage


leekelleher (approved):
Tested out, working as expected! 🙌

PR #1867, opened 2024-05-23
Bugfix: Collection: determine if `filter.skip` has preconfigured value

leekelleher:
## Description

In the default Collection context, determine if `filter.skip` has a preconfigured value.

/cc @nathanwoulfe 

## Types of changes

### DIFF
--- a/src/packages/core/collection/default/collection-default.context.ts
+++ b/src/packages/core/collection/default/collection-default.context.ts
@@ -120,11 +120,14 @@ export class UmbDefaultCollectionContext<
 			this.pagination.setPageSize(this.#config.pageSize);
 		}
 
+		const filterValue = this.#filter.getValue() as FilterModelType;
+
 		this.#filter.setValue({
 			...this.#defaultFilter,
 			...this.#config,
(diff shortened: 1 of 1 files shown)


src/packages/core/collection/default/collection-default.context.ts
```diff
@@ -120,11 +120,14 @@ export class UmbDefaultCollectionContext<
 			this.pagination.setPageSize(this.#config.pageSize);
 		}
 
+		const filterValue = this.#filter.getValue() as FilterModelType;
+
 		this.#filter.setValue({
 			...this.#defaultFilter,
 			...this.#config,
 			...this.#filter.getValue(),
```
madsrasmussen, line 128: It looks like we append this value twice. 🤔

leekelleher: Doh! 🤦  Thanks @madsrasmussen, I was checking if anybody noticed. 👀  I'll fix it up.

iOvergaard (approved):

PR #1872, opened 2024-05-23
Bugfix: Document Picker: Ensures `max = 1`

leekelleher:
## Description

To support the legacy data format of the [Content Picker](https://docs.umbraco.com/umbraco-cms/fundamentals/backoffice/property-editors/built-in-umbraco-property-editors/content-picker), and the backend will only accept a single `Guid` (`string`) value, not a comma-separated string, so we need to enforce that `max = 1`.

## Types of changes


madsrasmussen (approved):
It works. Thanks 🎉

PR #1873, opened 2024-05-23
Bugfix: Picker inputs - moved `addValidator()` calls to `constructor()`

leekelleher:
## Description

In the input components for some of the picker editors, the `addValidator()` was being made inside the `connectedCallback()`, this could lead to potential issues as the component may be added to the DOM multiple times. This PR moves the `addValidator()` calls to the `constructor()`.

## Types of changes


madsrasmussen (approved):
Looks good

PR #1879, opened 2024-05-23
Bugfix: Property descriptions should use uui-scroll-container

iOvergaard:
## Description

Instead of `overflow-x` we should use `uui-scroll-container` which preserves some styling choices.

I have also made sure that the title field works with very long word.

I also considered using `word-break: break-word` on the description, but as we don't know what's in there and it now has a scroll container, I don't think it's a good idea to limit the width of regular text to the outer width of the box.

**Before**

image

**After**

image

### DIFF
--- a/src/packages/core/property/property-layout/property-layout.element.ts
+++ b/src/packages/core/property/property-layout/property-layout.element.ts
@@ -126,7 +126,6 @@ export class UmbPropertyLayoutElement extends UmbLitElement {
 
 			#description {
 				color: var(--uui-color-text-alt);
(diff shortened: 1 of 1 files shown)


bjarnef:
It seems this has a side-effect that it now focus every property description, e.g. in configuration of datatypes.

image

But also in content and media. So if using descriptions or properties, it requires a lot on tabs to get to the input editors.

image

iOvergaard:
Good point, @bjarnef. I will open a new issue based on what you said.

PR #1880, opened 2024-05-23
Reset data-type configs when switching property editor

nielslyngsoe:
Revert some potentially bad code and correcting the way we set default values in data-type workspace

The code to be reverted is of this PR:https://github.com/umbraco/Umbraco.CMS.Backoffice/pull/1760
That code seems to work fine, but could lead to unexpected results and would be hard to debug in the future.
Also the scenario that was built on top of was wrong, which leads to the rest of the changes of this PR.

In data-type configuration editor, Instead of setting a default value when a value is undefined, we now will only set the default values when a data-type is new or when the property editor is changed.

problem with old approach was that if new default came along after the creation of a data-type then you might get that set without your awareness.

before this PR your values was also not reset when changing a property editor, which could have lead to bad configurations.

## Types of changes

## Checklist

### DIFF
--- a/src/packages/core/property/property/property.element.ts
+++ b/src/packages/core/property/property/property.element.ts
@@ -237,8 +237,6 @@ export class UmbPropertyElement extends UmbLitElement {
 					(value) => {
 						// Set the value on the element:
 						this._element!.value = value;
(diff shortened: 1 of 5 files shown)


leekelleher (approved):
Tested out, works as expected. 👍

PR #1882, opened 2024-05-23
Bugfix: fixing cyclic dependencies continued

madsrasmussen:
## Description

This PR is part of the pursue of getting rid of our cyclic dependencies. These are some "easy" updates that mostly changed some module internal imports using the module alias instead of using relative imports.

As far as I can see the PR should be "safe", and not include any breaking changes.

## Types of changes


leekelleher (approved):
Tested out, (by testing the areas affected by the changes), all still working as expected. 👍

PR #1883, opened 2024-05-23
move imaging module to media package

madsrasmussen:
## Description

This PR moves the imaging module from "core" to the media package. The imaging module has an import to the media module and we cannot import from packages in "core". 

The move should be safe as the import alias is still the same and has the same exports.

## Types of changes

## Checklist


leekelleher (approved):
Makes sense! 👍 
Tested out, all good.

PR #1884, opened 2024-05-23
Bugfix: Move upload field and component

madsrasmussen:
## Description

The circular dependencies quest continues. The App element is dependent on the core, the `input-upload-field` inside core is dependent on the app context. We need to fix that.

My best suggestion for a new home for the Upload Field Property Editor UI and `input-upload-field` is the media module. Ideally, it would have been in the root of the media package, but as we don't have a module for the root of our packages this will have to do.

> [!CAUTION]
> Breaking Change
> Import of the `UmbInputUploadFieldElement` has moved from `@umbraco-cms/backoffice/component` to `@umbraco-cms/backoffice/media`

## How to test:

Make sure that the Upload Field property editor UI still renders the correct UI.

## Types of changes


leekelleher (approved):
Tested it out, upload field still working as expected! 👍

PR #1887, opened 2024-05-23
Bugfix: Restore custom validation on property editors

iOvergaard:
## Description
If there is a custom regex for validation on a property editor, and we cannot match it with our existing regex'es, it is always a custom one, and we should just show it to the user.

Also added localization to the area.

Also added a default "custom validation" regex string, so the user doesn't save an empty regex. This also lets the uui-select underneath actually show it as selected (for some reason it cannot immediately select an empty value, which I think is a bug in UUI).

Fixes https://github.com/umbraco/Umbraco-CMS/issues/16356

## How to test
1. Create a property editor on a document and choose one of the predefined validations, e.g. validate as number
2. Test that the regex is saved, and after reload, that it is still visible
3. Now change it to "...or enter a custom validation" and set a custom regex and save
4. Check that the custom regex is shown in the UI after a reload and that "...or enter a custom validation" is still selected

### DIFF
--- a/src/assets/lang/en.ts
+++ b/src/assets/lang/en.ts
@@ -1682,6 +1682,7 @@ export default {
 		elementDoesNotSupport: 'This is not applicable for an Element Type',
 		propertyHasChanges: 'You have made changes to this property. Are you sure you want to discard them?',
 		displaySettingsHeadline: 'Appearance',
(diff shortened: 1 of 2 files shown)


leekelleher (approved):

PR #1889, opened 2024-05-24
fix: the backoffice does not work with window.open

iOvergaard:
## Description

Only assume we are in iframe mode if the origin of the opener matches the current origin. This allows the user to open the Backoffice with `window.open` without staying in this "iframe mode" where the user just sees a message to close the window again.

Fixes https://github.com/umbraco/Umbraco-CMS/issues/16373

## How to test

1. Make sure you dont have any tokens in local storage
2. Now open a fresh Backoffice with `window.open("https://localhost:44339/umbraco")`
3. Ensure that you are redirected to the content section after login and dont get stuck on the "You may now close this window" page
4. (Optional) Let the session time out, so you see the timeout screen and make sure that the _actual_ iframe mode still works

### DIFF
--- a/src/apps/app/app.element.ts
+++ b/src/apps/app/app.element.ts
@@ -79,8 +79,9 @@ export class UmbAppElement extends UmbLitElement {
 				// If we are in the main window, the signal will be caught right here and the user will be redirected to the root.
 				if (
 					window.opener &&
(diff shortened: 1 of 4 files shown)


leekelleher (approved):
Tested out, works as expected. 👍

PR #1890, opened 2024-05-24
Block Grid Editor: Entity Range Validation

nielslyngsoe:
This feature implements entity range validation, for root Blocks and Blocks or Areas. It does not take the types into account jet.

But it proves that a client validation that the server does not know anything about is preventing a publish.
It does prove one problem that still is to be solved, which is that we might get identical client validation messages and server validation messages, so something needs to be done to solve that.
As well HTML of the messages are not rendered correctly at this stage.

Pre-fixes NumberRangeValueType with Umb — not sure it would be used by anyone, if so, then its breaking.

## Types of changes

## Checklist


leekelleher (approved):

leekelleher:
I've tagged as a breaking-change, to cover ourselves for the `NumberRangeValueType` to `UmbNumberRangeValueType` rename.

PR #1893, opened 2024-05-24
Bugfix: Property label does not appear on top

iOvergaard:
## Description

Setting the property label to appear "on top" does not have an effect. This PR maps up the appearance object through the property element to set the eventual value on the umb-property-layout element, which already had a property for "orientation". I also had to reflect that property on the host element so that the styling worked.

Fixes https://github.com/umbraco/Umbraco-CMS/issues/16378


JesmoDev (approved):
LGTM

PR #1897, opened 2024-05-24
Feature: Extendable section routes

madsrasmussen:
## Description

This PR introduces a new extension point to the default section element. Section routes make it possible to add custom routes to all sections. We use it ourselves to register the workspace route. This is the solution To fix up a circular dependency between the section and workspace route constants.

***Manifest example***
``` js
const manifest = {
 type: 'sectionRoute',
 alias: 'Umb.SectionRoute.Workspace',
 name: 'Workspace Section Route',
 element: () => import('../workspace.element.js'),
 api: () => import('./workspace-section-route.route-entry.js'),
 meta: {
  path: '/some-path' // optional
 }
};
```

***API example***
``` ts
export class UmbWorkspaceSectionRouteEntry implements UmbApi, UmbRouteEntry {
 getPath(): string {
  return UMB_WORKSPACE_PATH_PATTERN.toString();
 }

 setup(element: PageComponent, info: IRoutingInfo) {
  (element as UmbWorkspaceElement).entityType = info.match.params.entityType;
 }

 destroy(): void {}
}

export { UmbWorkspaceSectionRouteEntry as api };
```

## Types of changes


leekelleher (approved):
Looking good, exciting feature! 🎉

PR #1902, opened 2024-05-24
Chore: Lee's code sweep (for RC5)

leekelleher:
## Description

Hey-ho, let's go! It's another round of **Lee's Codesweep™**! 🦄 

For RC5, I've focused on the property-editor UIs, many bug fixes and UI polish in here. 🚀
No breaking-changes.

## Types of changes


madsrasmussen:
Multi URL picker all nodes seems to be grayed out 🤔

leekelleher:
@madsrasmussen Thanks for reviewing, I'll take another look at the Multi-URL Picker tweaks.

src/packages/documents/documents/components/input-document/input-document.element.ts
```diff
@@ -139,6 +139,10 @@ export class UmbInputDocumentElement extends UUIFormControlMixin(UmbLitElement,
 		return undefined;
 	}
 
+	#isDraft(item: UmbDocumentItemModel) {
+		return item.variants[0]?.state === 'Draft' ?? false;
```
madsrasmussen, line 143: I think the `?? false` is redundant here

src/packages/documents/documents/components/input-document/input-document.element.ts
```diff
@@ -189,7 +193,7 @@ export class UmbInputDocumentElement extends UUIFormControlMixin(UmbLitElement,
 	#renderItem(item: UmbDocumentItemModel) {
 		if (!item.unique) return;
 		return html`
-			<uui-ref-node name=${item.name} id=${item.unique}>
+			<uui-ref-node name=${item.name} id=${item.unique} ?disabled=${this.#isDraft(item)}>
```
madsrasmussen, line 196: We should go with a class here, the item is not disabled. If you set the opacity to 0.6 then it's the same as the tree item.

leekelleher:
@madsrasmussen I've reverted the `disabled`/draft change for the Multi-URL Picker, as there's more work involved (than me doing a code sweep), so I'll look to fix it up in a separate PR. I've made the other amends from your feedback. Ready to re-review.

madsrasmussen (approved):

PR #1904, opened 2024-05-25
Bugfix: Media Workspace Info details

leekelleher:
## Description

Implements the the Media info details to match v13.

Fixes
- https://github.com/umbraco/Umbraco-CMS/issues/16396

## Types of changes


madsrasmussen (approved):
I have swapped the detail repo for the item repo. The item has all the things we need in this case.

I also did some smaller tweaks here and there.

PR #1910, opened 2024-05-26
Blocks for RTE

nielslyngsoe:
Feature to insert and edit Blocks in TinyMCE.

## Types of changes

## Checklist


iOvergaard:
I also could not get the "edit block by clicking on it" to work

src/packages/block/block-rte/context/block-rte-manager.context.ts
```diff
@@ -0,0 +1,54 @@
...
+
+	insert(
+		layoutEntry: BlockLayoutType,
+		content: UmbBlockDataType,
+		settings: UmbBlockDataType | undefined,
+		modalData: UmbBlockRteWorkspaceData,
+	) {
+		if (!this.#editor) return false;
+
+		if (layoutEntry.displayInline) {
```
iOvergaard, line 38: This never seems to be "true" even though I set it up for my block, so the block is always inserted as a "display:block" element:

image

Also, I noticed something for custom_elements in the TinyMCE documentation, where they support specifying [custom_elements](https://www.tiny.cloud/docs/tinymce/6/content-filtering/#custom_elements) (and have been for a while), so we might consider using that instead of our custom pre-init filter to unwrap inline elements:

```js
custom_elements: 'umb-rte-block,~umb-rte-block-inline',
```

iOvergaard: Agree on the content editable. You should be able to do that using the pre-init filter.

iOvergaard: But for inline, I didn't see it as missing CSS. Because the property was not true, it didn't even seem to insert the inline element.

nielslyngsoe: But I fixed the UI/Action-bar and the Icons, which needed another interesting Event-Proxy. So now its far more complete. :-)

src/packages/block/block-rte/components/block-rte-entry/block-rte-entry.element.ts
```diff
@@ -0,0 +1,146 @@
...
+
+	#renderBlock() {
+		return html`
+			<umb-extension-slot
+				type="blockEditorCustomView"
+				default-element=${'umb-ref-rte-block'}
+				.props=${this._blockViewProps}
+				>${this.#renderRefBlock()}</umb-extension-slot
+			>
+			<uui-action-bar>
```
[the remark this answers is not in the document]
nielslyngsoe: Solved

iOvergaard (approved):
LGTM

PR #1919, opened 2024-05-27
Bugfix: navigating between workspaces throws errors

madsrasmussen:
## Description

Use the router component resolver to handle the route component instead of the ExtensionElemenInitializer cached component.

This is a tiny bit hacky, as the "correct" solution would be to make a custom route resolver that handles it. But as we are in control of both libs and they work almost the same I have just passed the manifest element prop directly to the router. Then the router will correctly construct the element and update the DOM.

## Types of changes


leekelleher (approved):
Tested out, I can navigate between workspace types and sections without error! 🙌

PR #1930, opened 2024-05-27
Feature: Document Type Descriptions

iOvergaard:
## Description

This adds the description field to the document type editor. It also adds the description to the Document Create Options Modal.

This partly fixes https://github.com/umbraco/Umbraco-CMS/issues/16391

NB! We are now able to save descriptions, and not lose migrated data, however we still lack descriptions on the different item pickers because the item models do not carry the description with them.

## How to test

Add a description to a doc type:
image

Make sure it shows up in the options creator modal:
image


leekelleher (approved):

Initially, I thought the transparent border looked odd, but then I took a look at v13 and it's the same there. All good for approval! 👍

PR #1933, opened 2024-05-27
Feature: Member Type Descriptions

iOvergaard:
## Description

Adds descriptions to member types.

NB! There are currently no usages for this description as we do not have the create options modal for members, because the member section is currently implemented as a section without folders or a menu tree, where that options modal would normally be used.

## How to test

1. Add a description to a member type
2. Verify the description is still set upon reload


leekelleher (approved):

PR #1941, opened 2024-05-28
Fix: editing block in Area

nielslyngsoe:
Fixes a problem where a Block Modal does not have enough data about the area of which the editing is taking place, resulting in the Block begin appended to the root, not in the area.

To fix this I move the workspace modal registration into each implementation to ensure type awareness — This can maybe be optimised later, but for now it gives clarity.
Simply because the problem was that the modal did not knew about the areaKay or parentUnique, which resulted in the update taking place in the root.

## Types of changes

## How to test?

Create similar scenario like this. Explore when adding another Block and editing that block, a copy will be placed in the root.
image

## Checklist


iOvergaard (approved):
Tests out ok!

PR #1943, opened 2024-05-29
Chore: EsLint Rule No relative import to import map module

madsrasmussen:
## Description

This PR adds an es lint rule to help us import modules with an import alias instead of a relative import. The relative import becomes problematic when we start bundling each module because then the module will include the relatively imported code instead of leaving the import as an external dependency.

## Types of changes


leekelleher (approved):
Tested out with the affected areas/property-editors/inputs, all still working as they were. 👍

PR #1953, opened 2024-05-29
The document `Custom` status is not translated correctly

iOvergaard:
## Description

Adds missing localization variable for the "Custom" status and improves the Node.js devops script to find unused keys.


JesmoDev (approved):
LGTM

PR #1956, opened 2024-05-29
Bugfix: Document Type Workspace, localizes the name and alias input placeholders

leekelleher:
## Description

In the Document Type Workspace, the "Enter a name..." and "Enter an alias..." placeholder labels for the inputs were hardcoded to English. This PR updates with the localization keys.

## Types of changes

### DIFF
--- a/src/packages/core/components/input-with-alias/input-with-alias.element.ts
+++ b/src/packages/core/components/input-with-alias/input-with-alias.element.ts
@@ -6,7 +6,7 @@ import { type UUIInputElement, UUIInputEvent } from '@umbraco-cms/backoffice/ext
 import { generateAlias } from '@umbraco-cms/backoffice/utils';
 
 @customElement('umb-input-with-alias')
-export class UmbInputWithAliasElement extends UmbFormControlMixin<string>(UmbLitElement) {
+export class UmbInputWithAliasElement extends UmbFormControlMixin<string, typeof UmbLitElement>(UmbLitElement) {
 	@property({ type: String })
 	label: string = '';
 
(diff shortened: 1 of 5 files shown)


src/packages/core/components/input-with-alias/input-with-alias.element.ts
```diff
@@ -68,17 +68,21 @@ export class UmbInputWithAliasElement extends UmbFormControlMixin<string>(UmbLit
...
+				id="name"
+				placeholder=${this.localize.term('placeholders_entername')}
+				label=${this.label}
+				.value=${this.value}
+				@input=${this.#onNameChange}>
 				<!-- TODO: should use UUI-LOCK-INPUT, but that does not fire an event when its locked/unlocked -->
 				<uui-input
 					name="alias"
 					slot="append"
 					label="alias"
```
iOvergaard, line 82: Looks good, I think we should copy the placeholder localization to this label attribute as well :)

src/packages/core/components/input-with-alias/input-with-alias.element.ts
```diff
@@ -68,17 +68,21 @@ export class UmbInputWithAliasElement extends UmbFormControlMixin<string>(UmbLit
 	}
 
 	render() {
-		// Localizations: [NL]
 		return html`
-			<uui-input id="name" placeholder="Enter a name..." label=${this.label} .value=${this.value} @input="${this.#onNameChange}">
+			<uui-input
+				id="name"
+				placeholder=${this.localize.term('placeholders_entername')}
+				label=${this.label}
```
iOvergaard, line 75: And copy the placeholder localization here as well

iOvergaard (changes requested):

src/packages/core/components/input-with-alias/input-with-alias.element.ts
leekelleher: @iOvergaard Sure, but the `label` property (for the "name" input) is set on the parent element, e.g. for the Document Type Workspace: https://github.com/umbraco/Umbraco.CMS.Backoffice/blob/main/src/packages/documents/document-types/workspace/document-type-workspace-editor.element.ts#L88

Did you mean add the localization on the parent element? (_This would be for Document Type, Media Type and Member Type workspaces_).

leekelleher: Updated, to add the localization on the parent component for the `label` text, with a fallback in the `umb-input-with-alias` to use the same localization (e.g. if it's `undefined` or someone wants to use different text/localization). 👍

iOvergaard (approved):
LGTM

PR #1960, opened 2024-05-30
Bugfix: Media workspace info references correct localization

leekelleher:
## Description

When a Media item hasn't been references elsewhere, the text in the workspace info references panel displayed `"This Data Type has no references."`, as it was using the incorrect localization key. This PR corrects the localization key.

Fixes https://github.com/umbraco/Umbraco-CMS/issues/16485

## Types of changes

### DIFF
--- a/src/packages/media/media/workspace/views/info/media-workspace-view-info-reference.element.ts
+++ b/src/packages/media/media/workspace/views/info/media-workspace-view-info-reference.element.ts
@@ -135,7 +135,7 @@ export class UmbMediaWorkspaceViewInfoReferenceElement extends UmbLitElement {
 	}
 
 	#renderItems() {
-		if (!this._items?.length) return html`<p>${this.localize.term('references_DataTypeNoReferences')}</p>`;
+		if (!this._items?.length) return html`<p>${this.localize.term('references_itemHasNoReferences')}</p>`;
 		return html`
 			<uui-table>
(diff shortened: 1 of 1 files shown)


src/packages/media/media/workspace/views/info/media-workspace-view-info-reference.element.ts
```diff
@@ -135,7 +135,7 @@ export class UmbMediaWorkspaceViewInfoReferenceElement extends UmbLitElement {
 	}
 
 	#renderItems() {
-		if (!this._items?.length) return html`<p>${this.localize.term('references_DataTypeNoReferences')}</p>`;
+		if (!this._items?.length) return html`<p>${this.localize.term('references_itemHasNoReferences')}</p>`;
```
iOvergaard, line 138: This is perhaps out of scope, but I would really prefer if we used umb-localize where possible because you then move the listener down to that element so it doesn't have to update all the time.

```suggestion
		if (!this._items?.length) return html`<p><umb-localize key="references_itemHasNoReferences">This item has no references.</umb-localize></p>`;
```

This also allows us to more easily be able to see what the translation is supposed to say 👍

leekelleher: Good point! 👍 I'll keep in mind to fix in other places too.

PR #1961, opened 2024-05-30
Bugfix: Property description, sets `max-width` for child items

leekelleher:
## Description

In a property description, when an `` tag is rendered (from Markdown syntax), if the width was wider than the overflow, it would be cut off.  This PR adds a style rule to set the `max-width: 100%` for any child elements.

Screenshot 2024-05-30 105537

This is a partial fix for https://github.com/umbraco/Umbraco-CMS/issues/16499.

## Types of changes

### DIFF
--- a/src/packages/core/property/property-layout/property-layout.element.ts
+++ b/src/packages/core/property/property-layout/property-layout.element.ts
@@ -130,6 +130,10 @@ export class UmbPropertyLayoutElement extends UmbLitElement {
 				color: var(--uui-color-text-alt);
 			}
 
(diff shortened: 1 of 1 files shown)


iOvergaard:
I added a new story to test out Markdown (and thereby discovered that pre+code elements caused an overflow, so I fixed that). You can see the story here: https://ambitious-stone-0033b3603-1961.westeurope.1.azurestaticapps.net/?path=/story/umb-property-layout--with-markdown

iOvergaard (approved):

PR #1969, opened 2024-05-31
Server models 2024-05-31

iOvergaard:
## Description

Generate new models and add `template` to the blueprint model. I am a little unsure of this since the `template` is not returned from the server in the details request anyway.

### DIFF
--- a/src/packages/documents/document-blueprints/repository/detail/document-blueprint-detail.server.data-source.ts
+++ b/src/packages/documents/document-blueprints/repository/detail/document-blueprint-detail.server.data-source.ts
@@ -41,6 +41,7 @@ export class UmbDocumentBlueprintServerDataSource implements UmbDetailDataSource
 				unique: '',
 				collection: null,
 			},
+			template: null,
 			values: [],
 			variants: [],
 			...preset,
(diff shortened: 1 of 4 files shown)


src/packages/documents/document-blueprints/repository/detail/document-blueprint-detail.server.data-source.ts
```diff
@@ -133,7 +135,7 @@ export class UmbDocumentBlueprintServerDataSource implements UmbDetailDataSource
 			id: model.unique,
```
madsrasmussen, line 135: Is this the correct create model? Should it be 
`CreateDocumentBlueprintRequestModel` ?

iOvergaard: Ohh, didn't notice that before. You may be right, let me see.

iOvergaard: Found the same problem with the update model (`UpdateDocumentBlueprintRequestModel`), albeit that didn't require a template. But now we use the correct models at least.

madsrasmussen (approved):
I wonder way the typescript didn't catch that we were passing the wrong model 🤔

iOvergaard:
> I wonder why typescript didn't catch that we were passing the wrong model 🤔

The two models were identical up until yesterday 💯

PR #1977, opened 2024-06-03
Chore: Bundling dictionary module

madsrasmussen:
## Description

This is the beginning of bundling our modules. Unfortunately, bundling all the modules before v15 might not be possible because some files are located incorrectly. These files can't be moved without a breaking change. Therefore, We will go through the modules one by one, and bundle the possible ones.

This PR sets up the basic build script for bundling. The build is made with npm workspaces which makes it possible to run a "local" build script in each package. The bundling is done by Vite.

We now have one for the npm package, which is not bundled, and one for the CMS which will be partly bundled. 

The idea is to build the same way as before but swap out the "old" module files for the bundled ones when a module is ready. 

## How to test

1. Run the `npm run build:for:cms` script in the CMS project. 
2. Check that you see the bundled javascript output in the `src/packages/dictionary` folder.
3. Clear your browser cache
4. Test the dictionary section/workspaces etc. All works as before

## Types of changes

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -21166,6 +21166,8 @@
 				"url": "https://github.com/sponsors/wooorm"
 			}
 		},
-		"src/packages/dictionary": {}
+		"src/packages/dictionary": {
+			"name": "@umbraco-backoffice/dictionary"
+		}
 	}
 }
--- a/package-lock.json
+++ b/package-lock.json
@@ -8,6 +8,9 @@
 			"name": "@umbraco-cms/backoffice",
 			"version": "14.1.0",
 			"license": "MIT",
+			"workspaces": [
+				"./src/packages/dictionary"
+			],
 			"dependencies": {
 				"@types/diff": "^5.2.1",
 				"@types/dompurify": "^3.0.5",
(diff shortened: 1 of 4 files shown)


iOvergaard (commented):
Tested on the dev server as well as the CMS. It definitely loads fewer files. Makes sense that is still has a number of files since much of it is lazy-loaded.

When I create a new dictionary item, I get an error in console, but I'm not sure if that is related:

image

Lastly, it seems all repository files are loaded, even before I'm clicking on the associated actions. For example, the `dictionary-export.repository-CACHE123.js` file always seems to load up front. When I then click on the "Export" action, even more files load such as the modal element and the icons in separate requests, which I guess is ok (and icons are probably not bundled anyway). Do we need to load repository class files up front?

iOvergaard:
It also seems there is a change to "package-lock.json" but I don't dare to commit it:

image

madsrasmussen:
1. The JS error is not caused by this PR. I think it is a problem in all workspaces when creating a new item.

2. I have updated the package.json. The update is ok.

3. The export repo is set to be lazy loaded in the manifest, so I guess it is loaded when we create the list of available entity actions 🤔 . As far as I can see there is not much more we can do from the bundle.

iOvergaard (approved):
Ok, thanks for confirming, @madsrasmussen. Having Vite bundle it up is really helpful to see what _might_ be lazy-loaded and what isn't. With that said, this LGTM.

PR #1978, opened 2024-06-03
Feature: TinyMce Custom Configuration

iOvergaard:
## Description

Some configurations such as `statusbar`, `menubar`, and other visuals are invalid to add after the editor has been rendered. To accommodate this, we are adding a `config` object to the manifest file which essentially supports the same options as the former `CustomConfig` in V13.

There is also a fix for async-loaded extensions, which were previously not considered after the editor was rendered. Now we have an observer on the extensions registry and in case a `tinyMcePlugin` is added or removed, we will re-render the editor and merge the new configuration.

## Types of changes

## How to test?

**Dev server**
1. Turn on the `VITE_UMBRACO_EXTENSION_MOCKS=on` environment variable
2. Notice that the RTE on the "All properties" page are shown with a statusbar and the wordcount plugin

You have to remove these lines from input-tiny-mce.defaults.ts to get the dev server to work:
image

**CMS**
1. Add a manifest for `tinyMcePlugin` and attempt to add configuration
2. Check that the API is also able to load and receive an arguments object with `host` and `editor`, which you can use to interact with the editor (this is already so before this fix, but make sure this hasn't been broken)

image

### DIFF
--- a/public-assets/App_Plugins/tinyMcePlugin.js
+++ b/public-assets/App_Plugins/tinyMcePlugin.js
@@ -2,7 +2,7 @@ export default class UmbTinyMceMockPlugin {
 	/**
 	 * @param {TinyMcePluginArguments} args
 	 */
-	constructor(host, args) {
+	constructor(args) {
 		// Add your plugin code here
 		console.log('editor initialized', args)
 	}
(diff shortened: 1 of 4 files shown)


iOvergaard:
For whoever is reviewing this: The `npm run test` seems to fail on Github, but I cannot reproduce it locally. Could someone else run through this, please?

leekelleher (approved):

PR #1998, opened 2024-06-14
Bugfix: Installer does not take the minimum password length into account

CarlSargunar:
The password length is not correctly being set on the UI element, this is a hacky fix which forces the validation to fail. I need to refactor this (hopefully with some help)

Fixes https://github.com/umbraco/Umbraco-CMS/issues/16601


CarlSargunar:
I THINK this is now ready to go, but if someone is able to review it I'd be grateful

iOvergaard (approved):
I updated it slightly so that we optimize for Lit's update checker by extracting the minimum password length into its own state variable. This has the effect that the template is only re-rendered if the actual number changes, whereas otherwise it would re-render when the object is set again.

LGTM.

CarlSargunar:
Thanks @iOvergaard 

I must admit the architecture of the new Back Office is still a little new to me, but I'm learning :-)

PR #2010, opened 2024-06-18
Bugfix: Render calculated user start nodes in User Workspace

madsrasmussen:
## Description

This PR implements the new endpoint to fetch the calculated start nodes for a user (https://github.com/umbraco/Umbraco-CMS/pull/16609)

What to test:
* Set up a user and assign user groups that will result in different start nodes.
* Test that the calculated result is shown to the user. Be aware that you might have to restart the server to see the effect (more details in PR for management API (https://github.com/umbraco/Umbraco-CMS/pull/16609)

## Types of changes


iOvergaard (approved):
Looks and tests good. I tried several scenarios with varying degrees of start nodes. The only thing I could find is that you have to click out of the user and back into them when you change their user groups to get the UI to call the "calculate-start-nodes" endpoint again. Perhaps we could listen on that when user groups change. Approved nonetheless!

PR #2017, opened 2024-06-19
Bugfix: Add correct Block Type label to Block Workspace editor

rickbutterfield:
## Description

Correctly loads the Block Type from the Block Manager, gets the `label` property and saves it within the Block Workspace context for loading in the create/edit modal.

## Types of changes

## Motivation and context

Fixes https://github.com/umbraco/Umbraco-CMS/issues/16463

## How to test?

Try creating or editing any block within a Block List or Block Grid, and make sure the modal shows the correct Block Type name

Before:
Before screenshot

After:
On create
image

On edit
image

## Checklist

### DIFF
--- a/src/packages/block/block/workspace/block-workspace-editor.element.ts
+++ b/src/packages/block/block/workspace/block-workspace-editor.element.ts
@@ -5,34 +5,34 @@ import { UmbLitElement } from '@umbraco-cms/backoffice/lit-element';
 @customElement('umb-block-workspace-editor')
 export class UmbBlockWorkspaceEditorElement extends UmbLitElement {
 
-  @property({ type: String, attribute: false })
-  workspaceAlias?: string;
+	@property({ type: String, attribute: false })
+	workspaceAlias?: string;
 
-  @property({ type: String, attribute: false })
-  label?: string;
+	@property({ type: String, attribute: false })
+	label?: string;
 
-  render() {
-    return this.workspaceAlias && this.label
-      ? html` <umb-workspace-editor alias=${this.workspaceAlias} headline=${this.label}> </umb-workspace-editor> `
-      : '';
-  }
+	render() {
+		return this.workspaceAlias && this.label
+			? html` <umb-workspace-editor alias=${this.workspaceAlias} headline=${this.label}> </umb-workspace-editor> `
+			: '';
+	}
 
-  static styles = [
-    UmbTextStyles,
-    css`
+	static styles = [
+		UmbTextStyles,
+		css`
 			:host {
 				display: block;
 				width: 100%;
 				height: 100%;
 			}
 		`,
-  ];
+	];
 }
 
 export default UmbBlockWorkspaceEditorElement;
 
 declare global {
-  interface HTMLElementTagNameMap {
-    'umb-block-workspace-editor': UmbBlockWorkspaceEditorElement;
-  }
+	interface HTMLElementTagNameMap {
+		'umb-block-workspace-editor': UmbBlockWorkspaceEditorElement;
+	}
 }


src/packages/block/block/context/block-manager.context.ts
nielslyngsoe, line 1: Hi Rick, it seems there is a ´code-style´ issue here, but only in this file. we Use tabs not spaces and here you got all tabs changed into spaces. If you use VSCode I can recommend getting the Extention EditorConfig, so you get the right configuration for the project.

src/packages/block/block/workspace/block-workspace.context.ts
```diff
@@ -218,6 +229,11 @@ export class UmbBlockWorkspaceContext<LayoutDataType extends UmbBlockLayoutBaseM
 			throw new Error('Block Entries could not create block');
 		}
 
+		const blockType = this.#blockManager!.getBlockTypeOf(contentElementTypeId);
+		if (blockType?.label) {
+			this.setName(blockType.label);
```
nielslyngsoe, line 234: We are still missing the Label interpretation feature, meaning using the Label (template) would not necessary be very pretty.

How was that i v.13 did we show the interpolated label in the header of the modal? :-)

src/packages/block/block/context/block-manager.context.ts
rickbutterfield: Really odd, I already have that extension so not sure why it was so badly affected

src/packages/block/block/workspace/block-workspace.context.ts
rickbutterfield: Ah, you're right! The modal label shows the interpolated version of it. 🤦‍♂️ Shall I draft this until that gets resolved?

src/packages/block/block/workspace/block-workspace-editor.element.ts
```diff
@@ -4,13 +4,16 @@ import { UmbLitElement } from '@umbraco-cms/backoffice/lit-element';
 
 @customElement('umb-block-workspace-editor')
 export class UmbBlockWorkspaceEditorElement extends UmbLitElement {
-	//
+
 	@property({ type: String, attribute: false })
 	workspaceAlias?: string;
 
+	@property({ type: String, attribute: false })
```
nielslyngsoe, line 11: Hi @rickbutterfield 

I'm so happy that you are getting your hands dirty and getting into the code.

To prevent, too much prop drilling and hard connections, we encourage using the Context API. So in this case you would consume the Block Workspace Context, and observe the name, which should be available via an observable. 

In this way there is a bit of plumbing below in the routes that can be cleaned out :-)

I hope that makes sense :-)
Lets me know what you think and if you have ideas for better DX

src/packages/block/block/workspace/block-workspace.context.ts
nielslyngsoe: Well, we need the label soon.. so maybe you could leave those parts commented out and instead have the name set to the Element Type Name?

src/packages/block/block/workspace/block-workspace.context.ts
```diff
@@ -277,7 +293,10 @@ export class UmbBlockWorkspaceContext<LayoutDataType extends UmbBlockLayoutBaseM
 	}
 
 	getName() {
-		return 'block name content element type here...';
+		return this.#label.value;
+	}
+	setName(value: string): void {
```
nielslyngsoe, line 298: I view the Workspace Context as a public API for extensions(here under property editors, workspace views etc), so we would not like any of those to change the name. So in that perspective a setName method should not be available. :-)

It could be an internal one, like #setName but I would be fine by setting the value directly on the label, like this.#label.setValue(...)

leekelleher:
Hi @rickbutterfield. Thanks for the PR, it's been the basis of several discussions I've had with @nielslyngsoe about how the Block editor modal headline should be displayed; especially how it fits in with the new Umbraco-flavoured Markdown (UFM, #2045) feature we've been working on for Block labels.

We came to the conclusion that a dynamic label (based on the Block item data), may not be an ideal description for the intention of the Block editor modal, _(although noted, that how it currently is in v13)_, so we'll be exploring an alternative approach.

I'll close off this PR, _(only as we wouldn't merge it in as is)_, but it's definitely been useful for us to progress the conversation, which we should have a solution for soon. Thanks again, it's much appreciated!

PR #2018, opened 2024-06-19
Make pickers value undefined when empty

nielslyngsoe:
this change, makes pickers value become `undefined` in their empty state. previously they gave a empty string.
Which has been fine for Property Editors, as their Property Value Converters will resolve this, but when we store values as part of a larger JSON Value. Those does not kick in and then we end up getting the empty string stored which is not a value the server code in general accepts. As well the empty string is a confusing value in JS world, as we tend to just test for undefined.

This somehow is breaking if anyone where expecting the value to always be a string, as it can now be undefined as well. Otherwise it should be fine.

## Types of changes

## Checklist


madsrasmussen (approved):
LGTM

PR #2025, opened 2024-06-21
Feature: Allow dynamic import of icon for easier bundling

madsrasmussen:
## Description

Currently, we import icons based on a relative path. This causes some trouble when bundling the icon-registry module as files might not be located in the same place after bundling. This PR adds the extra option to "manually" import the icon file (the same way we do in manifests) like `path: () => import('./path/to/icon.js').` This way the bundler knows where the files are located and will handle them correctly.

## What to test
* Make sure that icons render correctly when running with Vite
* Make sure that icons render correctly when building for the CMS.
* All icons are still lazy loaded when they appear on the screen.

## Types of changes


iOvergaard (commented):
I like the premise of this. Can I ask what the need is to change all icons to .ts files?

madsrasmussen:
@iOvergaard The only reason was that I started to get type errors with js files. I don't know how we could avoid that before. 🤔 

Screenshot 2024-06-24 at 10 31 25

iOvergaard:
> @iOvergaard The only reason was that I started to get type errors with js files. I don't know how we could avoid that before. 🤔
> 
> Screenshot 2024-06-24 at 10 31 25

I think it worked because it was just a variable, and we additionally had Vite ignore it on the dev server with:
```
import(/* @vite-ignore */ iconPath)
```

iOvergaard (approved):

PR #2039, opened 2024-06-24
V14/bugfix/missing css vars

JesmoDev:
Adds missing css vars to the current user modal.

Before
image

After
image

## Types of changes

## Checklist

### DIFF
--- a/src/packages/user/current-user/modals/current-user/current-user-modal.element.ts
+++ b/src/packages/user/current-user/modals/current-user/current-user-modal.element.ts
@@ -84,6 +84,9 @@ export class UmbCurrentUserModalElement extends UmbLitElement {
 	static override styles: CSSResultGroup = [
 		UmbTextStyles,
 		css`
+			:host {
+				color: var(--uui-color-text);
+			}
 			#main {
 				display: flex;
 				flex-direction: column;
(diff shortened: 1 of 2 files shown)


src/packages/user/current-user/theme/current-user-theme-user-profile-app.element.ts
```diff
@@ -50,15 +50,28 @@ export class UmbCurrentUserThemeUserProfileAppElement extends UmbLitElement {
 		return html`
 			<uui-box headline="Theme">
 				<uui-tag slot="headline" look="placeholder">Experimental</uui-tag>
-				<uui-select label="Select theme" .options=${this._themes} @change=${this.#onThemeChange}></uui-select>
```
madsrasmussen, line 53: Can we make the change in the UI Library instead? Then will the select be styled correctly across the entire backoffice?

madsrasmussen (changes requested):

src/packages/user/current-user/modals/current-user/current-user-modal.element.ts
```diff
@@ -84,6 +84,9 @@ export class UmbCurrentUserModalElement extends UmbLitElement {
 	static override styles: CSSResultGroup = [
 		UmbTextStyles,
 		css`
+			:host {
```
nielslyngsoe, line 87: It would be better if you use the `class="uui-text"` on the first element in the render. so we avoid local color implementations. :-)

madsrasmussen: I have tried with the class but it doesn't work. I will accept the current solution.

madsrasmussen (approved):

PR #2045, opened 2024-06-25
Feature: Umbraco Flavored Markdown

leekelleher:
## Description

The artist formerly known as Dynamic Labels, (aka Label Templates), shall now be known as **Umbraco Flavored Markdown (UFM)**!

This feature uses Markdown syntax for label templates in Blocks (Block Grid, Block List) and Collection View columns (in Grid and Table views), as well as used to render Property descriptions.

The beauty of using Markdown for labels is that as well as offering basic text formatting, it natively supports the use of HTML markup, meaning that web components can be used to handle more complex label templating scenarios.

The Umbraco "flavor" of Markdown is to add lightweight layer of syntax that can be parsed into Umbraco web components, with access to relevant Context APIs and resources.

The UFM syntax uses single curly braces `{ }` to designate a component, here are examples...

1. `{=heading}` would be parsed and transformed into `<ufm-label-value alias="heading"></ufm-label-value>`
2. `{#general_chose}` would be parsed and transformed into `<umb-localize key="general_chose"></umb-localize>`

Each of those components will perform a specific task to render the relevant label.

Additional UFM components can be added using the `ufmComponent` extension type, like so...

```
{
	type: 'ufmComponent',
	alias: 'My.Markdown.CustomComponent',
	name: 'My Custom Markdown Component',
	api: () => import('./components/my-custom.component.js'),
	meta: {
		marker: '%',
	},
}
```

The corresponding JavaScript/TypeScript API would contain a method to render the custom label/markup.

```
export class MyCustomComponent implements UfmComponentBase {
	render(token: Tokens.Generic) {
		return `<ufm-custom-component text="${token.text}"></ufm-custom-component>`;
	}
}
```

## Types of changes


src/packages/ufm/components/document-name.component.ts
```diff
@@ -0,0 +1,15 @@
+import type { UfmComponentBase } from './ufm-component-base.js';
+import type { Tokens } from '@umbraco-cms/backoffice/external/marked';
+
+import './document-name.element.js';
+
+// eslint-disable-next-line local-rules/umb-class-prefix
+export class UfmDocumentNameComponent implements UfmComponentBase {
+	render(token: Tokens.Generic) {
+		return `<ufm-document-name alias="${token.text}" debug></ufm-document-name>`;
```
iOvergaard, line 9: Should there be a "debug" here?

src/packages/core/components/ufm-render/ufm-render.element.ts
```diff
@@ -0,0 +1,125 @@
...
+		return this.#context.getValue();
+	}
+
+	constructor() {
+		super();
+		this.#observe();
+	}
+
+	async #observe() {
+		this.observe(umbExtensionsRegistry.byType('ufmComponent'), async (manifests: Array<ManifestUfmComponent>) => {
```
nielslyngsoe, line 73: Here we should use some of the Extension Initializers, which would be more performant as they would only spin up new APIs. :-)

src/packages/ufm/components/document-name.component.ts
leekelleher: Good spot, I'll remove it. 👍

src/packages/core/components/ufm-render/ufm-render.element.ts
leekelleher: Thanks! Implemented in https://github.com/umbraco/Umbraco.CMS.Backoffice/pull/2045/files#diff-204dc0a9aa5ce0ecdf69b1551fc0db0ae55c02e7a1d40f7052d2d0fc76030927R71

nielslyngsoe (approved):
Looks and work very well 🥳

iOvergaard (approved):

quentinmeliard:
Just to be sure, there is absolutely no way of retrieving, for example, the name property of a media picker with this change ?

leekelleher:
@quentinmeliard With this particular change? No, not yet, but soon. ☺️

The UFM feature is a layer on top of Markdown, to add a syntax sugar for rendering Umbraco-based web components. While there munch functionally lacking from this initial feature release, it does lay the foundation for upcoming releases... and we do have several planned components, for document/media pickers, string transformation, array concatenation, etc.

In the meantime, it is possible to develop a custom UFM component for rendering dynamic labels as you need it.
<https://docs.umbraco.com/umbraco-cms/reference/umbraco-flavored-markdown#available-ufm-components>

PR #2061, opened 2024-06-28
Feature: Route not found + router features

nielslyngsoe:
Implementing Not Found views for our routers.

This includes making the router-slot more reactive, responding to late appearing Routes. As well making sure to cancel a Route Resolving if a new Route appears in the mean time. (its clear now, that the Router Slots routes was never made to be changing over time, but now I have made it so. )

This also fixes a problem when switching between variants, when returning to a variant that was opened earlier, you ended at a blank screen before. (Which was repported in issue as well, found it after this PR was made: ( Fixes https://github.com/umbraco/Umbraco-CMS/issues/16460 )

Be aware this affects almost the whole application, so please try to come up with some edge case scenario and test that.

Fixes https://github.com/umbraco/Umbraco-CMS/issues/16382

## Types of changes

## How to test?

Type wrong URLs and see the 'Not Found' page appearing.

As well testing late appearing and later removed Routes, you can add this condition to a extension of choice that provides a Route:
```
{
				alias: 'Umb.Condition.Switch',
				frequency: '2000'
			}
```

## Checklist


iOvergaard (commented):
It seems to work, but I honestly can't get the not found component to show up. It seems no matter where I type in a wrong URL, the UI corrects itself. For example, if I type a different non-existing dashboard pathname, it redirects me back to the default dashboard.

I saw the `**` path added to members so I tried to go to something like "/member-groups/collectionssss/table` but it corrected itself back to "collection".

The only time I see the not found component is when the backoffice is loading, when it takes a while for the router to set in, the component will briefly show in the background. That actually seems to make the experience worse because I initially think the route is wrong, but then the UI show up anyway. I tried to record it here:

2024-07-02 at 10 18 02 - Amaranth Bobcat

So in summary, it seems to work...? But I'm not seeing the big difference. But it definitely seems like it's waiting longer for late-appearing routes, which I think is a good thing...?

madsrasmussen:
I have merged this PR into this (#2075) and there seems to be an issue when deep linking to the "Content" workspace view. In my test cases, I always end up on the "Info"-view.

nielslyngsoe:
@iOvergaard There might be even more places where we need to implement a 'no-route route'. For now its visible when removing the unique of a path like this `section/content/workspace/document/edit/40381655-78b2-4d72-8005-4c61c9d59260`

it may help that Mads have now tested this(he has merged this PR into his) as well as he was been implementing conditions in his PR (https://github.com/umbraco/Umbraco.CMS.Backoffice/pull/2075) and after a minor correction for the part he specifically worked on it also work in that case. :-)

iOvergaard (approved):

PR #2063, opened 2024-06-30
[FIX] Scroll Bar / Panel Width Adjustment cursor

OwainWilliams:
You can't scroll in the side panel, only make the panel wider.

Example of the issue.
scrollBarNotWorking

You can only make the panel wider. 

## Motivation and context

From a users point of view, you can't scroll down to select any other items that might be visible in the left hand side.
The issue is the div with the class id divider-touch-area is sitting over the top of the scroll bar. The fix was to move this div 5px to the left and this then allows the user to scroll and still make the panel wider. 

## How to test?

See screenshot above, and the fix demo : 

scrollBarWorking

## Checklist

### DIFF
--- a/src/packages/core/components/split-panel/split-panel.element.ts
+++ b/src/packages/core/components/split-panel/split-panel.element.ts
@@ -294,7 +294,7 @@ export class UmbSplitPanelElement extends LitElement {
 		#divider-touch-area {
 			position: absolute;
 			top: 0;
-			left: 0;
+			left: 5px;
 			height: 100%;
 			width: var(--umb-split-panel-divider-touch-area-width);
 			transform: translateX(-50%);


iOvergaard (approved):
LGTM!

OwainWilliams:
@iOvergaard Will this be in the next RC?

iOvergaard:
@OwainWilliams There probably won't be any more RCs of 14.1 but it should land up in the final version of 14.1 👍

iOvergaard:
Hi @OwainWilliams, I noticed this pull request was not included in the automated GitHub release notes. I have put it in manually, though, as the fix should be there in 14.1 released yesterday.

OwainWilliams:
@iOvergaard thanks. Ye, I tested 14.1 and it is there. 👍

PR #2078, opened 2024-07-02
Bugfix: Show correct msg for ApiError

loivsen:
## Description

Shows a more detailed ApiError description
Are there any place I have missed?

## Types of changes

## How to test?
- Save something where it throws an ApiError: For example attempt to create a new Template with no name or alias.


src/packages/core/resources/resource.controller.ts
```diff
@@ -129,10 +141,11 @@ export class UmbResourceController extends UmbControllerBase {
 					default:
 						// Other errors
 						if (this.#notificationContext) {
+							const message = error.body?.errors ? this.#buildApiErrorMessage(error.body.errors) : undefined;
 							this.#notificationContext.peek('danger', {
 								data: {
 									headline: error.body?.title ?? error.name ?? 'Server Error',
-									message: error.body?.detail ?? error.message ?? 'Something went wrong',
+									message: message ?? error.body?.detail ?? error.message ?? 'Something went wrong',
```
iOvergaard, line 148: Would it make sense to forward the error messages in another property to the notification context so that you let that context handle the rendering and styling of the error messages? At least that way you could avoid defining `<div>`'s in the message property as you do on line 46?

src/packages/core/resources/resource.controller.ts
```diff
@@ -1,10 +1,11 @@
+import { html } from '@umbraco-cms/backoffice/external/lit';
+import type { UmbControllerHost } from '@umbraco-cms/backoffice/controller-api';
+import { UmbControllerBase } from '@umbraco-cms/backoffice/class-api';
+import { UmbContextConsumerController } from '@umbraco-cms/backoffice/context-api';
```
iOvergaard, line 4: I think this happened because the eslint/prettier config had some trouble. If you run the formatted again on this file, it should correct itself back (`npm i && npx eslint --write src/packages/core/resources/resource.controller.ts`).

iOvergaard (changes requested):
Nice to see the errors starting to be shown. I added some suggestions down here. I think we should consider how to translate these messages as they are only fallback texts from the server. Do you know if they forward some sort of error code, we could hook in to?

src/packages/core/resources/resource.controller.ts
```diff
@@ -1,12 +1,18 @@
...
 import { UmbControllerBase } from '@umbraco-cms/backoffice/class-api';
 import { UmbContextConsumerController } from '@umbraco-cms/backoffice/context-api';
+import {
+	UMB_NOTIFICATION_CONTEXT,
+	type UmbNotificationDefaultData,
+	type UmbNotificationOptions,
+} from '@umbraco-cms/backoffice/notification';
 import type { UmbDataSourceResponse } from '@umbraco-cms/backoffice/repository';
 
+export type ErrorMessageText = { property: string; messages: string[] };
```
iOvergaard, line 14: We are not using this type anymore, right?

loivsen: No 😊

src/packages/core/notification/layouts/default/notification-layout-default.element.ts
```diff
@@ -16,11 +24,46 @@ export class UmbNotificationLayoutDefaultElement extends LitElement {
...
+	static override styles = [
+		UmbTextStyles,
+		css`
+			.structured-list ul {
+				margin: 0;
+			}
+			.structured-list span {
+				display: block;
+				padding: var(--uui-size-3) 0 var(--uui-size-1);
+			}
```
iOvergaard, line 64: I would think that a `<p>` element gives approx. the same styling without having to force any display:block or padding. Have you tried that?

loivsen: `<p>` gives a lot of margin, about 24px top and bottom. This is all set with the `.uui-text` class. But at least wouldn't have to force display:block.
Originally I had `h2` on, thought it would make more sense HTML scematic wise, but that requires  even more styling due to our class.

iOvergaard (approved):

PR #2084, opened 2024-07-03
Bugfix: Remove html tags in validation localizations

loivsen:
## Description

The validator prints error messages as a simple string and does not accept html tags. This PR removes html tags from some localizations. In general we would like to avoid html tags inside localizations.

Fixes https://github.com/umbraco/Umbraco-CMS/issues/16504

## Types of changes


iOvergaard (changes requested):
I think this could be one of the cases where we do want HTML inside the string to outline what the localization is. The UmbLocalizeElement is able to render HTML in fact. So there must be a case where either a) that element is not used, or b) there is a missing `unsafeHTML` around a localize.term call somewhere.

Can you figure out where those keys are being used?

loivsen:
as we discussed, adding the solution in UUI library instead.

PR #2092, opened 2024-07-04
Bugfix: Auto generate alias

loivsen:
## Description

Should be working for workspaces and properties.

Re-enables autogeneration of an alias if the alias input become locked while it is empty 🤯🎉

## Types of changes

## How to test
- Check if alias autogeneration is enabled for new properties and workspaces.
- Check if alias autogeneration is disabled for existing properties and workspaces.
- Check that autogeneration is disabled and stays disabled when the alias has become unlocked.
- Check if autogeneration can be turned back on by emptying the alias and locking it.


src/packages/core/content-type/workspace/views/design/content-type-design-editor-property.element.ts
```diff
@@ -81,8 +83,32 @@ export class UmbContentTypeDesignEditorPropertyElement extends UmbLitElement {
...
+	#checkAliasAutoGenerate(unique: string | undefined) {
+		if (unique === this.#propertyUnique) return;
+		this.#propertyUnique = unique;
+		this.#autoGenerateAlias;
+
+		if (this.#context.getAlias()) {
+			this.#autoGenerateAlias = false;
+		}
+
+		/* TODO: Do we know if we are loading an existing property (do not auto generate alias) or added a new property (auto-generate only if matching expected alias)?
```
nielslyngsoe, line 97: I would say please check how the existing Backoffice works, make sure to make the experience match. :-)

leekelleher:
I've separated out the `<uui-input-lock>` task to its own PR, #2175.

Once that's merged in, I'll update this branch with the latest, then it can focus on the auto-generation task.

leekelleher (approved):
Finally got around to testing this out, it works better (or more consistent) than v13 did.

PR #2096, opened 2024-07-04
Chore: Circular check with madge

madsrasmussen:
## Description

Our current no-cycle Eslint plugin stopped working after we upgraded to Eslint 9. This is an alternative solution to check for circular dependencies until the plugin works again. 

The check is currently not added to the build script as there is a circular dependency we can't fix at the moment.

## How to test
`npm i` && `npm run check:circular`

The script will check the src folder for circular dependencies. It will currently find 4.

## Types of changes


leekelleher (approved):
Tested out, works as described, finds 4 circular dependencies! 💪🤯👍

PR #2101, opened 2024-07-05
hotfix: block grid live editing race condition

nielslyngsoe:
Makes Block Grid Editor only set its data if its not in Live Editing Mode. Cause otherwise they end up competing with each other and eventually out rule the corrections made by the Property Editor.

Fixes: https://github.com/umbraco/Umbraco-CMS/issues/16227

## Types of changes

This makes sure the Block Workspace observes data of the Property Editor while in create-state, in this way the corrections made of the Property Editor, like rowSpan setting, gets set and the Block Workspace receives that correction.

Only sets the Blocks data if its not in live editing mode. Cause setting these will eventually overwrite the corrections from the Property Editor.

## Checklist

### DIFF
--- a/src/packages/block/block/workspace/block-workspace.context.ts
+++ b/src/packages/block/block/workspace/block-workspace.context.ts
@@ -130,9 +130,74 @@ export class UmbBlockWorkspaceContext<LayoutDataType extends UmbBlockLayoutBaseM
 			'observeLayoutInitially',
 		);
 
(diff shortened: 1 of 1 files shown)


iOvergaard (approved):
I was able to reproduce the problem before applying this fix, and after changing to this branch and re-creating the block grid in the exact same way, I was no longer able to reproduce the bug! LGTM!

PR #2102, opened 2024-07-05
Bugfix: Show dialog when property not supported

loivsen:
## Description

Shows a dialog when a picked property is not supported in current workspace, such as ImageCropper in blocks.
When do we know something is not supported?

## Types of changes

### DIFF
--- a/src/packages/block/block/workspace/views/edit/block-workspace-view-edit-properties.element.ts
+++ b/src/packages/block/block/workspace/views/edit/block-workspace-view-edit-properties.element.ts
@@ -5,12 +5,17 @@ import { UmbTextStyles } from '@umbraco-cms/backoffice/style';
 import type { UmbContentTypeModel, UmbPropertyTypeModel } from '@umbraco-cms/backoffice/content-type';
 import { UmbContentTypePropertyStructureHelper } from '@umbraco-cms/backoffice/content-type';
 import { UmbLitElement } from '@umbraco-cms/backoffice/lit-element';
+import { UmbDataTypeItemRepository } from '@umbraco-cms/backoffice/data-type';
 
 @customElement('umb-block-workspace-view-edit-properties')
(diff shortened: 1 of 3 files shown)


src/packages/block/block/workspace/views/edit/block-workspace-view-edit-properties.element.ts
```diff
@@ -39,25 +49,74 @@ export class UmbBlockWorkspaceViewEditPropertiesElement extends UmbLitElement {
...
+			async (propertyStructure) => {
+				await this.#checkForDisallowed(propertyStructure);
 				this._propertyStructure = propertyStructure;
 			},
 			'observePropertyStructure',
 		);
 	}
 
+	async #checkForDisallowed(properties: Array<UmbPropertyTypeModel>) {
+		const { data } = await this.#dataTypeItemRepository.requestItems(
```
nielslyngsoe, line 80: I have been reading the code of this PR, and I would suggest to move this logic into ´umb-property-type-based-property´.

I think the approach would to make the ability of parsing a list of `unsupported-property-editor-schemas`.

As the umb-property-type-based-property already knows its data-type and the extension manifest then it does not need to do retrieve any data to make this happen. Also because I'm not fond of requesting the data type here just to check if its supported. Also the amount of code for this is quite a lot, and that is bad for maintainability and readability.

This would as well make it possible to easily implement similar in other places and ensure consistency.

A side note, is it really not possibly for us to support these, I really hopped the architecture for the temp files would enable us to support this.

src/packages/core/property/property/property.element.ts
```diff
@@ -359,6 +369,13 @@ export class UmbPropertyElement extends UmbLitElement {
 				display: block;
 			}
 
+			.not-supported {
```
nielslyngsoe, line 372: Ideally we should not make this with just a class, instead we should make a component for an alert-message.

PR #2105, opened 2024-07-08
V14: The option `usernameIsEmail` is not supported

iOvergaard:
## Description

This fixes a few things seeing that before, the option `usernameIsEmail` was not supported -- at least not set to `false`.

This adds a "username" field to the user profile settings as well as a repository/store combo to fetch the user configuration settings, which have now been added to the backend (new API models are included).

Fixes https://github.com/umbraco/Umbraco-CMS/issues/16512

## Types of changes

## Motivation and context

Ideally, we'd like to have some kind of model state on the user details model, but for now, we can get the information through the user configuration endpoint. The only scenario we cannot solve in the UI at the moment is a more complex one where a login provider is set to "deny local login". In that case, the server will refuse any username changes.

## How to test?

1. First spin up the backoffice and check that only the email field is shown and that you can change it (it should update both the `email` and `userName` properties).
2. Next, go to appsettings and set `Umbraco:CMS:Security:usernameIsEmail` to `false`
3. Go back to the backoffice and check that you'll now see both a "username" and an "email" property on the user details, and that you can change each independently on the other

image


iOvergaard:
Uh, I need to use your feature from #2106 as well, @leekelleher !

leekelleher (approved):

PR #2107, opened 2024-07-10
Bugfix: Resized media urls result in the request path being too long

iOvergaard:
## Description

Some web servers carry a limit to the length of the request path. This harms the media thumbnail generator, which is a GET endpoint. We cannot change the backend as that would be a breaking change, nor do we wish to do it because of the purity of REST. So we are going to make a network request for each thumbnail instead. This is exactly what the old Backoffice did as well.

There are some advantages to doing this besides fixing the web servers: We can now space out the calls and wait until the elements are visible before initiating the request -- we will look into some of this in subsequential PRs.

The thumbnails are now also being cached in memory through a store in a nested map based on the media key (GUID) and thumbnail options (width, height, mode). This results in far fewer network calls when you have a lot of media showing up in the media library.

Fixes https://github.com/umbraco/Umbraco-CMS/issues/16716

## Types of changes

## How to test?

1. Upload more than 50 media items in a single folder in the Media library
2. Check that no requests result in a 500 error (this error most likely won't happen on local Kestrel anyway)
3. Check that all medias have their thumbnail URLs generated in separate requests

### DIFF
--- a/src/packages/media/imaging/constants.ts
+++ b/src/packages/media/imaging/constants.ts
@@ -1,1 +1,2 @@
 export const UMB_IMAGING_REPOSITORY_ALIAS = 'Umb.Repository.Imaging';
+export const UMB_IMAGING_STORE_ALIAS = 'Umb.Store.Imaging';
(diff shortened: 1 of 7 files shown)


leekelleher (approved):
Tested out against IIS Express (Windows/Visual Studio), I could reproduce the issue and this fix now resolves the issue. 🚀

PR #2110, opened 2024-07-10
Test: Enable more FlatConfig ESLint rules

iOvergaard:
If an eslint plugin supports the new FlatConfig format natively, then we don't need to specify the rules, unless we want to overwrite them.

- This enables the recommendations of the plugins, that were previously enabled in v8.
- Specifies the rules of eslint-plugin-import since that doesn't work with FlatConfig but still supports v9
- Removes the rule declarations of "local-rules" since the recommended config now works
- Removes lit-a11y and storybook plugins since they do not support v9 at all


loivsen (approved):
Looks good to me

PR #2112, opened 2024-07-11
Bugfix: Use the EyeDropper Color Picker for choosing colors in blocks

loivsen:
## Description

Color picker for picking colors in blocks, rather than typing them manually 🥰
- Added some localizations
- Fixed a bug in blocks where the color picked for icons would not show

Fixes https://github.com/umbraco/Umbraco-CMS/issues/16737

## Types of changes


src/packages/core/components/icon/icon.element.ts
```diff
@@ -48,7 +72,9 @@ export class UmbIconElement extends UmbLitElement {
 	}
 
 	override render() {
-		return html`<uui-icon name=${ifDefined(this._icon)} style=${ifDefined(this._color)}></uui-icon>`;
+		return html`<uui-icon
+			name=${ifDefined(this._icon)}
+			style=${ifDefined(this._color ?? this._fallbackColor)}></uui-icon>`;
```
iOvergaard, line 77: Do you think we could use [styleMap](https://lit.dev/playground/#sample=examples/directive-style-map) to avoid some of the extra logic around calculating a CSS variable?

src/packages/core/components/icon/icon.element.ts
```diff
@@ -31,14 +41,28 @@ export class UmbIconElement extends UmbLitElement {
...
 
+	#setFallbackColorStyle(value: string) {
+		const alias = value.replace('color-', '');
+		const variable = extractUmbColorVariable(alias);
+		this._fallbackColor = alias
+			? variable
+				? `--uui-icon-color: var(${variable})`
+				: `--uui-icon-color: ${alias}`
+			: undefined;
+	}
```
iOvergaard, line 52: This seems to do almost the same thing as `#setColorStyle`, consider unifying the logic in a common function.

loivsen: There we go 😊

src/packages/core/components/icon/icon.element.ts
```diff
@@ -16,29 +16,48 @@ export class UmbIconElement extends UmbLitElement {
...
 	public get color(): string {
 		return this._color ?? '';
 	}
 
-	#setColorStyle(value: string) {
+	#setColorStyle(value: string, dominant = false) {
 		const alias = value.replace('color-', '');
 		const variable = extractUmbColorVariable(alias);
-		this._color = alias ? (variable ? `--uui-icon-color: var(${variable})` : `--uui-icon-color: ${alias}`) : undefined;
+		const color = alias ? (variable ? `--uui-icon-color: var(${variable})` : `--uui-icon-color: ${alias}`) : undefined;
```
iOvergaard, line 41: I am not completely sure I understand the logic here, if it finds a variable it uses that = great. If it doesn't find a variable, it tries to set the icon alias in the `--uui-icon-color` variable. Could you explain that a little bit?

src/packages/core/components/icon/icon.element.ts
```diff
@@ -16,29 +16,48 @@ export class UmbIconElement extends UmbLitElement {
...
+		if (dominant) {
+			this._color = color;
+		} else {
+			this._fallbackColor = color;
+		}
```
iOvergaard, line 46: Consider returning the color string instead, and let the consumers set it into whichever state property they would like. I think the it would be much more obvious what happens and when, if you did that.

src/packages/core/components/icon/icon.element.ts
```diff
@@ -16,29 +16,48 @@ export class UmbIconElement extends UmbLitElement {
...
-		const [icon, alias] = value ? value.split(' ') : [];
+		const [icon, color] = value ? value.split(' ') : [];
 
-		if (alias) {
-			this.#setColorStyle(alias);
+		if (color) {
+			this.#setColorStyle(color);
 		} else {
-			this._color = undefined;
+			this._fallbackColor = undefined;
```
iOvergaard, line 60: I think you don't need to explicitly set this to undefined if there is no color. The undefined value is the default of `_fallbackColor` anyway.

src/packages/core/components/icon/icon.element.ts
```diff
@@ -16,29 +16,48 @@ export class UmbIconElement extends UmbLitElement {
...
+	 * If a color has been set via the name property, this property will override it.
+	 * */
 	@property({ type: String })
 	public set color(value: string) {
-		if (!value) return;
-		this.#setColorStyle(value);
+		if (value) {
+			this.#setColorStyle(value, true);
+		} else {
+			this._color = undefined;
```
iOvergaard, line 31: I think you don't need to explicitly set this to undefined if there is no value. The undefined value is the default of `_color` anyway.

iOvergaard (changes requested):

src/packages/core/components/icon/icon.element.ts
loivsen: This would probably be less confusing with better naming. Basically, it doesn't necessary have to be an alias, so a variable name of "color" may make more sense.
This means that `icon.element` allows any color, not just a color alias. It just checks if the color is an alias for an existing color variable. Hope that makes sense.

This PR really shouldnt have been so much about the icon.element 😅 but I'll see about adding these changes.

src/packages/core/components/icon/icon.element.ts
loivsen: We also need to be able to reset the color value, if the color is removed again

src/packages/core/components/icon/icon.element.ts
loivsen: Just pushed a commit using stylemap now

iOvergaard (commented):
Looks really great, just one last comment on the undefined vs 'inherit' when there is no color.

iOvergaard (approved):

PR #2113, opened 2024-07-11
Feature: Collection: Sets label and icon of Workspace View

leekelleher:
## Description

> Attempts to fix:
> - https://github.com/umbraco/Umbraco-CMS/issues/16633

When setting a custom label and icon for a Collection (data-type), the workspace view's label and icon were not being updated.

This PR adds a `UmbWorkspaceEditorContext`, which moves logic from the `UmbWorkspaceEditorElement` to collect the manifests and build the routes & views.

The context exposes a method `.setWorkspaceViewMeta()` to enable a workspace view's label and icon be configured, (based on the workspace view's extension alias).

## Types of changes

## How to test?

1. Configure a document-type to use a collection, (Structure > Presentation > Collections)
2. In the corresponding data-type for the Collection, set the "Content app icon" and "Content app name" fields
3. Go to the document (either create new or existing). Has the icon and label changed?

## Notes

1. There is an edge-case, where a server-side migration (for v14) sets the default icon for a Collection to be [`icon-badge color-black`](https://github.com/umbraco/Umbraco-CMS/blob/release-14.0.0/src/Umbraco.Infrastructure/Migrations/Upgrade/V_14_0_0/MigrateDataTypeConfigurations.cs#L718), but that particular icon doesn't exist in the backoffice for v14. So I have added a check for this, and set it to `icon-grid`.

2. There is an annoying UI glitch, where the default icon and label (from the Collection manifest) is displayed, before changing to the custom icon and label. (Sometimes it happens faster, other times slow, **try setting the video playback speed to 0.25x**).

https://github.com/umbraco/Umbraco.CMS.Backoffice/assets/209066/6bbdaf22-4b3e-4430-962a-102f75cc6990

This is because the Workspace Editor loads in the Workspace View manifests first, then it can calculate the routes, then the Workspace View Collection is loaded in (from that route), which then has access to the data-type configuration and sets the custom icon and label.

3. I have commented out code to handle the `showContentFirst`, but I hit various issues that I'm not currently able to resolve. e.g. moving the "Content" workspace to the first position, but the route doesn't update, so it displays the "Collection" instead. There is also a similar UI glitch with the Workspace View tabs updating, (it shows "Collection" first, then suddenly swaps with "Content").

I'll mark this PR as a draft until we've addressed the above issues.

### DIFF
--- a/src/packages/core/workspace/components/workspace-editor/workspace-editor.context.ts
+++ b/src/packages/core/workspace/components/workspace-editor/workspace-editor.context.ts
@@ -1,0 +1,100 @@
+import type { ManifestWorkspaceView } from '../../types.js';
+import { createExtensionElement, UmbExtensionsManifestInitializer } from '@umbraco-cms/backoffice/extension-api';
+import { umbExtensionsRegistry } from '@umbraco-cms/backoffice/extension-registry';
+import { UmbArrayState } from '@umbraco-cms/backoffice/observable-api';
+import { UmbContextBase } from '@umbraco-cms/backoffice/class-api';
+import { UmbContextToken } from '@umbraco-cms/backoffice/context-api';
+import type { UmbControllerHost } from '@umbraco-cms/backoffice/controller-api';
+import type { UmbRoute } from '@umbraco-cms/backoffice/router';
+
+export type UmbWorkspaceEditorView = {
+	alias: string;
+	label: string;
+	icon: string;
+	pathName: string;
+	weight: number;
+};
+
+export class UmbWorkspaceEditorContext extends UmbContextBase<UmbWorkspaceEditorContext> {
+	#routes = new UmbArrayState<UmbRoute>([], (x) => x.path);
+	public readonly routes = this.#routes.asObservable();
+
+	#views = new UmbArrayState<UmbWorkspaceEditorView>([], (x) => x.alias);
+	public readonly views = this.#views.asObservable();
+
+	constructor(host: UmbControllerHost) {
+		super(host, UMB_WORKSPACE_EDITOR_CONTEXT);
+
+		new UmbExtensionsManifestInitializer(this, umbExtensionsRegistry, 'workspaceView', null, (workspaceViews) => {
+			const manifests = workspaceViews.map((view) => view.manifest);
+			this.#createRoutes(manifests);
+			this.#createViews(manifests);
+		});
+	}
+
+	#createRoutes(manifests: Array<ManifestWorkspaceView> | null) {
+		let routes: UmbRoute[] = [];
+
+		if (manifests?.length) {
+			routes = manifests.map((manifest) => {
+				return {
+					path: `view/${manifest.meta.pathname}`,
+					component: () => createExtensionElement(manifest),
+					setup: (component) => {
+						if (component) {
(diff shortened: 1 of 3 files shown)


src/packages/core/workspace/components/workspace-editor/workspace-editor.context.ts
```diff
@@ -0,0 +1,108 @@
...
+
+		const views = [...this.#views.getValue()].map((view) =>
+			view.alias === alias ? { ...view, ...(icon && { icon }), ...(label && { label }) } : view,
+		);
+
+		this.#views.setValue(views);
+
+		// TODO: Review how `showContentFirst` is implemented, and how to re-route/redirect the path.
+		// e.g. the "Content" tab shows the "Collection" view. [LK]
+		if (showContentFirst) {
```
nielslyngsoe, line 94: I think the only right way to do this, would be to manipulate the weight of the Collection based on this setting. (
Maybe assuming that content has a certain weight?)

Not completely sure how at the moment of writing this, but if feels wrong to sort the views, also notice how they can come and go depending on conditions. Atleast of now, then it should use the weight to sort and then make sure to sort(& change weight based on setting) in the `#createViews` method.

src/packages/core/workspace/components/workspace-editor/workspace-editor.element.ts
```diff
@@ -1,11 +1,12 @@
...
+import type { UmbWorkspaceEditorView } from './workspace-editor.context.js';
 import { css, customElement, html, nothing, property, repeat, state, when } from '@umbraco-cms/backoffice/external/lit';
-import { createExtensionElement, UmbExtensionsManifestInitializer } from '@umbraco-cms/backoffice/extension-api';
-import { umbExtensionsRegistry } from '@umbraco-cms/backoffice/extension-registry';
 import { UmbLitElement } from '@umbraco-cms/backoffice/lit-element';
 import { UmbTextStyles } from '@umbraco-cms/backoffice/style';
-import type { ManifestWorkspaceView } from '@umbraco-cms/backoffice/extension-registry';
 import type { UmbRoute, UmbRouterSlotInitEvent, UmbRouterSlotChangeEvent } from '@umbraco-cms/backoffice/router';
 
+const elementName = 'umb-workspace-editor';
```
nielslyngsoe, line 8: Is this a optimization we want to implement in general, I think it is a bit of an overload and I must admit i'm in for consistency. And using a variable twice use is not directly enough to trigger my need for this, especially because the later one is only used by TypeScript, so basically we are crating a const to use it once.

madsrasmussen: There have been many cases where there has been a mismatch between the registered element name and the type because we have renamed elements over time. I think this is a reasonable way to avoid that.

leekelleher:
With the [upcoming CMS repositories merge](https://github.com/umbraco/Umbraco-CMS/discussions/17269), I'll close this pull request off and create a new one post-merge.

umbracocommunity:
This pull request has been mentioned on **Umbraco community forum**. There might be relevant details there:

https://forum.umbraco.com/t/change-in-workspace-settings-for-listview-v16-arent-shown/4331/2

PR #2115, opened 2024-07-15
Bugfix: Date Picker with only "date" or "time" does not display correctly

iOvergaard:
## Description

The server always expects a regular DateTime string because the property value converter does not consider the config. The value will be unset if the string does not validate as a DateTime.

This PR aims to differentiate between input and output dates to communicate the expected value to the underlying date input component. The UI component expects the following:

| Format | Inner input | Output |
|--------|-------|--------|
| datetime-local | YYYY-mm-dd HH:ii(:ss) | YYYY-mm-dd HH:ii:ss |
| date | YYYY-mm-dd | YYYY-mm-dd HH:ii:ss |
| time | HH:ii(:ss) | YYYY-mm-dd HH:ii:ss |

Because of this, we have to patch the value coming out of the UI into a regular DateTime. Vice-versa we have to patch the value from the database into the expected _inner input_ value.

Fixes https://github.com/umbraco/Umbraco-CMS/issues/16779

## How to test

1. Configure a Date Picker with a format of `HH:mm:ss`
2. Test that it is possible to save a value
3. Test that the saved value is displayed correctly to the end-user
4. Verify that the updated unit tests make sense

## Screenshots

image

### DIFF
--- a/src/packages/property-editors/date-picker/property-editor-ui-date-picker.element.ts
+++ b/src/packages/property-editors/date-picker/property-editor-ui-date-picker.element.ts
@@ -42,13 +42,16 @@ export class UmbPropertyEditorUIDatePickerElement extends UmbLitElement implemen
 	@property()
 	value?: string;
 
+	@state()
+	private _inputValue?: string;
+
 	public set config(config: UmbPropertyEditorConfigCollection | undefined) {
 		if (!config) return;
 
 		// Format string prevalue/config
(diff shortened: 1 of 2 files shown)


iOvergaard:
Sorry, more information has been uncovered that we need to fix. I'll put this Pr back into draft.

leekelleher (approved):

PR #2118, opened 2024-07-17
Build: Upgrade Lit: 3.1.4, Monaco: 0.50, Marked: 13.0.2, uuid: 10.0

iOvergaard:
## Description

Upgrades npm dependencies to their latest version as of July 2024.

**Note:** There are many changed files because the icons have been re-generated.


leekelleher (approved):
Tested out various npm tasks and parts of the backoffice that would be affected by the dependency updates, (RTE, Markdown editor, UFM), and all appear to be working as expected.

PR #2121, opened 2024-07-17
Feature: Entity Bulk Action `moveTo` kind

leekelleher:
## Description

Adds a reusable Entity Bulk Action `moveTo` kind, and implements this for the Media collection bulk move feature.

## Types of changes

## How to test?

- Ensure that "Allow bulk move" is enabled in the Collection (data-type) configuration.
- In the Media section, (at the root section view, or inside a Folder detail view), select 1 or more media items.
- In the Collections actions toolbar (at the bottom), select the "Move to" action button.
- A media picker modal should appear to select the destination/target of the items, select an item.
  - If a non-folder is selected (e.g. child media items are not allowed, an error will be displayed).
- If a folder is selected, (e.g. child media items are allowed), then the items will be moved, a success notification should be displayed and the Media Collection UI should be automatically refreshed.
- Go to the destination/target folder and check that the media items are there.


iOvergaard (approved):
Tested and works.

PR #2122, opened 2024-07-17
Feature: Document entity bulk action "move to"

leekelleher:
## Description

Implements the Document collection bulk move feature, (as a Entity Bulk Action `moveTo` kind).

## Types of changes

## How to test?

- Ensure that "Allow bulk move" is enabled in the Collection (data-type) configuration.
- In the Document section, (at the root section view, or inside a Folder detail view), select 1 or more document items.
- In the Collections actions toolbar (at the bottom), select the "Move to" action button.
- A document picker modal should appear to select the destination/target of the items, select an item.
  - If a non-folder is selected (e.g. child document items are not allowed, an error will be displayed).
- If a folder is selected, (e.g. child document items are allowed), then the items will be moved, a success notification should be displayed and the Document Collection UI should be automatically refreshed.
- Go to the destination/target folder and check that the document items are there.


iOvergaard (approved):
LGTM

PR #2125, opened 2024-07-17
Feature: Document entity bulk action "duplicate to"

leekelleher:
## Description

Implements the Document collection bulk duplicate (copy) feature, (as a Entity Bulk Action `duplicateTo` kind).

To note, I haven't implemented the Media equivalent feature, as the "copy" endpoint (for media) doesn't exist in the Management API. 

## Types of changes

## How to test?

- Ensure that "Allow bulk duplicate" is enabled in the Collection (data-type) configuration.
- In a document that has a Collection enabled, select 1 or more document items.
- In the Collections actions toolbar (at the bottom), select the "Duplicate to" action button.
- A document picker modal should appear to select the destination/target of the items, select an item.
  - If an item is selected where child document items are not allowed, an error will be displayed.
  - If an item is selected where child document items are allowed, then the items will be copied/duplicated, a success notification should be displayed and the Document Collection UI should be automatically refreshed.
- Go to the destination/target folder and check that the duplicated document items are there.

### DIFF
--- a/src/packages/property-editors/collection/config/bulk-action-permissions/permissions.element.ts
+++ b/src/packages/property-editors/collection/config/bulk-action-permissions/permissions.element.ts
@@ -76,7 +76,7 @@ export class UmbPropertyEditorUICollectionPermissionsElement
 			<uui-toggle
 				?checked=${this.value.allowBulkCopy}
 				@change=${(e: UUIBooleanInputEvent) => this.#onChange(e, 'allowBulkCopy')}
-				label="Allow bulk copy (content only)"></uui-toggle>
+				label="Allow bulk duplicate (content only)"></uui-toggle>
 			<uui-toggle
 				?checked=${this.value.allowBulkMove}
 				@change=${(e: UUIBooleanInputEvent) => this.#onChange(e, 'allowBulkMove')}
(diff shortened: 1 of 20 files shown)


src/packages/property-editors/collection/config/bulk-action-permissions/permissions.element.ts
```diff
@@ -76,7 +76,7 @@ export class UmbPropertyEditorUICollectionPermissionsElement
 			<uui-toggle
 				?checked=${this.value.allowBulkCopy}
 				@change=${(e: UUIBooleanInputEvent) => this.#onChange(e, 'allowBulkCopy')}
-				label="Allow bulk copy (content only)"></uui-toggle>
+				label="Allow bulk duplicate (content only)"></uui-toggle>
```
iOvergaard, line 79: I was surprised to learn we don't have localization keys for these. I guess that's a good up-for-grabs task :)

iOvergaard (approved):

src/packages/property-editors/collection/config/bulk-action-permissions/permissions.element.ts
leekelleher: Good point. 👍

PR #2126, opened 2024-07-18
Remove all filter option from User Status

NguyenThuyLan:
## Description
Fixes https://github.com/umbraco/Umbraco-CMS/issues/16252

Remove the "All" filter option from User Status


iOvergaard (approved):
LGTM!

PR #2131, opened 2024-07-19
Chore: Do not redefine `clamp`

iOvergaard:
## Description

The UI Library already exports a `clamp` function, so use that instead of redefining it.

Ref https://github.com/umbraco/Umbraco.CMS.Backoffice/pull/1021#issuecomment-2118387247


leekelleher (approved):
LGTM 👍

PR #2132, opened 2024-07-22
Bugfix: Data Type selection for Property Editors without schema in manifests

leekelleher:
## Description

Whilst working on the Code Editor property-editor #1227, the `propertyEditorSchemaAlias` uses the `Umbraco.Plain.String` schema (new in v14, to handle the database-storage/value-conversation), I'd noticed that when creating a new Data Type and selecting the new property-editor, then nothing would appear in the UI. Upon a subsequent selection, the configuration fields for the new property-editor would be displayed corrected.

Debugging the scenario, I found that if a property-editor uses a `propertyEditorSchemaAlias` that is not defined in the extension manifests (of `propertyEditorSchema` type), then on the initial selection, the UI would be empty... as the schema can not be found and the `defaultPropertyEditorUiAlias` would be `null`.

> [!NOTE]
> On a side-note, this means that a property-editor schema is bound to a `defaultPropertyEditorUiAlias`. 😨 
> This is problematic, as if a property-editor would like to reuse an existing schema. Associated bug:
> - https://github.com/umbraco/Umbraco-CMS/issues/16808

In the meantime, I found an interim workaround, by putting a null-reference check for the `defaultPropertyEditorUiAlias`, the property-editor's configuration fields are displayed in the Data Type after the initial selection.

## Types of changes

## How to test?

- Once PR #1227 has been merged in.
- Create a new Data Type item, press the "Select a property editor" button.
- Select "Code Editor", press the Select button.
- The Code Editor configuration fields will be displayed.


iOvergaard (approved):
It works and looks good to me.

PR #2134, opened 2024-07-23
fix: allow datepicker to send empty values

iOvergaard:
## Description

This fixes https://github.com/umbraco/Umbraco-CMS/issues/16793

The Datepicker UI should be allowed to send empty values to clear the saved value.


leekelleher (approved):
Tested out, works as expected! 👍

PR #2136, opened 2024-07-23
Bugfix: Textarea does not show validation

iOvergaard:
## Description

The textarea property editor ui should show validation. This is fixed by converting the element into a form mixin that supports validation.

Fixes https://github.com/umbraco/Umbraco-CMS/issues/16436

## How to test

1. Set the "max length" of a textarea data type
2. Verify that a validation message is shown in the UI

## Screenshots

image


leekelleher (approved):
Tested out, works when the document is being saved.

PR #2137, opened 2024-07-24
Bugfix: Show the action button inside a button-group only if there are extensions registered

iOvergaard:
## Description

By moving the calculation of the extension controllers from the action-menu element into the action element itself, we can save a few states, simplify the calculation, and conditionally render the action-menu element ensuring, that the UUI styling is correctly applied

Fixes https://github.com/umbraco/Umbraco-CMS/issues/16362


leekelleher (approved):
Tested out, LGTM! 🚀

PR #2146, opened 2024-07-27
Fixes with `umb-route-slot` inside routed modal

enkelmedia:
## Description
This provides a potential fix for the issue described here: https://github.com/umbraco/Umbraco-CMS/issues/16834

I'm thinking that the problem might be that the event that is fired when the route changes fires before the `umb-route-slot` inside the modal has been loaded. I tried some hacks with `setTimeout` to trigger the "routes"-attribute to re-load the routes in the modal. This worked so i figured that maybe if we set the routes of the underlying `router-slot` when a new route is loaded it might solve the problem.

I made a small change to the `firstUpdated` method of `UmbRouterSlotElement` and after this, the scenario that I was having trouble with works.

umb-issue-routes-fix

## Types of changes

## How to test?
I've also added a example-folder to the PR, so just run 

```
npm run example
```

And pick `modal routed`

## Checklist


nielslyngsoe:
Hi @enkelmedia 

I had a look at this one, there is clearly a problem. I did not manage to boil it down just jet.

But its clear that the modal does not act correct.

For starters it should go away when navigating browser history.

nielslyngsoe (approved):
Hi @enkelmedia Thanks for this Example and highlighting this problem. I've debugged a bit more and corrected a few things and now things are stable. Thanks

PR #2148, opened 2024-07-30
Use the right property editor UI alias for config options

nielslyngsoe:
Fixes https://github.com/umbraco/Umbraco-CMS/issues/16808

The code was a bit messy, so i refactored so the 'flow' got simpler.

### DIFF
--- a/src/packages/data-type/workspace/data-type-workspace.context.ts
+++ b/src/packages/data-type/workspace/data-type-workspace.context.ts
@@ -32,7 +32,7 @@ import {
 type EntityType = UmbDataTypeDetailModel;
 
 /**
- * @class uUmbDataTypeWorkspaceContext
+ * @class UmbDataTypeWorkspaceContext
  * @description - Context for handling data type workspace
  * There is two overall code flows to be aware about:
  *
--- a/src/packages/data-type/workspace/data-type-workspace.context.ts
+++ b/src/packages/data-type/workspace/data-type-workspace.context.ts
@@ -30,6 +30,24 @@ import {
 } from '@umbraco-cms/backoffice/entity-action';
 
 type EntityType = UmbDataTypeDetailModel;
+
+/**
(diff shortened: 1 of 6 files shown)


src/packages/data-type/workspace/data-type-workspace.context.ts
```diff
@@ -301,9 +328,15 @@ export class UmbDataTypeWorkspaceContext
...
 
+	getPropertyEditorSchemaAlias() {
+		return this.#currentData.getValue()?.editorAlias;
+	}
 	setPropertyEditorSchemaAlias(alias?: string) {
 		this.#currentData.update({ editorAlias: alias });
 	}
+	getPropertyEditorUiAlias() {
+		return this.#currentData.getValue()?.editorUiAlias;
+	}
```
bjarnef, line 339: @nielslyngsoe would it be simpler to havd a single function `getPropertyEditor()` which retun both properties?

So if both it needed one only has to call a single function.

```
getPropertyEditorUiAlias() {
const currentData = this.#currentData.getValue()?.editorUiAlias;

if (!currentData) return null
return { currentData.editorAlias, currentData.editorUiAlias }
	}
```

leekelleher (approved):
Tested out, works as expected, LGTM! 🚀 

(I added the localization for "No configuration". 👍 )

src/packages/data-type/workspace/data-type-workspace.context.ts
nielslyngsoe: Hi @bjarnef 

I guess you can say that, but coming from a general architectural point of view then we have many 'get...' methods of our different contexts. And if we had to make these different combinations then we would end up with a lot of code.
I would say if there was a very specific, highly used scenario where we needed both, then it would make sense. But currently I'm not sure even sure we use this method. So I do not thing we need this kind of optimization currently.

PR #2157, opened 2024-07-31
Use constant for language workspace alias

bjarnef:
## Description

This add a constant for language workspace alias similar to other `ManifestWorkspaces`.
However it doesn't really seems consistent. Some are included in `manifests.ts` as here:
https://github.com/umbraco/Umbraco.CMS.Backoffice/blob/4e840b9de043f8ddb03352897c5ef2a575ca8776/src/packages/documents/document-blueprints/workspace/manifests.ts#L14C14-L14C52

while other are just in a constant.ts as here:
https://github.com/umbraco/Umbraco.CMS.Backoffice/blob/4e840b9de043f8ddb03352897c5ef2a575ca8776/src/packages/user/user/workspace/constants.ts

or  `index.ts`:
https://github.com/umbraco/Umbraco.CMS.Backoffice/blob/4e840b9de043f8ddb03352897c5ef2a575ca8776/src/packages/block/block/workspace/index.ts#L5

It would be great if it could be more consistency and easier to work with codebase.

Some places I noticed we have  workspace under `workspace/entity` and `entity/workspace`.

I also noticed the fallback language picker seems broken at language and under user section.
It set the value, but somehow `items` always seems to be an empty array.

image

## Types of changes

## Checklist

### DIFF
--- a/src/packages/language/workspace/language/constants.ts
+++ b/src/packages/language/workspace/language/constants.ts
@@ -1,0 +1,1 @@
+export const UMB_LANGUAGE_WORKSPACE_ALIAS = 'Umb.Workspace.Language';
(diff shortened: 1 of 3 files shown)


madsrasmussen (approved):
Hi @bjarnef 

Thanks for the PR. 

Yes, you are correct that we have inconsistencies between where we put the constants. We are currently moving all constants inside manifests and files with implementations (mostly contexts) to a constants file. This makes it possible to import the constant without loading the entire implementation.

 I have updated your PR with this change.
 
 Thanks 🥳

PR #2169, opened 2024-08-05
Bugfix: Document Type Picker - enable boolean properties

leekelleher:
## Description

I'd noticed when configuring a Block type (in Block Grid, Block List or RTE Blocks), the Document Type picker UI was ignoring the `"onlyPickElementTypes"` configuration. It turns out that the `elementTypesOnly` property wasn't reading the attribute value.

Also, in a Webhook edit workspace, the "Content Type" option should only be enabled for `documentTypesOnly`, but element-types were selectable.

## Types of changes

## How to test?

**Blocks**
Go to any of the Block editor data-type configurations, create a new or select an existing Block type, in the configuration modal panel, open the document-type picker modal for either "Content model" or "Settings model"... you should see that only element-types are selectable.

**Webhooks**
In the Webhooks section, create a new or select an existing webhook, open the document-type picker modal for "Content Type"... you should see that only document-types are selectable.


madsrasmussen (approved):
I have made a little change to the PR. 

Normally we want attributes to be kebab-cased, so the two attributes would be: element-types-only etc. I was lazy and changed so the values are set as properties instead of attributes.

PR #2170, opened 2024-08-05
multiple-app-languages-condition

nielslyngsoe:
Only shows App Language Selector when more than one app language available.


nielslyngsoe:
Hi @madsrasmussen 

When you review this, then please have a look at the current data-relation between the LanguageCollectionRepository and the act of adding or removing a language. Currently that seems not to update at runtime, meaning if re-configured the user needs to refresh their browser.

Thanks :-)

madsrasmussen (approved):
Looks good to me. I will leave the missing runtime language updates for another PR.

PR #2174, opened 2024-08-06
Chore: Lint JsDocs

madsrasmussen:
## Description

This PR adds a linter to the JS doc comments. This will ensure that we don't write JS docs that don't work and that we follow the same guidelines. 

I have chosen to run with the recommended rules for javascript instead of the typescript version as the typescript version prevents us from having types in JS docs. I think we should keep the possibility of using js docs as an alternative to typescript.

I have run the auto fixer as part of the PR. There are still a bunch of warnings we can fix over time.

## What to test
`npm i` && `npm run lint`

## Types of changes


leekelleher (approved):
Tested out. I get a whole bunch of warnings, (3206 of them; as you say we can fix over time 💪), so yes the configuration works. 👍

PR #2167, opened 2024-08-02
Block type custom view guide

nielslyngsoe:
This PR brings:

- Umb Manifest Icon
- Manifest Viewer Modal.
- Block Custom View button implemented in Block Types of Block Grid Editor Data Type Configuration.
- View the manifest that matches the Block Type.
- If none, then a button that can generate a manifest that will match this Block Type.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Chore (minor updates related to the tooling or maintenance of the repository, does not impact compiled assets)

## Checklist

- [ ] If my change requires a change to the documentation, I have updated the documentation in this pull request.
- [ ] I have read the **[CONTRIBUTING](<(https://github.com/umbraco/Umbraco.CMS.Backoffice/blob/main/.github/CONTRIBUTING.md)>)** document.
- [ ] I have added tests to cover my changes.

PR #2179, opened 2024-08-06
Manifest Viewer modal: Replaces `JSON.stringify` function

leekelleher:
@nielslyngsoe Following on from #2167, I've experimented with serializing the manifest data to JSON string using an alternative approach from `JSON.stringify`, as I'd noticed that all properties and values are wrapped with strings (which is valid JSON, but not the most copy-n-paste friendly).


nielslyngsoe (approved):
Looks good, i trust it works :-)

PR #2185, opened 2024-08-07
Chore: temp solution: move manifest viewer files to the blocks package

madsrasmussen:
The Manifest Viewer in the extension-registry module wants to import the modal token from the modal module but the modal module is not loaded in the browser because it exists later in the import map.

As the Manifest Viewer is currently only used in the blocks package. The temp solution is to move the files to the blocks package.


leekelleher (approved):
Tested out, works for me! 🚀

PR #2189, opened 2024-08-07
Feature: Data Type / Property Editor UI picker modal amends

leekelleher:
## Description

Part feature, part bugfix, part chore... I've refactored the Data Type / Property Editor UI picker modals!

A few things...

- Aligned the property-editor UI groups aliases (in manifests) with v13, (we had a lot under "common" for v14)
- Deprecated `submitLabel` property (on `UmbDataTypePickerFlowModalData` and `UmbPropertyEditorUIPickerModalData` interfaces)
- Refactored/fixed the grouping code, to sort items alphabetically, and friendly group labels (`fromCamelCase`)
- Fixed the grouping code when filtering data-types, (performs a lookup against the property-editor UI data)

## Types of changes

## How to test?

- In the Document Type workspace > "Add property" > "Select property editor"...
  - Notice the property-editor UI grouping labels
  - Try filtering, e.g. "text", notice the "Available configurations" are now grouped
- Data Type workspace > "Select a property editor"... notice the property-editor UI grouping labels


src/packages/core/modal/token/property-editor-ui-picker-modal.token.ts
```diff
@@ -1,8 +1,6 @@
 import { UmbModalToken } from './modal-token.js';
 
-export interface UmbPropertyEditorUIPickerModalData {
-	submitLabel?: string;
```
madsrasmussen, line 4: Should this be marked as deprecated instead of removed? Won't it break the typescript compiler if someone uses it?

madsrasmussen (changes requested):
It looks good and works as described 🎊 . I only have one comment about the removed submitLabel field.

src/packages/core/modal/token/property-editor-ui-picker-modal.token.ts
leekelleher: I'd considered it, I'd made the assumption that no one would be using `UmbPropertyEditorUIPickerModalData` externally yet... but you never know, so yes good point. I'll update with a deprecation. 👍

PR #2190, opened 2024-08-08
Chore: Bundle Log Viewer Package

madsrasmussen:
See https://github.com/umbraco/Umbraco.CMS.Backoffice/pull/1977 for more details about the bundling changes.

This PR bundles the Log Viewer Package including all modules.

## Types of changes

### DIFF
--- a/package.json
+++ b/package.json
@@ -137,6 +137,7 @@
 		"./src/packages/documents",
 		"./src/packages/health-check",
 		"./src/packages/language",
+		"./src/packages/log-viewer",
 		"./src/packages/media",
 		"./src/packages/members",
 		"./src/packages/models-builder",
--- a/package-lock.json
+++ b/package-lock.json
@@ -18,6 +18,7 @@
 				"./src/packages/documents",
 				"./src/packages/health-check",
 				"./src/packages/language",
+				"./src/packages/log-viewer",
 				"./src/packages/media",
 				"./src/packages/members",
 				"./src/packages/models-builder",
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -137,6 +137,7 @@
 		"./src/packages/documents",
 		"./src/packages/health-check",
 		"./src/packages/language",
+		"./src/packages/log-viewer",
```
bjarnef, line 140: @madsrasmussen ideally `documents`, `members`, `packages`, `property-editors`, `tags` should be in singular form as well?

madsrasmussen: Hi Hi,

Both yes and no 😄 Ideally I would like a name for the "Documents" package that indicates that it covers more than just documents. It includes everything needed for "Document Management" - Documents, Document Types, Blueprints, etc. The obvious one would be to call it Document Management but then we have to add "Management" to many of the packages which also feels a bit off.

If you have any suggestions we are all ears 😄

madsrasmussen: But everything inside a package should be singular. I can see we still miss to rename folders here and there.

bjarnef: Ahh okay 😅
Hmm 🤔 `document-editor`, `document-editing` .. but I guess it is same challenge as with `management`.

However I think some of the other packages also cover many different parts like `core` which isn't named `cores` 😁

madsrasmussen: Yes, you are correct. I just wanted to avoid renaming it twice but it might be better than the inconsistent naming 👍

leekelleher (approved):

PR #2195, opened 2024-08-08
Chore: Delete Object Type files

madsrasmussen:
## Description

These files are a leftover from when a relation type was editable from the UI. They are not exported anywhere or registered in the browser so they should be safe to delete.

## Types of changes


madsrasmussen:
Apparently, they are exported. I will move them to the core instead of a "standalone package".

PR #2196, opened 2024-08-08
Chore: Move object type files to core

madsrasmussen:
## Description

These files are a leftover from when a relation type was editable from the UI. They aren't used anywhere but they are unfortunately exported. They are currently under "packages" but aren't set up as a package. I have moved them to the core package until they can be deleted. They have their own import map entry so it won't affect imports.

## Types of changes


leekelleher (approved):

leekelleher (changes requested):
Sorry, I tested against the wrong branch, (I'd confused with #2195)! 🤦 

I'm getting an error locally and on the GitHub actions...

```
Error: Could not find export: ./object-type -> ./dist-cms/core/object-type/index.js in the build output.
```

I'm unsure what else needs to be changed, as `package.json` looks correct to me.

madsrasmussen:
@leekelleher Saved by the validate package script 💯 The path to the object-type module was wrong in `package.json`. It was missing the `packages` part.

leekelleher (approved):
Re-tested, all builds successfully! 🚀

PR #2206, opened 2024-08-12
Feature: System for searching in Pickers

madsrasmussen:
This PR adds the next step in the journey of searchable pickers.

It introduces:
* a Picker Context
* a Picker Search Manager
* a new element to render a picker search field
* a new element to render a picker search result
* a new extension point to register a picker search result item.

### General idea:
A modal implements a "picker ui". It currently consists of 2 parts: The selection ui + a search ui. These UIs connect through a "Picker Context". The picker context implements two managers: a selection manager and a search manager.

A Picker reuses the already registered "Search Provider" to handle searching.

It is now possible to add a "searchProviderAlias" as part of a Picker modal token. This will enable the search functionality. It is also possible to set any search query params that should be sent to the Search Provider as part of the search. This will become handy when we need to filter the search. Search for element types, exclude trashed items, scope the search from this tree item, etc.

**Next steps for searching:**
* Enable search in all supported pickers when Management Api supports needed filter options.

**Next steps for pickers**
* Introduce a "Picker" element that combines the "selection ui" with the "search ui" for easy reusability.
* Reusable "Collection Item Pickers" - Like the tree one.

### What to test:
* The search is currently enabled on the Document Picker, The document configuration for the Content Picker.
* The already searchable Member Picker has also been updated to use the same system.
* Make sure you can pick items
* Make sure you can search for items
* Make sure your selection is persisted between "selection ui" and "search UI".


leekelleher (approved):
Tested out, looking great! 🚀 

I made a UI amend, swapped the result item's `<uui-ref-node>` with `<umb-ref-item>`, as that removes the node-name link hover/underline and it also handles the icon part.

PR #2209, opened 2024-08-14
Chore: Aligns "Picker Input Context" class names

leekelleher:
## Updated description

Following the Picker Search feature work, (PR #2206) and [@madsrasmussen's feedback](https://github.com/umbraco/Umbraco.CMS.Backoffice/pull/2209#discussion_r1720205507) on this original PR, the `Umb{entity}PickerContext` class names have become confusing, so to clarify this, I've aligned the `Umb{entity}PickerInputContext` class names (note, added "Input") to be clearer.

With export aliases for backwards-compatibility, along with a deprecation notices.

## Original description

I'd noticed that the "[input-member.context.ts](https://github.com/umbraco/Umbraco.CMS.Backoffice/blob/v14.2.0-rc2/src/packages/members/member/components/input-member/input-member.context.ts#L7)" and "[input-member-group.context.ts](https://github.com/umbraco/Umbraco.CMS.Backoffice/blob/v14.2.0-rc2/src/packages/members/member-group/components/input-member-group/input-member-group.context.ts#L7)" both export a type called `UmbMemberPickerContext`, which seems to be a copy & paste typo in the member group one.

This PR corrects the member group export and adds an alias for backwards-compatibility, along with a deprecation notice.

## Types of changes


src/packages/members/member-group/components/input-member-group/input-member-group.context.ts
```diff
@@ -4,8 +4,11 @@ import { UMB_MEMBER_GROUP_PICKER_MODAL } from '../member-group-picker-modal/memb
 import { UmbPickerInputContext } from '@umbraco-cms/backoffice/picker-input';
 import type { UmbControllerHostElement } from '@umbraco-cms/backoffice/controller-api';
 
-export class UmbMemberPickerContext extends UmbPickerInputContext<UmbMemberGroupItemModel> {
+export class UmbMemberGroupPickerContext extends UmbPickerInputContext<UmbMemberGroupItemModel> {
```
madsrasmussen, line 7: I have come to realize that we will at some point end in a name collision when we start implementing a "Picker" extension because these class names aren't as specific as the class they extend. So I think we should use this opportunity to make it a bit more explicit and call it `UmbMemberGroupPickerInputContext`.

see #https://github.com/umbraco/Umbraco.CMS.Backoffice/pull/2206 for more context

madsrasmussen (changes requested):

src/packages/members/member-group/components/input-member-group/input-member-group.context.ts
leekelleher: I've updated all the class names that extend from the `UmbPickerInputContext` class.

madsrasmussen (approved):

PR #2213, opened 2024-08-15
[v14.2.0-RC] Fix Block Grid `layoutStylesheet` not using correct URL

rickbutterfield:
## Description
When configuring a `layoutStylesheet` against a BlockGrid datatype, the path to the CSS isn't resolved correctly when the front end is built and included in the NuGet package for release. The URL it is trying to be loaded from includes the backoffice path as part of the URL.

When running the Backoffice project locally, `appUrl` will always resolve to `http://localhost:5173`, whereas when running within the context of a C# project, `appUrl` will resolve to the absolute backoffice path, e.g. `https://localhost:44338/umbraco`.

## Types of changes

## Motivation and context
See https://github.com/umbraco/Umbraco-CMS/issues/16897.

Fixes the above linked issue so that BlockGrid layouts can be rendered correctly in the backoffice, and in Custom Views, with applied CSS.

This also fixes the same issue with Block icons.

## How to test?
Will need to be tested against a built version of the backoffice running in a .NET project (unsure how to do this, sorry!)

Unable to show screenshots as this only affects the built project when running on a .NET project

## Checklist

### DIFF
--- a/src/mocks/data/data-type/data-type.data.ts
+++ b/src/mocks/data/data-type/data-type.data.ts
@@ -702,6 +702,10 @@ export const data: Array<UmbMockDataTypeModel> = [
 				alias: 'blockGroups',
 				value: [{ key: 'demo-block-group-id', name: 'Demo Blocks' }],
 			},
+      {
+        alias: 'layoutStylesheet',
+        value: '/wwwroot/css/umbraco-blockgridlayout.css'
+      },
 			{
 				alias: 'blocks',
 				value: [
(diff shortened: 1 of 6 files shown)


nielslyngsoe:
Hi @rickbutterfield

Thanks a lot for reporting this, and fixing it.

We did get a double slash in our Development setup with Vite, but I have fixed that by removing last slash for the appUrl instead. So the responsibility is switched.

Let me know what you think about this change, and then I'm ready to approve and merge :-)

Thanks in advance :-)

rickbutterfield:
Looks great @nielslyngsoe! Appreciate you reviewing this so quickly.

nielslyngsoe (approved):
Fantastic PR!

rickbutterfield:
@nielslyngsoe is this not going to be targeted for the final release of 14.2? Custom Views are a bit broken without this

PR #2214, opened 2024-08-16
Block Workspace: Inherit Property VariantId

nielslyngsoe:
Parse VariantId from Property Context to Block Workspace to its Block DataSet Variant ID.

Though Blocks Data is invariant, it can take part of a varying Property, making the data belong to a specific Variant.

This should result in the following cases:

Document with invariant Property, will provide a Invariant Variant ID for the Blocks Workspace.

Document with varying Property, will inherit the culture(& eventually segment) a provide it for the Blocks Workspace.

Fixes https://github.com/umbraco/Umbraco-CMS/issues/16859

## Types of changes

## Motivation and context

This is needed for Property Editors that want to use the Culture for their responsibility. That could be picking other content, then that should be of the same culture as the property.

## Checklist


madsrasmussen (approved):
It works as expected 👍

PR #2218, opened 2024-08-19
Chore: Combine settings files

madsrasmussen:
## Description

We had settings files located in 2 places. This PR combines them into one. We only exported from one of the places so this should be a safe merge.

## Types of changes


leekelleher (approved):
Tested out with a build and test, Settings section still working as expected. 👍

PR #2234, opened 2024-08-23
Feature: FileUploadPreview Extension

loivsen:
## Description

Previously known as Media Preview.
Extension type for creating file previews for the FileUpload property editor ui.

## Types of changes


src/packages/media/media/components/input-upload-field/preview/manifest.ts
iOvergaard, line 1: I think we need to add a default weight for our built-in previews so that extension developers can make sure theirs run before ours.

src/packages/media/media/components/input-upload-field/input-upload-field.element.ts
```diff
@@ -45,17 +48,39 @@ export class UmbInputUploadFieldElement extends UmbLitElement {
...
 	#manager = new UmbTemporaryFileManager(this);
 
+	#previewers = new UmbArrayState(<Array<ManifestFileUploadPreview>>[], (x) => x.alias);
+
+	constructor() {
+		super();
+		new UmbExtensionsManifestInitializer(this, umbExtensionsRegistry, 'fileUploadPreview', null, (previews) => {
+			previews.forEach((preview) => {
+				this.#previewers.appendOne(preview.manifest);
+			});
```
iOvergaard, line 80: You will get the full updated list whenever you get an item here. So if you append one to the existing list, you will not remove any unregistered extensions. Have you considered storing the preview list in an array and overwriting each time?

src/packages/core/extension-registry/models/file-upload-preview.model.ts
```diff
@@ -0,0 +1,7 @@
+import type { UmbFileUploadPreviewElement } from '../interfaces/file-upload-preview.interface.js';
+import type { ManifestElement } from '@umbraco-cms/backoffice/extension-api';
+
+export interface ManifestFileUploadPreview extends ManifestElement<UmbFileUploadPreviewElement> {
+	type: 'fileUploadPreview';
+	forMimeTypes: Array<string>;
```
iOvergaard, line 6: I believe we need to add some jsdocs to this property with examples on what to put into it.

src/packages/core/extension-registry/models/file-upload-preview.model.ts
```diff
@@ -0,0 +1,11 @@
+import type { UmbFileUploadPreviewElement } from '../interfaces/file-upload-preview.interface.js';
+import type { ManifestElement } from '@umbraco-cms/backoffice/extension-api';
+
+export interface ManifestFileUploadPreview extends ManifestElement<UmbFileUploadPreviewElement> {
+	type: 'fileUploadPreview';
+	/**
+	 * Array of the mime types that are supported by this extension.
+	 * @example ["image/png", "image/jpeg", "image/*"]
```
iOvergaard, line 8:
```suggestion
	 * @examples [["image/png", "image/jpeg"], "image/*"]
	 * @required
```

src/packages/core/extension-registry/models/file-upload-preview.model.ts
```diff
@@ -0,0 +1,11 @@
...
+	 */
+	forMimeTypes: Array<string>;
```
iOvergaard, line 10: It looks like we have used the type `string | Array<string>` elsewhere for other such "for*Types" to ease the development pain, so you can add whatever you feel like. Could we update this type similarly?

```suggestion
	forMimeTypes: string | Array<string>;
```

You will most likely need to update your code throughout the components to handle it:

```js
const types = Array.isArray(manifest.forMimeTypes) ? manifest.forMimeTypes : [manifest.forMimeTypes];
const filter = types.includes(previewType);
```

src/packages/media/media/components/input-upload-field/input-upload-field.element.ts
```diff
@@ -65,17 +87,58 @@ export class UmbInputUploadFieldElement extends UmbLitElement {
...
 
+	#getPreviewElementAlias() {
+		const fallbackAlias = this.#previewers.find((preview) => preview.forMimeTypes.includes('*/*'))?.alias;
+
+		const mimeType = this.#getMimeTypeFromPath(this._src);
+		if (!mimeType) return fallbackAlias;
+
+		// Check for an exact match
+		const exactMatch = this.#previewers.find((preview) => {
+			return preview.forMimeTypes.find((type) => type === mimeType);
```
iOvergaard, line 98: I think `.includes` is slightly faster:

```suggestion
			return preview.forMimeTypes.includes(mimeType);
```

iOvergaard (changes requested):
I think the jsdocs could work like this. Also added a suggestion to accept both strings and arrays like we do elsewhere:

iOvergaard (approved):

PR #2236, opened 2024-08-23
feature: Only render a single Custom view + example with settings

nielslyngsoe:
Includes:
- Makes Extension slot have a `single` attribute/property that can be set to limit its render to only one. + Unit tests for this
- Fixes so we only render one custom view for a block.
- Example updated to use settings data.

## Types of changes

## Checklist


madsrasmussen (approved):
Looks good to me 👍

PR #2242, opened 2024-08-26
Fix: Determine when to create Blocks directly or via Block Workspace

nielslyngsoe:
Fixes so Blocks are created via a modal Workspace, unless its Block List Editor in Inline Editing Mode.
There was a few missing parts of the current source that this PR corrects.

This now also considers if a Block has any content properties, if not it will not be created via the Workspace, instead directly.
Meaning we do not open Blocks in a Workspace if they dont have any Content Properties.

## Types of changes

## Checklist


madsrasmussen (approved):
Reviewed as part of #2245

PR #2244, opened 2024-08-28
Fix bug change password for current user

NguyenThuyLan:
## Description
This PR fix bug can not change the current user password from this issue [#16943 ](https://github.com/umbraco/Umbraco-CMS/issues/16943)
1. Can not change password when click "Change password" by clicking my user-icon in the top right corner
2. Got error when click change password on User section

## Types of changes

## Checklist


madsrasmussen (changes requested):
Hi @NguyenThuyLan 

In the client code we have separate modules for "a user" and "the current user". I think these changes would fit better into the "current user module" and could possibly simplify the code a bit.

You can find the data source here: 
https://github.com/umbraco/Umbraco.CMS.Backoffice/blob/main/src/packages/user/current-user/repository/current-user.server.data-source.ts

Then you can add a method to the "current user"-repository to change the password here:
https://github.com/umbraco/Umbraco.CMS.Backoffice/blob/main/src/packages/user/current-user/repository/current-user.repository.ts

NguyenThuyLan:
Hi @madsrasmussen 
yes, but I couldn't avoid to change file` change-user-password.repository.ts` with adding OldPassword and IsCurrentUser params here, is that ok? Because when a user is a current user click to button "Change the password" at User Section, it will hit `change-user-password.action.ts`

madsrasmussen:
I did a bit of digging, and I can see you are right that we use the same modal for both a user and the current user which results in the "wrong" endpoint being called.

I would still like us to move your changes in the data source and repository as mentioned in the previous comment. Then we also have to change the repository depending on the type of user.

The repository is set up in the entity action code here: 
https://github.com/umbraco/Umbraco.CMS.Backoffice/blob/main/src/packages/user/user/entity-actions/change-password/change-user-password.action.ts#L26

You can make a check to see if the user unique matches the current user and then spin up the UmbCurrentUserRepository instead. You can see how we do something similar here:
https://github.com/umbraco/Umbraco.CMS.Backoffice/blob/main/src/packages/user/modals/change-password/change-password-modal.element.ts#L67

Feel free to reach out on Slack if it doesn't make any sense :D

madsrasmussen (approved):

PR #2247, opened 2024-08-30
Feature: Readonly mode for umb-collection element

madsrasmussen:
## Description

This PR adds `readonly` mode to the `umb-collection` element

General concept:
* The collection context includes a `ReadOnlyStateManager` that keeps track on read-only states.
* A state is added to the manager when the `readonly` attribute gets set on the umb-collection element.
* A new condition has to be added so extensions can check if a workspace is read-only. The condition is currently added to create actions so they only appear when a collection is writable.

**Notice:**
I have added a new workspace view for a "Content collection". A collection based on a content type with a data type configuration and variable data. It is currently a duplication of the more generic named "workspace-view-collection".

## How to test

* Add the `readonly` attribute to the element through developer tools

## Types of changes


nielslyngsoe (changes requested):
We figured out in a pair review that the Language Permissions should not affect Create abilities (this is the change to be reverted)
So we are left with a PR that just does some clean up/refactoring. Which should be completed, so the default 'Collection'-kind Workspace View does not depend on the Content Workspace Configuration/DataType. (Mads knowns the mission)
Therefor I marked it as Draft.

PR #2248, opened 2024-08-30
Feature: Readonly mode for Block List Property Editor UI

madsrasmussen:
## Description

This PR adds `readonly` mode to the Block List Property Editor UI

## How to test

* Add the `readonly` attribute to the element through developer tools

## Types of changes


src/packages/block/block-list/components/block-list-entry/block-list-entry.element.ts
```diff
@@ -252,40 +261,53 @@ export class UmbBlockListEntryElement extends UmbLitElement implements UmbProper
...
+						${this._settingsInvalid
+							? html`<uui-badge attention color="danger" label="Invalid settings">!</uui-badge>`
+							: nothing}
+					</uui-button>`
+				: nothing}
+		`;
+	}
+
+	#renderDeleteAction() {
+		if (this.readonly) return nothing;
```
nielslyngsoe, line 305: This seems not to work, unless I've done something wrong.

Its still rendered, despite begin in readonly mode.
(the way i got to it was by tabbing into the element. But I just wanted to highlight that this line does not prevent the rendering of this action)

I would also expect to be able to navigate into Blocks. With the overlay i get currently I cant. It also does not seem like the Block Workspace got the readonly state.

image

nielslyngsoe: Here a screenshot from tabbing to open the Block Workspace. Fields did not get readonly.
image

nielslyngsoe: it works now

PR #2257, opened 2024-09-06
Fix: Update dictionary's name fallback value when using it as a placeholder in `dictionaryItem_description` localization term

detjonmataj:
Used `zero-width space` character instead of `unnamed` instead for dictionary item name fallback in `dictionaryItem_description`.

## Description

As noted in umbraco/Umbraco-CMS#16385, using `unnamed` as a fallback for empty dictionary item names can be confusing, potentially being thought as an existing item. Displaying a quoted empty string is clearer and directly reflects the empty value.

In the code, placeholders in the format `%index%` and `{index}` are replaced with values from `args`. If an empty string is passed, it is treated as falsy, so placeholders are not replaced:

```javascript
if (typeof term === 'string') {
    if (args.length > 0) {
        term = term.replace(/(%(\d+)%|\{(\d+)\})/g, (match, _p1, p2, p3): string => {
            const index = p2 || p3;
            return String(args[index] || match);
        });
    }
}
```

Thus I used a zero-width space (HTML entity) because it is a truthy value and ensures placeholders are replaced correctly. I intentionally used the html entity value because if i copied the Unicode char it would cause more chaos for developers unless the text editor highlights it. Alternatively, using a visible space (` `) could be a simple solution and is barely noticeable in the UI. Another option might be to remove the placeholder completely or update it only when the item is saved. Either way the impact is small.

resolves  umbraco/Umbraco-CMS#16385

## Types of changes

## Motivation and context

I have been interested in contributing to Umbraco and saw this issue as a good opportunity to get involved with the project. Usually, I pay attention to details and try to ensure a smooth experience. This issue seemed like a perfect starting point for me.

## How to test?

1. Follow the [installation instructions](https://github.com/umbraco/Umbraco.CMS.Backoffice?tab=readme-ov-file#installation-instructions) to set up `Bellisima` locally.
2. Open `Bellisima` in your web browser.
3. Go to the `Dictionary` section from the top navigation bar.
4. Create new dictionary items with various values to observe their behavior.

### Using empty string for fallback
image

### Using single space as the fallback value
image

### Before the Fix
no-fix

### After the fix (using zero-width char)
fix

## Checklist

### DIFF
--- a/src/packages/dictionary/workspace/views/workspace-view-dictionary-editor.element.ts
+++ b/src/packages/dictionary/workspace/views/workspace-view-dictionary-editor.element.ts
@@ -84,7 +84,7 @@ export class UmbWorkspaceViewDictionaryEditorElement extends UmbLitElement {
 	override render() {
 		return html`
 			<uui-box>
-				${unsafeHTML(this.localize.term('dictionaryItem_description', this._dictionary?.name || 'unnamed'))}
+				${unsafeHTML(this.localize.term('dictionaryItem_description', this._dictionary?.name || '&#8203;'))}
 				${repeat(
 					this._languages,
 					(item) => item.unique,


detjonmataj:
Judging by the [results of the previews tests](https://github.com/umbraco/Umbraco.CMS.Backoffice/actions/runs/10733978984/job/29768295002), it appears that using `zero-width space` char as an html entity was a bad idea. When i replaced with a single space, the tests passed. Could be something to component support for zero-width chars or maybe it just an coincidence (unlikely).

nielslyngsoe (approved):
Thanks, Looks good

PR #2263, opened 2024-09-10
Bugfix: Blocks in RTE redirects to the frontpage

iOvergaard:
## Description

Adds the `UmbBlockRteEntriesContext` to the tinymce property editor UI so that the blocks plugin can calculate the path to the block picker modal. This is modeled after the block-list itself and is the missing piece to make the block picker work again following a regression between 14.1 and 14.2.

Fixes https://github.com/umbraco/Umbraco-CMS/issues/17003

Fixes https://github.com/umbraco/Umbraco-CMS/issues/16949

## Types of changes

## How to test?

1. Add an RTE to a page
2. Add blocks to the RTE and enable the "Insert blocks" button
3. Verify that the block picker opens when you click the button
4. Verify that a block is inserted into the editor (click the view source button) of the rich text editor


nielslyngsoe (approved):
Looks good

PR #2264, opened 2024-09-11
Build: Treat src/packages/* as workspaces

iOvergaard:
## Description

It is now possible to treat all src/packages/* modules as workspaces. We can change package.json to accommodate this fact, which improves the DX of adding new packages.

I took the liberty to optimize the newest package (sysinfo) a little bit so not all files are loaded upfront.


madsrasmussen (approved):
Looks good

PR #2270, opened 2024-09-11
Feature: Version check

iOvergaard:
## Description

Adds functionality to query the management api to check for new available versions.

## Screenshots

https://github.com/user-attachments/assets/51708657-3df5-4dec-8127-6cf4f54db8bc


leekelleher (approved):

I made a tweak to the "read more" by moving the icon next to the label, rather than below it.

Screenshot 2024-09-12 162803

PR #2287, opened 2024-09-13
Fix null ref exception in Context Data

Migaroez:
## Description

Fixes a simple nullcheck when the code determines a certain item in an array is an object but the value of the reference is null

## Types of changes

## Motivation and context

I was testing out the debugElement inside a user profile app and noticed it wasn't showing the console error directed me to the line and showed the missing truthy check.

The routerContext router value was null and thus threw and error when the system tried to figure out if type was an observable


nielslyngsoe (approved):
Great

PR #2312, opened 2024-09-18
Feature: Tiptap toolbar elements

leekelleher:
## Description

- Adds support for custom elements in the toolbar
- Adds a `button` kind for `tiptapExtension` for default toolbar buttons
- Amends how the toolbar extensions are loaded in/rendered

## Types of changes

## How to test?

If the Media Picker button is rendered on the toolbar, then it's working.

_**Note**: I've noticed a re-rendering when editing content, which is causing the toolbar to glitch... I expect that we'll revisit that particular code when we do the data-type configuration plumbing._

### DIFF
--- a/src/packages/rte/tiptap/components/input-tiptap/input-tiptap.element.ts
+++ b/src/packages/rte/tiptap/components/input-tiptap/input-tiptap.element.ts
@@ -75,7 +75,6 @@ export class UmbInputTiptapElement extends UmbFormControlMixin<string, typeof Um
 	}
 
 	async #loadEditor() {
-
 		const element = this.shadowRoot?.querySelector('#editor');
 		if (!element) return;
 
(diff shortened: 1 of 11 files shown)


src/packages/rte/tiptap/components/input-tiptap/input-tiptap.element.ts
```diff
@@ -120,13 +113,18 @@ export class UmbInputTiptapElement extends UmbFormControlMixin(UmbLitElement, un
...
-				.editor=${this._editor}
-				.extensions=${this._extensions}></umb-tiptap-fixed-menu>
+			${when(
+				!this._editor && !this._extensions?.length,
+				() => html`<uui-loader></uui-loader>`,
+				() => html`
+					<umb-tiptap-hover-menu .editor=${this._editor}></umb-tiptap-hover-menu>
+					<umb-tiptap-fixed-menu
+						.editor=${this._editor}
+						.extensions=${this._extensions}
```
iOvergaard, line 124: This isn't a property on the fixed menu anymore, is it?

leekelleher: Good spot, I've updated the PR. 👍

iOvergaard (approved):

PR #2318, opened 2024-09-19
Feature: Tiptap drag and drop upload

iOvergaard:
## How to test
1. Drag images and save
2. Check that the HTML is replaced by the actual media item
3. (Optional) Check that you can "edit" the item through the mediapicker plugin when that is set up


JesmoDev (approved):
LGTM

PR #2335, opened 2024-09-24
Feature: Tiptap blockpicker

iOvergaard:
## Description

- Remove the tight coupling between TinyMCE and the Block RTE context(s) changing from a push to a pull configuration to allow each editor (TinyMCE and Tiptap) to handle how they would like to insert blocks
- Add a new extension (block picker) to Tiptap that opens the blockpicker modal and adds commands to insert blocks
- Remove certain attributes marked as deprecated in V14 that were no longer used (originData)
- (Temporarily?) disable the 'delete' button on blocks, since you should remove any objects in a rich text editor by pressing the actual delete button on your keyboard (and I couldn't get Tiptap to play well with this function in due time)
- Finally fix the issue where TinyMCE could not load on the dev server due to a module not being loaded

## How to test

Create element types, add them to the rich text editor as blocks, test that they are inserted. Make sure you test both "inline" true and false.

**Test the frontend:**
Refer to the [Umbraco Docs](https://docs.umbraco.com/umbraco-cms/fundamentals/backoffice/property-editors/built-in-umbraco-property-editors/rich-text-editor/rte-blocks) to set up partial views to match the block views.

**Test Backoffice custom views:**
Go to this file `src/packages/block/custom-view/manifest.ts` and change the parameters to match your element types, for example:

```ts
export const manifest: ManifestBlockEditorCustomView = {
	type: 'blockEditorCustomView',
	alias: 'Umb.blockEditorCustomView.TestView',
	name: 'Block Editor Custom View Test',
	element: () => import('./custom-view.element.js'),
	forContentTypeAlias: 'numberBlock',
	forBlockEditor: 'block-rte',
};
```

Make sure `forBlockEditor` says "block-rte", otherwise it won't affect that editor.

### DIFF
--- a/index.html
+++ b/index.html
@@ -6,6 +6,13 @@
 		<link rel="icon" type="image/svg+xml" href="umbraco/backoffice/assets/favicon.svg" />
 		<meta name="viewport" content="width=device-width, initial-scale=1.0" />
 		<title>Umbraco</title>
+		<script type="importmap">
+			{
+				"imports": {
+					"@umbraco-cms/backoffice/block-rte": "/src/packages/block/block-rte/index.ts"
+				}
+			}
+		</script>
 		<script src="node_modules/msw/lib/iife/index.js"></script>
 		<link rel="stylesheet" href="node_modules/@umbraco-ui/uui-css/dist/uui-css.css" />
 		<link rel="stylesheet" href="src/css/umb-css.css" />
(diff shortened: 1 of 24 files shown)


leekelleher (approved):
Tested out against the relevant CMS branch (for PR https://github.com/umbraco/Umbraco-CMS/pull/17118).

Whilst testing against the CMS (as opposed to the dev server), I'd found that we were missing some extra Vite config and the build script highlighted a few compile errors, which I've fixed it.

✅ Blocks in Tiptap work great (both inline and non-inline/block).
✅ Custom views work great, (interestingly I wasn't able to get them to render in TinyMCE).

PR #2246, opened 2024-08-28
Feature: UFM component filters

leekelleher:
> [!IMPORTANT]
> **The details in this pull request may be outdated.**
> [Please refer to the official documentation for the latest UFM syntax.](https://docs.umbraco.com/umbraco-cms/reference/umbraco-flavored-markdown)

## Description

To support a "filter" feature to Umbraco-Flavored Markdown (UFM), this pull request adds an `ufmFilter` extension type and a base class (`UmbUfmElementBase`) for the UFM components in order to support reuse the filter logic.

Also in this pull request...

- Syntax for content pickers, `{~value}`, to support document, media and member pickers (both single and arrays), rendering the entity's name as the label
- Adds UFM filter functions (extensions), these are piped within the UFM curly bracket syntax; there is...
  - Fallback, i.e. `{= value | fallback:No items }`
  - Lowercase, i.e. `{= value | lowercase }`
  - Strip HTML, i.e. `{= value | strip-html }`
    - this is compatible with the RTE data structure, e.g. `{ markup, blocks }`
  - Title Case, i.e. `{= value | title-case }`
  - Truncate, i.e. `{= value | truncate:10:false }`
    - this supports multiple arguments: `length: number` and `tail: string` (defaulting to "…", and omitted if `"false"`)
  - Uppercase, i.e. `{= value | uppercase }`
  - Word Limit, i.e. `{= value | word-limit:10 }`
    - this supports a `limit: number` argument

Note, UFM filters can be piped together, e.g. `{= value | strip-html | word-limit:30 }`. The filters will also work for the localization and content picker UFM components, e.g. `{#general_choose | uppercase}`; `{~value | title-case | fallback: No items }`.

## Types of changes

- [x] New feature (non-breaking change which adds functionality)

## How to test?

For general testing of dynamic labels can be done within Collection Views, Block List/Grid and Document property descriptions.

For specific testing of the filters, e.g. in a Collection View, try adding the piped filter syntax to `{=value}` labels; also try piping filters, e.g. `{= value | lowercase | uppercase | title-case | truncate:4:true}`.

PR #2345, opened 2024-09-26
Adding Umbraco Flavored Markdown to handle block grid  expression  in a block list/grid and extract expressions from data source

IbrahimMNada:
## Description

When Creating a block list that contains url picker at the block list label, it shows only the document type alias.
which is not very helpful like this :. 

image

so using Umbraco Flavored Markdown I have developed a new markdown that enable us to see what ever expression we put inside of a block list label:

image

## Types of changes

## Motivation and context

When Creating a block list that contains url picker at the block list label, it shows only the document type alias.
which is not very helpful.

## How to test?
Go to the list configuration at the label section add 
`{@<url-picker-alias>[<target-index>].<prop-alias>}` exmaple : `{@mainNavigation[0].name}`

image

image

## Checklist


leekelleher:
@IbrahimMNada Thank you for your PR, I'll review it and feedback in due course. 🙏 

In the meantime, since Umbraco 14.3, there is a [content-name UFM component](https://github.com/umbraco/Umbraco.CMS.Backoffice/blob/v14.3.0/src/packages/ufm/components/content-name/content-name.element.ts) that may provide what you need, try `{~mainNavigation}`.
The idea of that specific component is to take a unique ID (or array list) and query the document (or media or member) repositories, then concatenate the names together by comma. This should work for the Content/Document Picker and Multi-URL Picker property-editors.

IbrahimMNada:
Thank you for your response, 
i tried to make it as generic as i could in the context of bloc-list 
not only for url content picker. 

this could help by giving full control over the label.  if you want to expand it on more than grid list 
please tell me im pretty sure we can agree on something

IbrahimMNada:
this feature was requested here as well , 
https://our.umbraco.com/forum/using-umbraco-and-getting-started/114798-v14-umbraco-flavored-markup-component-simple-example

just for you to know

leekelleher:
Hi @IbrahimMNada, thank you again for the time and effort you've put into this pull request. 🙏 

While I do understand the reasoning to use JavaScript syntax style to traverse down into the property data, this may lead implementors believing that other JavaScript expressions may be evaluated too; (e.g. as seen in [the linked forum post example](https://our.umbraco.com/forum/using-umbraco-and-getting-started/114798-v14-umbraco-flavored-markup-component-simple-example), `{= bodyText.replace('foo', 'bar')}`).

Support for an JavaScript `eval()`-esque feature isn't a goal for Umbraco-Flavoured Markdown (UFM), as there are potential performance and security concerns around this. Of course, we do want to UFM to support common scenarios, in more generalized ways. e.g. see the [UFM filters feature](https://github.com/umbraco/Umbraco.CMS.Backoffice/pull/2246) for v14.3.0.

With that said, I have chatted with several people in the community who are wanting to do similar with UFM as you've offered here. **Have you considered making an Umbraco package for this?** It would also [qualify as a Hacktoberfest contribution](https://github.com/umbraco/HacktoberfestActivityLog?tab=readme-ov-file#new-umbraco-packages).

I hope this isn't too disheartening, I'd encourage you to pursue the package creation path. I could see it becoming popular.

IbrahimMNada:
No this is not the case I do not use `eval `at all its simply walk through an object that's all 

I don't know man you can check the function

leekelleher:
@IbrahimMNada Sorry, I know the code here isn't using `eval()`, (I see the regex pattern). 👍 

My point is that if we offer support for `mainNavigation[0].name`, this is JavaScript array and property accessor syntax, so may open the suggestion that other JavaScript expressions are supported (which they are not), so more features get requested (in time) and it'll eventually become similar to `eval()`. Which isn't something we want to support in the core for UFM.

I still think this could be the basis for an awesome package.

IbrahimMNada:
This is the 2nd Pull Request , that is getting closed like this. 

there is no JavaScript evaluation its like a `visitor pattern ` that's all.

here is the function : its perfectly save 
```
getValueFromExpression(dataSource: any, expression: string) {
		const keys = [];
		const regex = /([a-zA-Z_$][\w$]*)|\[(\d+)\]/g;

		let match;
		while ((match = regex.exec(expression)) !== null) {
			if (match[1]) {
				keys.push(match[1]);
			} else if (match[2]) {
				keys.push(parseInt(match[2], 10));
			}
		}
		let target = dataSource;
		for (const key of keys) {
			if(target[key] === undefined) return undefined;
			target = target[key];
		}
		return target;
	}
```

IbrahimMNada:
Thanks brother , no worries !

IbrahimMNada:
hello , 
following your recommendations.

check out this new package :https://marketplace.umbraco.com/package/umbblocklist.dyanmiclabels

The (Dynamic Block List Labeling package) for Umbraco enhances the default block list by enabling dynamic, customizable labels. Instead of using static, predefined labels for block list items, you can now define labels that dynamically adapt based on the content of each block. This is especially useful for identifying and managing multiple entries in your block list, providing a more streamlined and user-friendly experience.

leekelleher:
@IbrahimMNada Super, thanks for the update! 🚀

PR #2346, opened 2024-09-26
Feature: Tiptap mediapicker

iOvergaard:
## Description

Adds the mediapicker to Tiptap including support for figures with figcaptions.

Also fixes a few issues with the "media-caption-alt-text-modal", namely: can now set caption and altText to preconfigured values when opening the modal and the image inside the modal is now also shown in a figure.

## How to test
1. Insert an image
2. Edit inserted images and check that the image is preselected in the modal
3. Insert an image with a caption and verify that it's inserted as a `<figure>` with the caption inside a `<figcaption>` element
4. Edit an inserted image with a caption and check that the image is preselected and that the stored caption and altText are shown in the modal
5. try and change values in the modal and see that the whole element (and nothing but the whole) is replaced


src/packages/rte/tiptap/extensions/umb/mediapicker.extension.ts
```diff
@@ -1,63 +1,143 @@
...
+						content: [
+							{
+								type: 'text',
+								text: media.caption,
+							},
+						],
+					},
+				],
+			});
+		}
```
iOvergaard, line 139: This could be moved to a command, but what should own it, image or figure? Let's wait and see if other extensions could benefit from it down the road.

PR #2384, opened 2024-10-01
Fix bug: No matching error display between New password and Confirm password

NguyenThuyLan:
## Description
This PR is fixing for the issue https://github.com/umbraco/Umbraco-CMS/issues/17157
When changing the password, input "New password" and "Confirm password" different values, it can still change the password. There is no matching error display.

## Types of changes

image

## Checklist


iOvergaard:
Thanks for taking a jab at this, @NguyenThuyLan. Looks like your fix is working, however, I think it might be good to add the message as custom validation to the confirm password field so that screen readers and the like are notified that there is an issue. If you add it as custom validation, the styling should also match out of the box.

You can see how we add custom validators for instance in this file: `src/packages/user/user/components/user-input/user-input.element.ts` with `this.addValidator`.

There is also the option to declare `error = true` and `errorMessage = 'xxx'` on the input field itself. I'm sure you can find inspiration for that in the code base :-)

iOvergaard (changes requested):
Thanks for taking a jab at this, @NguyenThuyLan. Looks like your fix is working, however, I think it might be good to add the message as custom validation to the confirm password field so that screen readers and the like are notified that there is an issue. If you add it as custom validation, the styling should also match out of the box.

You can see how we add custom validators for instance in this file: `src/packages/user/user/components/user-input/user-input.element.ts` with `this.addValidator`.

There is also the option to declare `error = true` and `errorMessage = 'xxx'` on the input field itself. I'm sure you can find inspiration for that in the code base :-)

leekelleher (approved):
With the backoffice repository being [merged with the main CMS repository this weekend](https://github.com/umbraco/Umbraco-CMS/discussions/17269), we wanted to go through all the outstanding pull requests, so I've reviewed this one and made some amends, following @iOvergaard's recommendations (using the built-in `.addValidator` mechanism).

Thanks again @NguyenThuyLan for tackling this issue! 🚀

PR #2400, opened 2024-10-02
V15 Hotfix: Media previews are not shown for anything other than images

iOvergaard:
## Description

This fix adds an `await` to the manifests initializer so we are sure the manifests have been loaded before trying to calculate which preview element to show.

It also optimizes the element slightly to get rid of the `_src` state and to rely more on the `value` property.


nielslyngsoe (approved):
Great, it works :-)

PR #2409, opened 2024-10-04
V15/feature/use uui ref href for document type input

nielslyngsoe:
Make clicking the name of the document-type in document-type input open the workspace — removing the edit action at right side.

image


madsrasmussen (approved):
Nice 🥳

PR #2422, opened 2024-10-07
V15: Ensure all localization variables are present in the default language

iOvergaard:
## Description

fix: ensure the fallback language `en.ts` has all of the same variables as the other languages, namely en-us had a few new variables

<!--- Describe


leekelleher (approved):
LGTM! 🈴

PR #2423, opened 2024-10-07
Feature: UFM as string

nielslyngsoe:
## Description

UFM Virtual Render Controller to enable rendering UFM without relying on a specific element implementation.

image

This branch needs considerations regarding whether it is acceptable to require UFM Element to be light-DOM(no shadow-DOM) to be read as part of a text-string version of the UFM.

Fixes

- https://github.com/umbraco/Umbraco-CMS/issues/17030
- https://github.com/umbraco/Umbraco-CMS/issues/17143
- https://github.com/umbraco/Umbraco-CMS/issues/17133
- https://github.com/umbraco/Umbraco-CMS/issues/17134

## Types of changes


src/packages/ufm/components/ufm-element-base.ts
```diff
@@ -36,6 +36,10 @@ export abstract class UmbUfmElementBase extends UmbLitElement {
 		});
 	}
 
+	override createRenderRoot() {
+		return this;
+	}
```
leekelleher, line 41: This is niggling at me, as inheriting from the `UmbUfmElementBase` class isn't a requirement for a UFM extension/component, so 3rd-party extensions may not implement this and lead to missing text/content in the rendered label.

I'm wondering if we could traverse the Shadow DOM to gather all the text content from the children/descendants?

e.g.

```ts
// snipped from ufm-virtual-render.controller.ts

override toString(): string {
	return this.#getTextFromDescendants(this.#element);
}

#getTextFromDescendants(element?: Element | null): string {
	if (!element) return '';

	const items: Array<string> = [];

	items.push(element.shadowRoot?.textContent ?? element.textContent ?? '');

	if (element.shadowRoot !== null) {
		Array.from(element.shadowRoot.children).forEach((element) => {
			items.push(this.#getTextFromDescendants(element));
		});
	}

	if (element.children !== null) {
		Array.from(element.children).forEach((element) => {
			items.push(this.#getTextFromDescendants(element));
		});
	}

	return items.filter((x) => x).join(' ');
}
```

There may be a more succinct way in code to do this, but you get the general idea.

leekelleher (approved):
LGTM! 🚀

PR #2444, opened 2024-10-10
Fix for issue with multiple server-validation errors

enkelmedia:
Fixes issue outlined here: https://github.com/umbraco/Umbraco-CMS/issues/17231 

## Description
This PR removes the constants used in the controller since this would prevent individual controllers for each field in a form. Before this change only one element (the last) in a form could be populated with a server validation error.

After moving/removing the constants each field not gets a unique binding and errors work as expected.

## Before this change
In this example, clicking the `Add server error`-button will add two errors to the context - one for each field. Only the last field shows the error.

umb-validation-before

## After this change
After applying the change, both of the fields show the error and resetting them by changing the value works.

umb-validation-after

## Types of changes

## How to test?
Use the provided example.

```
npm run example
```

Look in the Translation-section where there is a demo-dashboard.

## Checklist


enkelmedia:
Btw. I have a more extensive demo of using the validation context with tabs, that looks like this:

I could push this to the PR if you would like to include it as an example?

umb-validation-tab-demo

src/packages/core/validation/controllers/bind-server-validation-to-form-control.controller.ts
```diff
@@ -41,7 +41,7 @@ export class UmbBindServerValidationToFormControl extends UmbControllerBase {
 	}
 
 	constructor(host: UmbControllerHost, formControl: UmbFormControlMixinInterface<unknown>, dataPath: string) {
-		super(host, ctrlSymbol);
+		super(host);
```
nielslyngsoe, line 44: Maybe use this for the controllerAlias: `'umbFormControlValidation_'+simpleHashCode(dataPath)` — What do you think?

This is used to avoid the same one begin initialized twice. :-)

enkelmedia: I was thinking about that. My conclusion was that it would not work if multiple "things" bind to the same path - but when thinking closer - it would be a crazy thing to do :)

Do we have a "simpleHashCode" function or should we invent one?

enkelmedia: Found the simpleHashCode-function, it worked great. PR updated.

enkelmedia:
@nielslyngsoe I've updated the PR with the changes that you purposes and also pushed a cleaned up example with tabs.

enkelmedia:
@nielslyngsoe would be nice if you could have a look at this, let me know if you need any changes.

nielslyngsoe (approved):
really cool, thanks

PR #2448, opened 2024-10-11
Feature: reusable table column view components

madsrasmussen:
## Description

This PR is a small step towards reusable code across collection table views. We currently have duplicated code across our table column layout elements. This PR adds a "boolean" element and an "entity actions" element we now reuse across the tables. It will ensure the same UX across those columns.

The PR also implements entity actions for all collection table views.

I am unsure whether the word "layout" is used correctly in this instance. I am open to suggestions for a new name :)

Example:
Screenshot 2024-10-11 at 20 58 58

## Types of changes


bjarnef:
@madsrasmussen perhaps "column template" or "column view" as we have "collection view".

Or if specific to layout of cell, perhaps "table cell view"?

madsrasmussen:
hi @bjarnef 

I think "view" is better than "layout" so I have renamed it to that. I like that they are named "column" instead of "cell" because the view applies to all cells in a column. 🤔 

Thanks 🙌

leekelleher (approved):
Tested out, looking great! 🚀

PR #2449, opened 2024-10-12
Chore: Update's VS Code new `UmbLitElement` code snippet

leekelleher:
## Description

Updates the VS Code code snippet for creating new `UmbLitElement` classes.

## Types of changes

## Motivation and context

The original contents doesn't follow our current practices.

Changed the command to be `new umb element`, (as the previous `lit new` felt too generalised).

## How to test?

Create a new empty TypeScript file, start typing `new umb element` (which should appear in the autocomplete), press tab/enter to complete... and the class scaffold should appear, with the relevant naming of the class and element names.


.vscode/lit.code-snippets
```diff
@@ -1,30 +1,36 @@
...
+			"\tconstructor() {",
+			"\t\tsuper();",
+			"\t}",
+			"",
+			"\toverride render() {",
+			"\t\treturn html`$0`;",
 			"\t}",
 			"",
-			"\tstatic styles = [UmbTextStyles, css``];",
+			"\tstatic override styles = [UmbTextStyles, css``];",
```
iOvergaard, line 22: This could be `readonly` as it never changes (and it prevent a Sonarcloud issue):

```suggestion
			"\tstatic override readonly styles = [UmbTextStyles, css``];",
```

.vscode/lit.code-snippets
```diff
@@ -1,30 +1,36 @@
...
-			"import { customElement } from 'lit/decorators.js';",
+			"import { css, customElement, html } from '@umbraco-cms/backoffice/external/lit';",
 			"import { UmbLitElement } from '@umbraco-cms/backoffice/lit-element';",
+			"import { UmbTextStyles } from '@umbraco-cms/backoffice/style';",
 			"",
-			"@customElement('umb-${TM_FILENAME_BASE/(.*)\\..+$/$1/}')",
-			"export class Umb${TM_FILENAME_BASE/(.*)\\..+$/${1:/pascalcase}/}Element extends UmbLitElement {",
-			"\trender() {",
-			"\t\treturn html`${0:umb-${TM_FILENAME_BASE/(.*)\\..+$/$1/}}`;",
+			"const elementName = 'umb-${TM_FILENAME_BASE/(.*)\\..+$/$1/}';",
```
iOvergaard, line 10: We have this in a few places, but was it ever a convention? It prevents one of the custom eslint plugins from running, which checks for correct element names, I think?

.vscode/lit.code-snippets
```diff
@@ -1,30 +1,36 @@
...
+			"",
+			"@customElement(elementName)",
+			"export class Umb${TM_FILENAME_BASE/(.*)$/${1:/pascalcase}/}Element extends UmbLitElement {",
+			"\tconstructor() {",
+			"\t\tsuper();",
+			"\t}",
+			"",
```
iOvergaard, line 17: Sonarcloud will complain about empty constructors. I would say we better leave it out. Developers are probably better off using the lifecycle callbacks `connectedCallback` and `firstUpdated` instead:

```suggestion
```

.vscode/lit.code-snippets
leekelleher: I didn't know this was causing an issue with an eslint plugin. I can revert this.

iOvergaard: I think it's nicer this way, as you only have to change an element's name in one place. Maybe we can update the eslint rule accordingly. But let us keep it here!

iOvergaard (approved):

.vscode/lit.code-snippets
leekelleher: Oh, I'd already removed this code before the merge. Let's discuss with the rest of the frontend team at our next meeting.

PR #2461, opened 2024-10-16
Removes unnecessary margin in ufm-render-element

skttl:
I noticed that property labels now use umb-ufm-render, which (because of the nature of markdown I guess), turns simple text into paragraphs. This makes it have a top and bottom margin from the user agent stylesheet.

I added some css to remove the top margin from the first element and bottom margin from the last to combat this.

Before:

thorium_OksH6ml63u

After:

thorium_xNvOPDNZWy

### DIFF
--- a/src/packages/ufm/components/ufm-render/ufm-render.element.ts
+++ b/src/packages/ufm/components/ufm-render/ufm-render.element.ts
@@ -59,6 +59,14 @@ export class UmbUfmRenderElement extends UmbLitElement {
 			pre {
 				overflow: auto;
 			}
+
+			:host > :first-child {
+				margin-block-start: 0;
+			}
+
+			:host > :last-child {
+				margin-block-end: 0;
+			}
 		`,
 	];
 }


src/packages/ufm/components/ufm-render/ufm-render.element.ts
```diff
@@ -59,6 +59,14 @@ export class UmbUfmRenderElement extends UmbLitElement {
 			pre {
 				overflow: auto;
 			}
+
+			:host > :first-child {
+				margin-block-start: 0;
```
leekelleher, line 64: I'm happy with reducing the excessive margins (caused by the `<p>` tags). 👍 

I'm thinking that a small top margin (3px) would be a good fit. What do you think @skttl?

```suggestion
				margin-block-start: var(--uui-size-1);
```

nielslyngsoe: Hi @leekelleher just chipping in, I think it would be good to have a minimalistic approach on the space taken by UFM-render. I think its better if the specific implementation append the margin to it, if its needed there — in that way enabling UFM to be used in very narrow spaces, or in cases where other things makes up the white-space. :-)

leekelleher (approved):
Thanks @skttl! 🙏

PR #2462, opened 2024-10-16
Bugfix: TinyMCE anchor links without base URL

leekelleher:
## Description

When adding/editing a link in TinyMCE that has an anchor or querystring, but no URL part, the link markup would include the current backoffice path, along with an `undefined` suffix.

Fixes https://github.com/umbraco/Umbraco-CMS/issues/17009

## Types of changes

## How to test?

Add/edit a link without entering a URL, but adding either an anchor hash or querystring value. The link's `href` should **not contain** either `undefined` or the current backoffice path.


iOvergaard (approved):
LGTM!

PR #2463, opened 2024-10-16
bug: 17282

### LINKED ISSUE #17282: v15.0-rc1 Variant Blocks create duplicates in the frontend when created in an area
### Which Umbraco version are you using? (Please write the *exact* version, example: 10.1.0)

15.0-rc

### Bug summary

When creating a simple block inside a slot in an area, across variant cultures, the "first" variant will succed, and the second "greyed" out block will create a duplicate.

### Specifics

[Watch demonstration](https://vimeo.com/1019870791)

Watch 30 sec video demonstration follow steps to reproduces

### Steps to reproduce

Create a block grid on an umbraco site with 2 or more languages, on  a document type that allows vary by culture.

The block grid should be invariant, and should have an invariant element type for the area and a variant element type to put inside the area.

If you create  the first variant (ie. English or something) and fill out the area, then when it comes to fill in the other variant, which now should have a greyed out block, creating it will result in a duplicate being added below the area as well.

They must also have duplicate keys, because deleting the duplicate, will also delete the original on all variants.

### Expected result / actual result

I'd expect the blocks to only be created inside the area slots without duplicates

nielslyngsoe:
Fixes: https://github.com/umbraco/Umbraco-CMS/issues/17282

Sort of a duplicate of:
https://github.com/umbraco/Umbraco.CMS.Backoffice/pull/2464


iOvergaard:
There are no files changed in this PR after merging #2464 into release/15.0 and then release/15.0 -> main, so closing this PR.

PR #2465, opened 2024-10-16
Tiptap: allow images to be uploaded from the clipboard

iOvergaard:
## Description

Add a listener for "paste" events and upload any files that might be included in the clipboard.

Fixes https://github.com/umbraco/Umbraco-CMS/issues/17258

## How to test

Take a screenshot or copy an image into the clipboard, then paste it (CTRL+V) into a Tiptap instance with the media upload extension enabled.


leekelleher (approved):
Tested out, works great. 🚀 

The code fix was much more simpler than I'd expected, kudos! 🙌

PR #2467, opened 2024-10-17
Feature: Collection Toolbar Filter Field

leekelleher:
## Description

Since reviewing the Dictionary Collection toolbar filter (PR #2415) last week, I looked at the other Collection toolbars to see if we could de-duplicate the filter input code, to make a reusable component.

This PR introduces a `<umb-collection-filter-field>` component.

## Types of changes

## How to test?

Try out the Collection toolbar filter field on the following areas, to make sure that it still works as expected.

- Dictionary
- Documents
- Extension Insights
- Media
- Members
- Users
- User Groups

### DIFF
--- a/src/packages/core/collection/components/collection-action-bundle.element.ts
+++ b/src/packages/core/collection/components/collection-action-bundle.element.ts
@@ -1,4 +1,4 @@
-import { html, customElement } from '@umbraco-cms/backoffice/external/lit';
+import { html, customElement, css } from '@umbraco-cms/backoffice/external/lit';
 import { UmbLitElement } from '@umbraco-cms/backoffice/lit-element';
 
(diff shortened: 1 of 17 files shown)


iOvergaard (approved):
Looks good to me. I'm not sure if any of the now-deprecated elements were ever part of any index.ts export and available from the outside, so they could potentially be removed or have their exports changed already, but no comments other than that.

leekelleher:
@iOvergaard Thanks for the merge. 🙏 

My thinking was that a 3rd-party extension may have use something like `<umb-document-collection-toolbar>` in their own components, as opposed to the `UmbDocumentCollectionToolbarElement` class/type explicitly. Personally I doubt anyone would be using that particular component externally, but still we don't know for sure. 😬

PR #2471, opened 2024-10-18
Content type group sorting placeholder

bjarnef:
## Description

Currently I don't think it is clear when a content type group is sorted as when `drag-placeholder` attribute is added the entire element is invisible.
Sorting is initially enabled (which is shouldn't be here) because the group has `draggable="true` - I think I created an issue for that in *Umbraco.CMS.Backoffice* repository, but the issue tracker were later disabled (and issues reported in *CMS* repository instead), so I can't access my own initially reported issues 😅

image

I think it is best practice to have a drop placeholder indication where the sorted item is placed when releasing dropping item (normally using mouse, but could be touch as well).

image

## Types of changes

## Checklist

### DIFF
--- a/src/packages/core/content-type/workspace/views/design/content-type-design-editor-group.element.ts
+++ b/src/packages/core/content-type/workspace/views/design/content-type-design-editor-group.element.ts
@@ -228,8 +228,12 @@ export class UmbContentTypeWorkspaceViewEditGroupElement extends UmbLitElement {
 				opacity: 0.5;
 			}
 
-			:host([drag-placeholder]) > * {
+			:host([drag-placeholder]) > uui-box > * {
 				visibility: hidden;
+			}
+			
+			:host([drag-placeholder]) > uui-box > div[slot='header'] {
+				visibility: visible;
 			}
 
 			div[slot='header'] {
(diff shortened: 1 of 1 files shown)


bjarnef:
@iOvergaard @nielslyngsoe draggable is set to `true` by default here:
https://github.com/umbraco/Umbraco.CMS.Backoffice/blob/dac0c038e3b299cd08b3f681b52aac9d317947f5/src/packages/core/sorter/sorter.controller.ts#L457

How can we explicit set this when enable sorting mode, while in other context may be fine to sort directly.

leekelleher (approved):
Thanks @bjarnef, the drag placeholder works as expected! 👍 

With regards to the `draggable` being set on the initial view, that's definitely an issue that we'll look in to, but I didn't want to hold up getting this PR merged in.

I'll cherry-pick this for v15.0.0-rc3.

bjarnef:
@leekelleher great. Maybe if was just I my machine, but I think it couldn't drag tabs and re-arrange them. Or then placeholder when sorting these could be better.

There are probably some other consistency to consider as well, when dragging items in pickers, dragging blocks (or groups) in Block List/Grid.

Dragging items in Block Grid (content) did have a placeholder style though :)

PR #2474, opened 2024-10-21
Feature: Add read only tag for read only languages in app language select

madsrasmussen:
## Description

This PR adds the read-only tag to all read-only languages in the app language picker.

It also adds a read-only state manager to the app language context so anyone can know if the current app language is read-only.

Screenshot 2024-10-21 at 20 21 32

## Types of changes


nielslyngsoe (approved):
Very nice

PR #2479, opened 2024-10-23
Bugfix: Prevent delete a template when has children

leekelleher:
## Description

Partial fix for https://github.com/umbraco/Umbraco-CMS/issues/16443

When a template has child templates, the Delete action is available, (which would delete the parent template and orphan the child templates).

This PR adds a **Template Has No Children Condition** to prevent the **Delete** action being displayed for templates that have child templates.

## Types of changes

## How to test?

1. Create a template.
2. Create another template below the first one created.
3. Try to delete the parent template with children; does the **Delete** action display?


madsrasmussen (commented):
Everything works as expected. Awesome!

I have done something similar for users where I went with a name like: `UmbTemplateAllowDeleteActionCondition`. 

In the user case I had multiple conditions I needed to check so I combined them into one. It will make it possible to add tests where it might be difficult when they are split in JSON. Sometimes the server also sends a "canDelete/canEdit" etc. attribute we need to include in the check.

Just for the sake of sharing ideas: I think a "hasChildren/noChildren" condition will make sense though. I would just make it very general with its own "hasChildren"-context so it can be used in trees, workspaces, collections, etc.

How does that sound to you?

leekelleher:
@madsrasmussen Thanks for the feedback, I've updated accordingly.

madsrasmussen (approved):
fantastic 💯

PR #2483, opened 2024-10-24
Feature: Content Detail Workspace Base

madsrasmussen:
## Types of changes

### DIFF
--- a/src/packages/core/content-type/types.ts
+++ b/src/packages/core/content-type/types.ts
@@ -13,6 +13,7 @@ export interface UmbPropertyTypeContainerModel {
 
 export interface UmbContentTypeModel {
 	unique: string;
+	entityType: string;
 	name: string;
 	alias: string;
 	description: string;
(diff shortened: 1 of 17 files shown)


nielslyngsoe:
Hi @madsrasmussen I have been testing but there is some routing problem, might have to do with Discard Changes, when you try to create a new Document but already are on one.

Also it asks if you like to Discard Changes, thought none was made? — maybe we should have a look at why that happens as well.
image

nielslyngsoe:
Hi @madsrasmussen another problem appears when creating a new Document, this triggers a redirect from 'create' to 'edit'. which triggers a 'discard change'

nielslyngsoe:
And just to give you a whole picture right here at hand, then there is some circular dependency problem, as you can see in the tests :-)

madsrasmussen:
> might have to do with Discard Changes, when you try to create a new Document but already are on one.

This is fixed by: https://github.com/umbraco/Umbraco.CMS.Backoffice/pull/2514

PR #2484, opened 2024-10-24
Bugfix: Hide template input when there are no allowed templates

madsrasmussen:
## Description

This PR hides the template input on a document info view when there are no configured allowed templates. We do this to prevent an experience where the editor opens an empty template picker or a template picker with only disabled items in it. The same UX will be useful when the CMS is fully headless and there are no templates.

Here is a screenshot of something that is not there  😄 

Screenshot 2024-10-24 at 23 39 23

## Types of changes


leekelleher (approved):

PR #2487, opened 2024-10-25
Hotfix: Correct user enable notification

madsrasmussen:
## Description

We currently display an incorrect notification when a user is enabled. The PR addresses this issue and also uses localized text when enabling and disabling.

## Types of changes


leekelleher (approved):
Tested out, works with the localization (displaying the user's name), all good! 🚀

PR #2492, opened 2024-10-28
Feature: Document Workspace Info reload components

leekelleher:
## Description

Fixes https://github.com/umbraco/Umbraco-CMS/issues/17183.

Half feature, half bug fix. When saving/publishing a document, the components on the Info workspace view were not automatically updating. Using the Request Reload Structure event (from the workspace context), the Info components are refreshed on save/publish.

For the Links component, this makes use of a new dedicated Document URL Repository, (as opposed to observing the workspace context's `.urls` property).

## Types of changes

## How to test?

Go to an existing document, go to the Info workspace view, rename the document (e.g. append a number), press Publish, notice that the URL is updated in the "Links" component and that the "History" component has the latest Save/Publish audit logs.

### DIFF
--- a/src/packages/core/components/history/history-item.element.ts
+++ b/src/packages/core/components/history/history-item.element.ts
@@ -63,6 +63,7 @@ export class UmbHistoryItemElement extends UmbLitElement {
 			.user-info div {
 				display: flex;
 				flex-direction: column;
+				min-width: var(--uui-size-60);
 			}
 
 			.detail {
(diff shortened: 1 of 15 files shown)


iOvergaard (approved):

mjpraxis:
@leekelleher @iOvergaard 

Is the "half bugfix" that it fixes the update of the info-boxes?

And the other half is that the bug regarding update of the property values is not fixed?
If so then this issue should still be "open" https://github.com/umbraco/Umbraco-CMS/issues/17183


image

leekelleher:
@mjpraxis Thanks for the follow-up. 🙏  My "half bugfix" comment was more that this PR also adds a feature, (the document URL repo code), not exclusively a bug fix, meaning it would be a bit cheeky of me to try to push it into the upcoming v15.0.0 release, so it'll be scheduled for v15.1.0.

I believe the other half of the issue you raised (about the server-side updates), should be resolved by PR #2491 (which will be part of v15.0.0), I just forgot to link to your issue on that PR. 🤦

PR #2499, opened 2024-10-30
Correct the notification message when a user is enabled

manutdkid77:
## Description

This PR corrects the notification message when a user is enabled.

Previously it used to display "User disabled" when enabling a user.

Replicated on v15.1.0-rc1.preview.16

## How to test

* Go the Users section in the backoffice.
* Create a new user or use an existing user
* Disable the user.
* Enable the user.
* The correct notification message should be displayed.

https://github.com/user-attachments/assets/91ef7533-00d6-4358-91f4-3dadbba77302


leekelleher:
@manutdkid77 Thank you for your pull request for this, but it's already been fixed in #2487 (for v15.0.0-rc3).

PR #2503, opened 2024-11-01
Feature: Add Swiss backoffice localizations

marwaldv:
Add missing backoffice languages as properties etc. are translated via translation section (e.g. with "#Backoffice Settings"). I've been asked to add the swiss languages files here by @iOvergaard here [https://github.com/umbraco/Umbraco-CMS/pull/17177#issuecomment-2450751562](https://github.com/umbraco/Umbraco-CMS/pull/17177#issuecomment-2450751562)


leekelleher (approved):
@marwaldv Thank you for your pull request! 🙏 

I've made a couple of amends to the pull request, firstly I added the localizations the manifest (to register them within the backoffice).

The second amend was that I'd noticed the Swiss German, French and Italian localizations were exact duplicates of the existing German, French and Italian localizations, effectively adding approx. 6,000 lines of code, so I opted to make use of the JavaScript approach and imported the native localizations, which can then be appended to if there are any future specific Swiss variations to the labels.  To note, this approach would not be possible in the old backoffice (v13 and prior) due to the nature of using the XML format.

Thank you again for making this contribution. 👍

PR #2510, opened 2024-11-04
Hotfix: UFM alias prefixes

leekelleher:
## Description

Following internal review and feedback, the single character marker prefixes can be confusing. For established syntax like outputting a raw value or using a localization key, the respected marker prefixes `=` and `#` are intuitive, however with the new UFM Content Name component, the chose marker prefix is `~`, which is not established and counter intuitive.

To aid discovery, the use of a UFM component alias will be encouraged, e.g. `{umbContentName: propertyAlias}`

The `~` marker prefix will still work and be supported in v15, but set to be deprecated in a future release, the `umbContentName` alias prefix will be used going forwards.

## Types of changes

## How to test?

First, test that the existing UFM syntaxes still work as expected. Then test that the "umbContentName" alias prefix works in various places, e.g. Block List, Collection Views, property descriptions.

### DIFF
--- a/src/packages/ufm/components/manifests.ts
+++ b/src/packages/ufm/components/manifests.ts
@@ -6,20 +6,20 @@ export const manifests: Array<ManifestUfmComponent> = [
 		alias: 'Umb.Markdown.LabelValue',
 		name: 'Label Value UFM Component',
 		api: () => import('./label-value/label-value.component.js'),
-		meta: { marker: '=' },
+		meta: { alias: 'umbValue', marker: '=' },
 	},
 	{
 		type: 'ufmComponent',
 		alias: 'Umb.Markdown.Localize',
 		name: 'Localize UFM Component',
 		api: () => import('./localize/localize.component.js'),
-		meta: { marker: '#' },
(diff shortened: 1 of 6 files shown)


src/packages/ufm/components/manifests.ts
```diff
@@ -15,6 +15,9 @@ export const manifests: Array<ManifestUfmComponent> = [
 		api: () => import('./localize/localize.component.js'),
 		meta: { alias: 'umbLocalize', marker: '#' },
 	},
+	/*
+	 * @deprecated since version 15.0-rc3
```
leekelleher, line 19: This isn't correct, as the UFM component (manifest) `Umb.Markdown.ContentName` isn't being deprecated, only the `~` marker prefix will be deprecated. We'll still support `{umbContentName: alias}`.

nielslyngsoe: oh yes, I though it as one registrating pr. 'marker' but I see it now. I have reverted.

nielslyngsoe (approved):

PR #2522, opened 2024-11-07
Bugfix: Tiptap Link Picker, fixes anchor with an empty URL

leekelleher:
## Description

In the Tiptap Link Picker extension, when adding a link with an anchor/querystring and empty URL (for in-page navigation), the `href` was being stringified with the URL's `null` value.

## Types of changes

## How to test?

In the Tiptap editor, open the Link Picker modal, set an anchor/querystring value without setting the URL field, press Submit.
Select the link you've just created, edit in the Link Picker modal, notice that the URL value is empty and **not** set to a stringified `null` value.

### DIFF
--- a/src/packages/tiptap/extensions/toolbar/link.extension.ts
+++ b/src/packages/tiptap/extensions/toolbar/link.extension.ts
@@ -75,7 +75,7 @@ export default class UmbTiptapToolbarLinkExtensionApi extends UmbTiptapToolbarEl
 
 		const anchor = this.#getAnchorFromQueryString(queryString);
 
-		if (anchor) url += anchor;
+		if (anchor) url = (url ?? '') + anchor;
 
 		if (!url) return null;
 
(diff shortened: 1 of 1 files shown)


src/packages/tiptap/extensions/toolbar/link.extension.ts
```diff
@@ -75,7 +75,7 @@ export default class UmbTiptapToolbarLinkExtensionApi extends UmbTiptapToolbarEl
 
 		const anchor = this.#getAnchorFromQueryString(queryString);
 
-		if (anchor) url += anchor;
+		if (anchor) url = (url ?? '') + anchor;
```
nielslyngsoe, line 78: In general I would not recommend writing if-sentences without the { }, I think it can be acceptable if its a return statement like this:

if (!missingThing) return;

But all other cases offen ends up with someone misunderstanding the scope of that code-block.
And then I also find it easier to read with { }

:-D

nielslyngsoe (approved):

PR #2523, opened 2024-11-07
Bugfix: creating new doc type properties sort order doesn't count existing properties - puts it at top (+1)

### LINKED ISSUE #17457: 15rc: Creating new doc type properties sort order doesn't count existing properties
### Which Umbraco version are you using? (Please write the *exact* version, example: 10.1.0)

15.1.0-rc1.preview.37

### Bug summary

Adding doc type properties to an existing doc type with 2+ existing properties puts them always in position 1 not at the bottom as expected.

### Specifics

Tested on firefox - same as Chrome. 

### Steps to reproduce

If you create a new doctype with a single tab and create 2+ properties. 

Save the doc type and reload.

Add a new doc type property - the new property is added **second** in the list of existing properties and not at the bottom. 

It appears to be given sort order = 0, and is positioned after the first, existing 0.

This might have been missed in testing if that you have only one existing property it does go to the bottom as expected, the problem is only visible with more than 1 existing property.  

### Expected result / actual result

I would expect new properties to be added to the bottom as per previous versions until you change the sort order.

cheeseytoastie:
## Description

Fixes https://github.com/umbraco/Umbraco-CMS/issues/17457

Creating new doc type properties sort order doesn't count existing properties.

Looking at this it seems that the hardwork is done to workout the sort order and assign it to preset but it's not returned undefined was (probably as a dev placeholder that's gone unnoticed!).

## Types of changes

## Checklist

### DIFF
--- a/src/packages/core/content-type/workspace/views/design/content-type-design-editor-properties.element.ts
+++ b/src/packages/core/content-type/workspace/views/design/content-type-design-editor-properties.element.ts
@@ -193,7 +193,7 @@ export class UmbContentTypeDesignEditorPropertiesElement extends UmbLitElement {
 					}
 					preset.sortOrder = sortOrderInt;
 				}
-				return { data: { contentTypeUnique: this._ownerContentTypeUnique, preset: undefined } };
+				return { data: { contentTypeUnique: this._ownerContentTypeUnique, preset: preset } };
 			})
 			.observeRouteBuilder((routeBuilder) => {
 				this._newPropertyPath =


JasonElkin (approved):
Tested. Fixes the problem as described.

Thanks @cheeseytoastie!

leekelleher (approved):
@cheeseytoastie Thank you for the fix! 🚀 
Thanks to @JasonElkin for testing, I've also verified the fix. Go team! 🙌

